PR #59, opened 2015-09-24
I made an icon.

ke7ofi:
It’s “sh” in one of the Elven scripts from Tolkien’s works. Shells don’t usually get icons (Terminals do.), but you might be able to use it on the website or as the profile image for @elves.


xiaq:
Good :)

But before I merge it, would you clarify under what license it is covered?

ke7ofi:
If possible, I’d like to put it under the MIT license (assuming icons count as documentation) or similar, but I don’t really care. I’m just trying to help out, and I’m happy as long as I get credit and it doesn’t get too restrictive.

xiaq:
@ke7ofi Hi! I have been come back to this project and found this conversation broken in the middle.

After some re-considerations, thank you - but no.

ke7ofi:
I'm not trying to change your mind here, but what were your considerations? Was it just that shells don't need icons? I'm a bit curious.

xiaq:
@ke7ofi I am not familiar with Tolkien's works and I don't want to identify elvish with them (yet). The icon also feels too arcane for me.

ke7ofi:
Fair enough, that makes sense.

PR #60, opened 2015-11-15
adding beep()

fionser:
:)


xiaq:
I left the `beep` function unimplemented since I intended to copy the behavior of fish, which instead of actually trying to beep, flashes the whole line for a short period of time.

I renamed to method to `flash` in 97302b5504a6d2a10341ef09e385036f9349c36a.

PR #113, opened 2016-02-12
support completion ~uname without slash

zhsj:
this enhances the #86


zhsj:
However maybe it's better to complete uname after type `~`. And add `/` after typed `~uname`.

xiaq:
Just like you said, ~ without a slash should complete username instead.

PR #124, opened 2016-02-13
[WIP]fix history unchanged when finding next history

zhsj:
If you type `Down` in history mode, you will find the cmd is not changed. Because it always gets the current one...


xiaq:
This breaks the test... better fix the user of it in `edit`.

xiaq:
Fixed in 011f4dcaa2add5e7772a4c53e4d239c579db18e1.

Issue #63, opened 2016-01-27
History listing


xiaq:
Implement a history listing functionality akin to [ptpython](https://github.com/jonathanslenders/ptpython).

PR #149, opened 2016-02-20
edit: Fix crash in history listing mode

Zirak:
When you had too few history entries to display in the terminal,
entering history listing mode attempted to access a negative index and
crashed.

### DIFF
--- a/edit/writer.go
+++ b/edit/writer.go
@@ -591,11 +591,26 @@ tokens:
 			}
 		} else if hist != nil {
 			n := len(hist.all)
-			for i := n - hListing; i < n; i++ {
+
+			i := 0
+			if n > hListing {
+				i = n - hListing
+			}
+
+			for ; i < n; i++ {
 				b.writes("\n"+hist.all[i], "")
 			}
+
 			n = len(b.cells)
-			b.trimToLines(n-hListing, n)
+
+			startIndex := 0
+			if n > hListing {
+				startIndex = n - hListing
+			}
+
+			if len(b.cells) > 0 {
+				b.trimToLines(startIndex, n)
+			}
 		}
 	}
 


Zirak:
I've wanted to make two additional edits but was unsure about their possible side-effects on other parts.

The first was not entering history listing mode if there was no history. Off I went to [`startHistoryListing`](https://github.com/Zirak/elvish/blob/master/edit/builtins.go#L373), but got stuck a bit in the semantics. What happens if `newHistoryListing` returns an error (which makes sense in our situation)? Would the user still be in history listing mode? Perhaps it'll be best then to move the mode switch below the check.

The second was in [`trimToLines`](https://github.com/Zirak/elvish/blob/master/edit/writer.go#L159), I badly yearned to put a `if len(b.cells) < low` check in the beginning, but it may introduce subtle bugs (why aren't things trimming?) if used in certain contexts, and it felt wrong to do it without more knowledge of the system.

Great job on the shell, btw. It's mighty impressive.

xiaq:
Hi! Thanks for the fix. You are correct that the mode switch should be moved down, and I prefer to leave `trimToLines` as it is, due to the concerns you just stated.

By the way, history listing was something I didn't finish; the ultimate goal is to steal the design ptpython's history listing (#63), which allows you to scroll through the whole history and more importantly, compose a chunk of code by cherry-picking multiple entries from the history. Before trying to implement it, I observed that it has quite a lot in common with existing completion and navigation listings -- esp. wrt. the scrolling and trimming behavior -- so a good abstraction should be made to capture the common behavior. Unable to come up with a good abstraction, I didn't bother to think hard and turned to other parts of elvish instead, leaving this unfinished thing in the code :)

If you find this interesting enough, you are more than welcome to contribute. Keep me informed about your progress, so that I won't rewrite the whole line editor in a midnight and ruin all your efforts.

PR #158, opened 2016-03-02
add a new builtin func: kill small word

zhsj:
It considers a word is splitted by following char:
'-', '_', '/', '—', '[', ']', '{', '}', '(', ')' and space.

### DIFF
--- a/edit/builtins.go
+++ b/edit/builtins.go
@@ -17,6 +17,7 @@ var builtins = []Builtin{
 	{"kill-line-left", killLineLeft},
 	{"kill-line-right", killLineRight},
 	{"kill-word-left", killWordLeft},
+	{"kill-small-word-left", killSmallWordLeft},
 	{"kill-rune-left", killRuneLeft},
 	{"kill-rune-right", killRuneRight},
 	{"move-dot-left", moveDotLeft},
--- a/edit/insert.go
+++ b/edit/insert.go
@@ -63,6 +63,17 @@ func killWordLeft(ed *Editor) {
 	ed.dot = space
 }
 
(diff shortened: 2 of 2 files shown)


xiaq:
Hmm, what about `@:;|`? They can all function as separators in certain circumstances :)

I think that instead of using a more or less arbitrary set of separators, we can define a small word as a run of characters that are either all alphanumeric, or all non-alphanumeric, with optional trailing whitespaces. For instance, `scp ~/** user@host:dir` contains small words `scp ` (with a trailing space), `~/** ` (with a trailing space), `user`, `@`, `host`, `:`and `dir`.

zhsj:
http://unix.stackexchange.com/questions/78990/what-are-the-readline-word-separators

it suggests bash has two types of `word`. One is alphanum, and another is separated by shell meta-characters.

So which is better? 

Besides, why you think `~/**` should be one word, instead of `~` and `**`

xiaq:
I find the word movement and killing behaviors in bash and zsh pretty complicated, so I want to make it easier to understand in elvish. Also, as I just looked up vim's manpage, this also happens to be vi's definition of words, except that vi considers `_` to be a alphanumeric.

`~/**` being one word is not very nice. However, as I imagine, most of the time you want to move past or remove a word, it is either something separated by space (which is how a word is defined now), or a long sequence of letters and perhaps numbers. Clusters of symbols like `~/**` are usually short, and in that case the normal (by-character) movement and killing commands work well enough.

xiaq:
Rejecting in favor of e2f12a6ab1e69f257fbe52d09f54000f18c5d201.

PR #177, opened 2016-04-11
add editorconfig: uniform position to place config for editor

huiyiqun:


xiaq:
gofmt takes care the formatting, so no editor configuration is needed.

PR #210, opened 2016-09-13
README: macOS homebrew

gaoyifan:
https://github.com/Homebrew/homebrew-core/pull/4724

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,12 @@ done
 
 [How To Write Go Code](http://golang.org/doc/code.html) explains how `$GOPATH` works.
 
+For macOS users, you can build elvish via [homebrew](http://brew.sh):
+
+```sh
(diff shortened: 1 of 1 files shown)


zmwangx:
Homebrew isn't a service for prebuilt binaries (although we do offer optional prebuilt binaries in certain cases), FYI.

gaoyifan:
maybe we should use homebrew as plan B....

zmwangx:
I'm all for adding Homebrew instructions to README (even a Homebrew badge for people who like that of stuff ), just saying it shouldn't be under the section "prebuilt binaries".

xiaq:
Please move the instruction to the "Building Elvish" section and make it clear that it will build the last release version of elvish (as opposed to HEAD).

PR #248, opened 2016-10-10
honor LS_COLORS env

### LINKED ISSUE #237: honor LS_COLORS env
Currently `edit/ls-colors.go` has default color. We should honor LS_COLORS env

zhsj:
this fix #237


xiaq:
The result of parsing `$E:LS_COLOR` should be cached. In this implementation, it is parsed every time a candidate gets written; if the pwd has 100 files, it will be parsed 100 times.

Also, I have manually stripped all `0` and `00` entries present in the default output of `dircolors` (e.g. `rs=0` and `mh=00`; see https://github.com/elves/elvish/commit/3e7ed66964996f7b94adaa0d3134b00810e36cfe#diff-0f54cd9217009b5e4ae70d47a84ee585L71) to avoid messing up with the default style of completio listing. This should be done in the program as well.

edit/ls-colors.go
```diff
@@ -82,6 +92,17 @@ func parseLsColor(s string) *lsColor {
 			continue
 		}
 		key, value := words[0], words[1]
+		filterValues := []string{}
+		for _, splitValue := range strings.Split(value, ";") {
+			if splitValue == "0" || splitValue == "00" || splitValue == "000" {
```
xiaq: You should also check for "" here, which is equivalent to "0". I think it makes sense to use `strings.Count(s, "0") == len(s)`.

xiaq:
This implementation does much less parsing, but still not as little as possible. You can save the last result of `getLsColor` along with the value of `$E:LS_COLOR` from which it was parsed, and only re-parse when `$E:LS_COLOR` gets changed. Use a mutex to make the caching concurrency-safe.

PR #300, opened 2016-12-27
README.md: Add description for RPM package

sztsian:
Signed-off-by: Zamir SUN <zsun@fedoraproject.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,16 @@ Try `cn.dl.elvish.io` if you are in China and the download is too slow. (In the 
 
 See also [Building Elvish](#building-elvish).
 
+### RPM package
+
+RPM Package for Fedora is available in [FZUG Repo](https://github.com/FZUG/repo/wiki/Add-FZUG-Repository). You can install it by add FZUG repo first:
+
+    dnf config-manager --add-repo=http://repo.fdzh.org/FZUG/FZUG.repo
+
+Then install elvish
+
+    dnf install elvish
+
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


zhsj:
Why the package name is golang-github-elves-elvish? Such name pattern is used for dev package in Debian, same Fedora?

For example,
docker is not named golang-github-docker-docker

sztsian:
> Why the package name is golang-github-elves-elvish? Such name pattern is used for dev package in Debian, same Fedora?

Ah, great catch. We misunderstood the naming guideline. Previously I think all the go package should follow the golang-website-developer-app style. As the Packaging guideline for Go apps in Fedora is delivered from Debian, so I guess you are right. I'll fix this in the repo soon.

Please keep this PR, I'll update to this PR directly after it's fixed.

sztsian:
Already fixed in repo. CI will finish the package soon. See https://github.com/FZUG/repo/pull/209

README.md
```diff
@@ -26,6 +26,16 @@ Try `cn.dl.elvish.io` if you are in China and the download is too slow. (In the
 
 See also [Building Elvish](#building-elvish).
 
+### RPM package
+
+RPM Package for Fedora is available in [FZUG Repo](https://github.com/FZUG/repo/wiki/Add-FZUG-Repository). You can install it by add FZUG repo first:
```
xiaq: **the** FZUG repo; by **adding**.

I think it is better to just say: "First add the FZUG repo", since there is no doubt that you are giving instructions on how to install elvish; the syntax is also consistent with "Then install elvish",

PR #301, opened 2017-01-11
Added styles type, which translates human-readable values to ANSI codes

muesli:
Added a translation table, so that we / users can specify human-readable color values and style settings (like "bold", "italic").

I've tried to keep it simple and not make the 'styles' struct keep a private array of style settings. That way, you can simply append to it and operate on it as you would with any other string slice. That in turn means, that we can't guarantee any values in that array being pre-translated before output. The styles.String() method compiles all set styles, translates them and returns a semicolon-concatenated string, which gets used by 'styled'.

There's an argument to be had, that pre-translating styles upon appending them would be faster if you re-use the 'styled' struct several times. I don't see this happening in code, though, so - as mentioned - I opted for simplicity.

### DIFF
--- a/edit/buffer.go
+++ b/edit/buffer.go
@@ -134,7 +134,7 @@ func (b *buffer) writes(s string, style string) {
 }
 
 func (b *buffer) writeStyled(s *styled) {
-	b.writes(s.text, s.style)
+	b.writes(s.text, s.styles.String())
 }
 
 func (b *buffer) writeStyleds(ss []*styled) {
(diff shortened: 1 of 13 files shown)


xiaq:
Thanks for the PR! In the future, it would be nice if you can add unit tests against your code though.

Fine print: Personally I am sometimes too lazy to write tests, which is why I accepted your PR anyway. But it would be awesome to have tests.

muesli:
Happy to still write / adapt those.

PR #303, opened 2017-01-11
Updated .gitignore

muesli:
Updated .gitignore to include elvish binary and common editor backup files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,10 @@ _testmain.go
 
 *.exe
 
+# Backup files
+*~
+
 # Project specific
 /t/ # Small applications for manual testing go inside this directory
 /cover/
+elvish
--- a/.gitignore
+++ b/.gitignore
@@ -21,10 +21,7 @@ _testmain.go
 
 *.exe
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -21,6 +21,10 @@ _testmain.go
 
 *.exe
 
+# Backup files
+*~
+
```
xiaq: I think ignores that should apply to all repositories belong in `~/.gitignore`; it is why I am not including vim swap files in this .gitignore.

.gitignore
```diff
@@ -21,6 +21,10 @@ _testmain.go
...
 # Project specific
 /t/ # Small applications for manual testing go inside this directory
 /cover/
+elvish
```
xiaq: /elvish

.gitignore
muesli: I'm aware of that. The problem is though, that you can't rely on everybody else's setup to be correct - hence why I personally tend to keep backup and swap files in a repo's .gitignore. Just my 2 cents, will remove it for now.

PR #310, opened 2017-01-13
Completion style

muesli:
Made remaining strings proper styles.

Slightly changed completion colors for consistency with common HIGs: bright background, darker text means current selection. I hope I'm not stepping on anyone's toe (and favourite color values) here. Eventually this should be user-configurable/-themeable - but I think sane default values / common practices are always an improvement.


xiaq:
Hmm, I think you need to banish the commit that fixes bg jobs from this branch. Also, this seems seems not up-to-date, since it uses the old underscored style name `bg_white` instead of `bg-white`.

muesli:
Whoops, must have created this branch from my other feature branch by accident. Will fix.

Issue #184, opened 2016-04-19
Give listings a "dropdown menu" feeling


xiaq:
This is inspired by ptpython, whose completion menu looks better than that of elvish (which is modeled after zsh):

![ptpython dropdown menu](https://raw.githubusercontent.com/jonathanslenders/ptpython/master/docs/images/file-completion.png)

The "dropdown" feeling can be attributed to two factors: a different background color, and margins on the sides.

krader1961:
FWIW, I didn't know about [ptpython](https://github.com/prompt-toolkit/ptpython) until I looked at that project in response to reading this issue. So thanks for bringing that project to my attention. Nonetheless, I don't think this issue has sufficient information to drive a change in the existing elvish behavior. Especially since four years have elapsed since this issue was opened and a lot has changed in that interval. The example in the problem statement produces decent, if different, output in a current elvish version. Could the output be better? Probably, but who gets to decide which is better? At the end of the day a lot of issues like this one require reasonable defaults plus the ability of the user to tweak it to suit their needs. Which seems to be loosely related to issue #1001.

PR #311, opened 2017-01-13
Completion style

muesli:
Made remaining strings proper styles.

Slightly changed completion colors for consistency with common HIGs: bright background, darker text means current selection. I hope I'm not stepping on anyone's toe (and favourite color values) here. Eventually this should be user-configurable/-themeable - but I think sane default values / common practices are always an improvement.

### DIFF
--- a/edit/style.go
+++ b/edit/style.go
@@ -16,10 +16,10 @@ var (
 	styleForScrollBarThumb   = styles{"magenta", "inverse"}
 	styleForSideArrow        = styles{"inverse"}
 
-	// Use white text on black background for completion listing.
-	styleForCompletion = styles{"white", "bg-black"}
-	// Use black text on white background for selected completion.
-	styleForSelectedCompletion = styles{"black", "bg-white"}
+	// Use default style for completion listing
+	styleForCompletion = styles{}
+	// Use inverse style for selected completion entry
+	styleForSelectedCompletion = styles{"inverse"}
 )
 
 var styleForType = map[TokenKind]styles{
--- a/edit/style.go
+++ b/edit/style.go
@@ -16,10 +16,10 @@ var (
 	styleForScrollBarThumb   = styles{"magenta", "inverse"}
 	styleForSideArrow        = styles{"inverse"}
 
-	// Use black text on white for completion listing.
-	styleForCompletion = styles{}
-	// Use white text on black for selected completion.
-	styleForSelectedCompletion = styles{"black", "bg-white", "bold"}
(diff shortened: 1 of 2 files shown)


edit/style.go
```diff
@@ -7,19 +7,19 @@ var (
...
 	styleForScrollBarArea    = styles{"magenta"}
 	styleForScrollBarThumb   = styles{"magenta", "inverse"}
 	styleForSideArrow        = styles{"inverse"}
 
 	// Use black text on white for completion listing.
-	styleForCompletion = styles{"black", "bg_white"}
+	styleForCompletion = styles{}
 	// Use white text on black for selected completion.
-	styleForSelectedCompletion = "inverse"
+	styleForSelectedCompletion = styles{"black", "bg-white", "bold"}
```
xiaq: This is assuming that the default is white-on-black, which may not be the case (e.g. macOS's terminal). Simply use `"inverse"` to agree with `styleForSelected`. I prefer to leave out `bold` as well to agree with it.

The decision to use black-on-white for completion and white-on-black for selected completion was said in #184, but it doesn't work well with ls colors, so it makes sense to revert it.

xiaq: Also, update the comments.

muesli: I see your point, even though I'd argue that 99% of terminals still have a black background - even if OSX defaults to black on white, personally I haven't seen anyone managing to cope with that for more than 5 minutes - especially power users who dare to switch their shell.

I'd suggest also setting a sane default for 'styleForCompletion' then. My initial decision to get rid of inversion was indeed to due the clash with ls colors.

edit/style.go
```diff
@@ -7,19 +7,19 @@ var (
...
 
-	// Use black text on white for completion listing.
-	styleForCompletion = styles{"black", "bg_white"}
-	// Use white text on black for selected completion.
-	styleForSelectedCompletion = "inverse"
+	// Use white text on black background for completion listing.
+	styleForCompletion = styles{"white", "bg-black"}
+	// Use black text on white background for selected completion.
+	styleForSelectedCompletion = styles{"black", "bg-white"}
 )
```
xiaq: Well, my intention was to use no style for `styleForCompletion`, and reverse video for selected completion, to agree with other listings. Sorry if this wasn't clear. :(

muesli: Ok, understood. I guess eventually those styles should be user-configurable anyway. Unless they already are and I've just missed it?

xiaq: Yes, they should be. No, they are not yet. :(

PR #312, opened 2017-01-14
Added Editor.NotifyOnce

### LINKED ISSUE #296: "End of history" turns up reaptly when down arrow is pressed
When the down arrow is pressed in a new line, "End of history" continually turns up.
Too much of this error is annoying, can it be hidden after new content is typed or up arrow is pressed?
Thanks.

muesli:
NotifyOnce can be used to make sure we're not spamming the terminal
with dupe notifications. "End of history" notification is using
this now. Fixes #296

### DIFF
--- a/edit/builtin-fn.go
+++ b/edit/builtin-fn.go
@@ -239,5 +239,5 @@ func redraw(ed *Editor) {
 }
 
 func endOfHistory(ed *Editor) {
-	ed.Notify("End of history")
+	ed.NotifyOnce("End of history")
 }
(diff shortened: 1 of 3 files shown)


xiaq:
If the user is troubled by the "end of history" message, s/he can just bind Down to a nop in Insert mode: `le:binding[Insert][Down]={ }`. Note that Down in history mode will still work, no functionality lost.

I believe NotifyOnce will create a grey area between notifications and non-notifications and is prone to abuse. My stand is: if something is notification-worthy, then it should notify the user each time, regardless of whether the last one was the same. If not, then elvish should be silent. So I am closing this PR - sorry :)

PR #313, opened 2017-01-14
Allow histlist to be filtered case-insensitive

muesli:
You can press Ctrl+F to toggle case-sensitivity mode.

### DIFF
--- a/edit/builtin-fn.go
+++ b/edit/builtin-fn.go
@@ -99,6 +99,7 @@ var builtins = []*BuiltinFn{
 	// History listing mode
 	{"start-histlist", startHistlist},
 	{"histlist-toggle-dedup", histlistToggleDedup},
+	{"histlist-toggle-case-sensitivity", histlistToggleCaseSensitivity},
 
 	// Bang mode
 	{"start-bang", startBang},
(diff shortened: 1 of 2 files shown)


xiaq:
I like this. Even though it is currently only for histlist mode, I think the natural plan is to support this in other modes as well. In completion mode, Ctrl-F is used for toggle **F**iltering, so we should use another key for consistency's sake.

Unfortunately we cannot use Ctrl-C because it sends SIGINT, nor Ctrl-I because it is the same as Tab. Ideas? It would be best if it can be justified in an interesting way. (Most keybindings in elvish are justified in one way or another, Ctrl-F coming from GUI applications, and Ctrl-L from browsers, for instance.)

xiaq:
Hmm, the next-next level is to cycle through different matching algorithms (my favorite being subsequence matching where `am` matches **a**lgorith**m**), so we might come up with a key along that line.

Perhaps `Ctrl-M` for **m**atcher is a good one...

muesli:
Isn't Ctrl-M usually [return]?

xiaq:
Stupid me.

BTW, the grand picture of user-programmable filterer (what I called matcher in the last comment -- let's use filterer from now) is this project: https://github.com/elves/elvish/projects/1

Issue #325, opened 2017-02-18
kill-path-component


aeosynth:
another nice feature from fish is the `backward-kill-path-component` function, bound to `^W`, which kills everything left of the cursor, to the next `/` or whitespace. or maybe i should implement it in my own config? or maybe it's not needed if i use the navigation mode.

are you interested in taking more features from fish? i was mostly happy with it as my default shell.

xiaq:
Does Ctrl-Left (or Alt-B) and Ctrl-Right (or Alt-F) work in the same way in fish?

aeosynth:
fish treats those sequences similarly to bash, similar to vim `b` and `w`

xiaq:
See #333.

PR #330, opened 2017-02-24
killDirLeft

aeosynth:


aeosynth:
fixes #325. probably names are bad. this is my first published go code.

xiaq:
Hi, thank you for the pull request, but I am not going to merge it for the following reasons:

1. This makes the behavior of ^W inconsistent with ^Left and ^Right.

2. This approach is not customizable. What if someone wants ^W to stop at `-` instead of `/`?

A good way to implement this is to let the user specify the set of word separators (currently it is defined by `unicode.IsSpace`). I haven't got around to implementing this though.

As for the naming per se, `kill-path-component-left` is a better name IMO.

aeosynth:
would you want one or many sets of word seperators? eg should it be possible for ^W and ^Left  to behave differently?

xiaq:
I don't like the idea for ^W and ^Left to use different sets of separators. The idea is that they both operate on "words", and they should agree on what a word is. It's easy to explain them to users this way.

This is, however, only about what I would ship in the default setting. On the other hand, I will support customization in the aspect. I opened #333.

PR #347, opened 2017-05-11
Seperated editor IO

xchenan:
In the code on master branch, both input and output use Editor.file. As such, editor writing to fd0 always happens when elvish is running. We seperate IO so as to avoid such scenario. 

Also, please note that prompts write to os.Stderr rather than os.Stdout (Line 125 in shell/shell.go)

Signed-off-by: xchenan <xchenan@gmail.com>


xiaq:
> As such, editor writing to fd0 always happens when elvish is running.

The editor is only initialized when fd0 is a tty (https://github.com/elves/elvish/blob/master/shell/shell.go#L58), and it is only intended to be used in such scenarios. As such, I don't see benefits in separating the input and output; the editor assumes that it is talking to a terminal that is both input and output.

rminnich:
Writing to fd 0 runs against the grain of about 45 years of unix usage :-)
While you have found in your usage that it always works, unfortunately we've found a case where it fails (u-root.tk), i.e. running linux from FLASH and then running elvish directly from a go-based init. 
Further, prompts should not necessarily go to the same file as stdin/stdout, they belong on stderr, historically.

The problem we were seeing is that prompts never appear. The fix in this PR resolves the problem. it's also a bit more correct IMHO: input is fd 0, output fd 1, info/err on fd 2; there's an historical reason why they are separate. Were you to adapt this change, by making input be an io.Reader and output an io.Writer., it would be easier to catch possible mistakes from others  where input might be incorrectly written to output, and vice-versa.

xiaq:
@rminnich That is a very interesting case, I wasn't aware of that. I still have several doubts though:

* I this PR the standard error is being used as the output of the editor; that also sounds like an "un-Unixy" thing. Wouldn't it be more correct if all 3 fds are separate? But if you were to separate all, how do you decide when to use stdout and when to use stderr? For instance, does the current command text go to stdout or stderr?

* What about the tty ioctl's? This PR uses stdin, is that an arbitrary choice?

* Last but not least, why does u-root.tk expect prompts to be written to stderr? Is that an oddity of u-root.tk, or is it some kind of established Unix tradition?

If you see it in another way, the reason I am now combining I/O in one fd is to avoid choosing one in ambiguous situations, so I'd like see justifications for the choices made here.

rminnich:
I think it would be best if you could follow the existing conventions of standard shells. Those conventions are:
input on fd 0
output from builtins (cd, echo, etc.) on fd 1 so they can be piped
prompt and completions on fd 2
errors, etc. on fd 2

It's really confusing right now to follow elvish with, e.g., strace, because it's doing things that just look wrong, e.g. writing to fd 0, which has been stdin since the days of core memory :-)

the stty stuff has always used fd 0 by convention, that goes back to 1970 ...

So, I'd most prefer it if we could adopt the standard conventions. 

Also, u-root is not really doing anything special. We've run all sorts of shells in that environment, and the only one that had the disappearing prompts problem was elvish. I think what's going on is we are putting elvish into an environment that's not quite what it's used to, and it's not working well.

Can you tell me more about the ambiguous situation? I am pretty sure there should be no ambiguity.

It can be useful to do something like strace -e trace=write,read /bash to see common usage. bash, for examples, does all its editing mode stuff to fd 2, its builtins go to fd 1, and input is from fd 0. I think it's a good idea to follow the common shell conventions.

I think elvish is terrific and we plan to use it for our shell on u-root -- but we really need the fix for fd 2 usage. Thanks again.

xiaq:
@rminnich Thank you. You just answered my questions on ambiguity.

PR #359, opened 2017-06-18
Fix list slice index check

xofyarg:
Given a list slice `li[i:j]`, the parser will crash when `i > j`. Add
a check to raise IndexOutOfRange.

Also add some tests.

### DIFF
--- a/eval/list_test.go
+++ b/eval/list_test.go
@@ -1,0 +1,140 @@
+package eval
+
+import (
+	"testing"
+
+	"github.com/elves/elvish/util"
+)
+
+var parseAndFixListIndexTests = []struct {
+	name string
+	// input
+	expr string
+	len  int
+	// output
+	shouldPanic, isSlice bool
+	begin, end           int
+}{
+	{
+		name: "stringIndex",
+		expr: "a", len: 0,
+		shouldPanic: true,
+	},
+	{
+		name: "floatIndex",
+		expr: "1.0", len: 0,
+		shouldPanic: true,
+	},
+	{
+		name: "emptyZeroIndex",
+		expr: "0", len: 0,
+		shouldPanic: true,
+	},
+	{
+		name: "emptyPosIndex",
+		expr: "1", len: 0,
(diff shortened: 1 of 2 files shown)


eval/list_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+func TestParseAndFixListIndex(t *testing.T) {
+	check := func(t *testing.T, name, value string, want, got interface{}) {
+		if want != got {
+			t.Errorf("%s value: [%s] want: [%v] got: [%v]",
+				name, value, want, got)
+		}
+	}
+
+	for _, item := range []struct {
```
xiaq: Can you move the definition of test data out of the function into a variable? Namely

```
var parseAndFixListIndexTests = ...
```

eval/list_test.go
```diff
@@ -0,0 +1,133 @@
+package eval
+
+import "testing"
+
+func TestParseAndFixListIndex(t *testing.T) {
+	check := func(t *testing.T, name, value string, want, got interface{}) {
```
xiaq: You can remove the `t` parameter as `check` is a closure. Consider renaming this function to `checkEqual`.

eval/list_test.go
```diff
@@ -0,0 +1,133 @@
...
+			expr: "1:1", len: 2,
+			isSlice: true, begin: 1, end: 1,
+		},
+		{
+			name: "sliceBeginAboveEnd",
+			expr: "1:0", len: 2,
+			shouldPanic: true,
+		},
+	} {
+		func(t *testing.T) {
```
xiaq: Instead of wrapping the whole body in a function, please use `util.PCall`:

```
var isSlice bool
var begin, end int
err := util.PCall(func() {
    isSlice, begin, end = ParseAndFixListIndex(...)
}) != nil
// Check isSlice, begin, end and err
```

xofyarg:
Amended to address all the concerns(so far). Rebased onto current master.

PR #361, opened 2017-06-19
Ignore OK when printing PipelineError

xofyarg:
The program crashes when trying to print PipelineError if there's an
OK inside since its content is barely empty. Simple ignore the OK
value will fix it.

Or maybe we should print that out as well to better align with the
pipeline?


xiaq:
Thanks!

No, there is no need to print the OK. If a command in a pipeline is OK, it is not responsible for the failure of the pipeline.

PR #362, opened 2017-06-20
Add a editor builtin command to compose complex candidate

xofyarg:
External commands and/or functions could leverage the complex
candidate to gain more control over completion.


xiaq:
The name `le:output-complex-candidate` is a bit too long. It can just be called `le:complex-candidate`.

xiaq:
Also, please add a todo comment for adding a `style` option once it is possible to construct `ui.Styles` from the shell.

xofyarg:
Name changed from `le:output-complex-candidate` to `le:complex-candidate`. `style` has been added as an option as well.

PR #365, opened 2017-06-22
Expose bindings for loc and bang mode

xofyarg:
Add keys of those two modes into default readline bindings.


scateu:
剑外忽传收蓟北，初闻涕泪满衣裳。

PR #367, opened 2017-06-25
Fix and cleanup.

xofyarg:
Add a PoC mode called "narrow", which tries to be a generic purpose list selection mode that can be easily extended.

Also includes fixup for "bang" to "lastcmd" renaming and register merging.

### DIFF
--- a/edit/registry.go
+++ b/edit/registry.go
@@ -19,7 +19,9 @@ var builtinMaps = map[string]map[string]*BuiltinFn{}
 // variable initializations to make sure every subnamespace is registered before
 // makeBindings is ever called.
 func registerBuiltins(module string, impls map[string]func(*Editor)) struct{} {
-	builtinMaps[module] = make(map[string]*BuiltinFn)
+	if _, ok := builtinMaps[module]; !ok {
+		builtinMaps[module] = make(map[string]*BuiltinFn)
+	}
 	for name, impl := range impls {
 		var fullName string
 		if module == "" {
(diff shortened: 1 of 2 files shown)


edit/narrow.go
```diff
@@ -0,0 +1,344 @@
+package edit
```
xiaq: We are discussing this on Telegram.

edit/api.go
```diff
@@ -281,3 +284,73 @@ func outputComplexCandidate(ec *eval.EvalCtx, a []eval.Value, o map[string]eval.
...
+		case eval.String:
+			return narrowStringItem(string(raw))
+		case eval.Map:
+		default:
+		}
+		return nil
+	}
+
+	switch rawSource := source.(type) {
+	case eval.CallableValue:
```
xofyarg: The idea behind it is to support both static source and dynamic source. The latter allows itself to be called whenever the filter changes.

edit/api.go
```diff
@@ -281,3 +284,73 @@ func outputComplexCandidate(ec *eval.EvalCtx, a []eval.Value, o map[string]eval.
...
+		nr.source = narrowSourceFunc(func() []narrowItem {
+			return nis
+		})
+	default:
+		return
+	}
+
+	nr.action = func(ed *Editor, item narrowItem) {
+		switch rawAction := action.(type) {
+		case eval.String:
```
xofyarg: Correct, it is supposed to be a TODO.

xofyarg:
Per conversation on telegram, will rework the "narrow mode" by extending listing mode. Removed that commit from this PR.

PR #372, opened 2017-06-29
Add a builtin function "builtin"

xofyarg:
To prevent being stuck into deadloop when shadowing builtin functions,
elvish needs to be told explicitly. The function calls its arguments
in turn except shows all the builtins when no argument is given.


xiaq:
The former functionality is already provided by the `builtin:` namespace. For instance, to call builtin `put`, you can use `builtin:put`.

The latter functionality is perhaps better designed as part of a general namespace introspection primitive.

PR #373, opened 2017-06-30
Fix process hangs when daemon starts

xofyarg:
The parent hangs on syscall When creating daemon from client. Instead,
use highlevel interface to bring daemon up.

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -24,9 +24,6 @@ type Daemon struct {
 	LogPathPrefix string
 }
 
-// closeFd is used in syscall.ProcAttr.Files to signify closing a fd.
-const closeFd = ^uintptr(0)
-
 // Main is the entry point of the daemon sub-program.
 func (d *Daemon) Main(serve func(string, string)) int {
 	switch d.Forked {
@@ -56,20 +53,15 @@ func (d *Daemon) Main(serve func(string, string)) int {
 
 		syscall.Umask(0077)
 		return d.pseudoFork(
-			&syscall.ProcAttr{
+			&os.ProcAttr{
 				// cd to /
 				Dir: "/",
 				// empty environment
 				Env: nil,
-				// inherit stderr only for logging
-				Files: []uintptr{closeFd, closeFd, 2},
-				Sys:   &syscall.SysProcAttr{Setsid: true},
+				Sys: &syscall.SysProcAttr{Setsid: true},
 			})
 	case 1:
-		return d.pseudoFork(
-			&syscall.ProcAttr{
-				Files: []uintptr{closeFd, closeFd, 2},
-			})
+		return d.pseudoFork(&os.ProcAttr{})
 	case 2:
 		serve(d.SockPath, d.DbPath)
 		return 0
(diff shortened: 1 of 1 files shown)


daemon/daemon.go
```diff
@@ -82,21 +79,22 @@ func (d *Daemon) Main(serve func(string, string)) int {
...
-			binPath = os.Args[0]
-		} else {
-			// Find elvish in PATH
-			paths := strings.Split(os.Getenv("PATH"), ":")
-			result, err := util.Search(paths, "elvish")
-			if err != nil {
-				return errors.New("cannot find elvish: " + err.Error())
-			}
-			binPath = result
+	if !filepath.IsAbs(binPath) {
```
xiaq: I feel that it is better to use the homegrown searching utility (`util.Search`) in place of `exec.LookPath` in consistency with the rest of Elvish.

daemon/daemon.go
```diff
@@ -82,21 +79,22 @@ func (d *Daemon) Main(serve func(string, string)) int {
...
+		var err error
+		binPath, err = exec.LookPath(os.Args[0])
```
xiaq: The original code has the upside of not crashing when size of `os.Args` is 0 :).

Also, when Elvish is used as a login shell, `os.Args[0]` contains `-elvish`, and this code will search `-elvish` instead of `elvish`.

I think the original code is fine here...

daemon/daemon.go
```diff
@@ -82,21 +79,22 @@ func (d *Daemon) Main(serve func(string, string)) int {
...
-	return forkExec(nil, 0, binPath, d.DbPath, d.SockPath, d.LogPathPrefix)
+
+	cmd := exec.Command(binPath, "-daemon",
+		"-forked", "0",
+		"-bin", binPath,
+		"-db", d.DbPath,
+		"-sock", d.SockPath,
+		"-logprefix", d.LogPathPrefix,
+	)
+	return cmd.Run()
```
xiaq: I think it's better to understand what is happening under the scene than using a different interface to do the job :) According to the [doc](http://godoc.org/os/exec#Cmd.Run) `cmd.Run` waits for the subprocess to terminate, that is probably some step we should add.

xofyarg: > I think it's better to understand what is happening under the scene than using a different interface to do the job :)

Agreed. The point here is, trying not to re-invent the wheel by using high level API when possible since it is not recommended to use "syscall" package directly according to its doc.

A quick gdb showed the daemon was stuck during the syscall, which remind me to check the flags done internally by `exec.Command`.

```
(gdb) bt
#0  syscall.Syscall () at /usr/lib/go-1.8/src/syscall/asm_linux_amd64.s:27
#1  0x000000000047e516 in syscall.readlen (fd=1, p=0xc420125ae0 "", np=8, n=13, err=...) at /usr/lib/go-1.8/src/syscall/zsyscall_linux_amd64.go:1085
#2  0x0000000000478313 in syscall.forkExec (argv0=..., argv=..., attr=0xc420125da0, pid=10, err=...) at /usr/lib/go-1.8/src/syscall/exec_unix.go:202
#3  0x0000000000478754 in syscall.ForkExec (argv0=..., argv=..., attr=0xc420125da0, pid=140729429880769, err=...)
    at /usr/lib/go-1.8/src/syscall/exec_unix.go:235
#4  0x0000000000541bfa in github.com/elves/elvish/daemon.forkExec (attr=0xc420125da0, forkLevel=1, binPath=..., dbPath=..., sockPath=..., logPathPrefix=...,
    ~r6=...) at /home/go/src/github.com/elves/elvish/daemon/daemon.go:120
#5  0x0000000000541a78 in github.com/elves/elvish/daemon.(*Daemon).pseudoFork (d=0xc4201145f0, attr=0xc420125da0, ~r1=842351593000)
    at /home/go/src/github.com/elves/elvish/.build/src/github.com/elves/elvish/daemon/daemon.go:104
#6  0x000000000054173a in github.com/elves/elvish/daemon.(*Daemon).Main (d=0xc4201145f0, serve={void (struct string, struct string)} 0xc420125e48,
    ~r1=14689824) at /home/go/src/github.com/elves/elvish/daemon/daemon.go:67
#7  0x0000000000816869 in main.main () at /home/go/src/github.com/elves/elvish/main.go:121
```

xiaq: It's probably best to use the `os/exec` interface to implement `forkExec`.

xofyarg: Deal. Using `os.StartProcess` instead to fully close stdin/stdout/stderr.

daemon/daemon.go
```diff
@@ -96,27 +88,38 @@ func (d *Daemon) Spawn() error {
 			binPath = result
 		}
 	}
-	return forkExec(nil, 0, binPath, d.DbPath, d.SockPath, d.LogPathPrefix)
-}
-
-// pseudoFork forks a daemon. It is supposed to be called from the daemon.
-func (d *Daemon) pseudoFork(attr *syscall.ProcAttr) int {
-	err := forkExec(attr, d.Forked+1, d.BinPath, d.DbPath, d.SockPath, d.LogPathPrefix)
+	p, err := os.StartProcess(binPath, []string{
```
xiaq: This is repeated below; please make it a function.

PR #374, opened 2017-07-02
Rework listing mode

xofyarg:
Generic functionalities are merged into listing mode, making it into a
widget with an interface to be extended from outside of the shell
core. histlist, location and lastcmd are re-implemented in an embedded
module, using the method mentioned above.

### DIFF
--- a/edit/api.go
+++ b/edit/api.go
@@ -62,7 +62,7 @@ func installModules(modules map[string]eval.Namespace, ed *Editor) {
 	binding := &eval.Struct{
 		[]string{
 			modeInsert, modeCommand, modeCompletion, modeNavigation, modeHistory,
-			modeHistoryListing, modeLocation, modeLastCmd, modeListing},
+			modeHistoryListing, modeLocation, modeLastCmd, modeListing, modeNarrow},
 		[]eval.Variable{
 			eval.NewRoVariable(BindingTable{keyBindings[modeInsert]}),
 			eval.NewRoVariable(BindingTable{keyBindings[modeCommand]}),
@@ -73,6 +73,7 @@ func installModules(modules map[string]eval.Namespace, ed *Editor) {
 			eval.NewRoVariable(BindingTable{keyBindings[modeLocation]}),
 			eval.NewRoVariable(BindingTable{keyBindings[modeLastCmd]}),
 			eval.NewRoVariable(BindingTable{keyBindings[modeListing]}),
+			eval.NewRoVariable(BindingTable{keyBindings[modeNarrow]}),
 		},
 	}
 	ns["binding"] = eval.NewRoVariable(binding)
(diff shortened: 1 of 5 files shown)


edit/listing.go
```diff
@@ -343,3 +384,201 @@ func getListing(ed *Editor) *listing {
...
+}
+
+type listingItemComplex struct {
+	eval.Map
+}
+
+func (c *listingItemComplex) Content() string {
+	key := eval.String("content")
+	if c.Map.HasKey(key) {
+		return c.Map.IndexOne(key).(eval.String).String()
```
xiaq: This will panic if the user supplied a map whose `content` is not a string.

Consider making `listingItemComplex` a struct of fields `content`, `display`, etc.. You can use a `*string` as their types and use `nil` to signal nonexistence.

edit/listing.go
```diff
@@ -343,3 +384,201 @@ func getListing(ed *Editor) *listing {
...
+	for i := start; i < end; i++ {
+
+		out <- eval.NewMap(map[eval.Value]eval.Value{
+			eval.String("id"):  eval.String(strconv.Itoa(i)),
+			eval.String("cmd"): eval.String(cmds[i]),
+		})
+	}
+}
+
+func InputInsert(ec *eval.EvalCtx, args []eval.Value, opts map[string]eval.Value) {
```
xiaq: Hmm, why not call this `insert-at-dot`?

edit/listing.go
```diff
@@ -343,3 +384,201 @@ func getListing(ed *Editor) *listing {
...
+	var text eval.String
+
+	eval.ScanArgs(args, &text)
+	eval.TakeNoOpt(opts)
+
+	ed := ec.Editor.(*Editor)
+	ed.insertAtDot(text.String())
+}
+
+func InputReplace(ec *eval.EvalCtx, args []eval.Value, opts map[string]eval.Value) {
```
xiaq: `replace-input` seems a bit nicer...

edit/listing.go
```diff
@@ -343,3 +384,201 @@ func getListing(ed *Editor) *listing {
...
+	ignoreDup  bool
+	ignoreCase bool
+}
+
+type listingItemString struct {
+	eval.String
+}
+
+func (s *listingItemString) Content() string {
+	return s.String.String()
```
xofyarg: Did you mean the representation of `eval.String.String()` could be different from its internal string? Since I don't see much saving from `s.String.String()` to `string(s.String)`.

edit/listing.go
xofyarg: The idea behind using eval.Value is to reduce object conversion when interact with elv environment. I think we could add some validations here.

eval/listing.elv
```diff
@@ -0,0 +1,53 @@
+fn location {
+    candidates = [(dirs | each {
```
xofyarg: Did you mean to rename the builtin function?

eval/listing.elv
```diff
@@ -0,0 +1,53 @@
...
+	score = (splits $args[0][score] &sep=. | take 1)
```
xofyarg: Looks stupid but...

eval/listing.elv
```diff
@@ -0,0 +1,53 @@
...
+        put [
+            &content=$args[0][path]
+            &display=$score" "$args[0][path]
```
xofyarg: My thinking was to let user control the display setting while listing will in charge of drawing.

edit/listing.go
xofyarg: Good catch.

edit/listing.go
xofyarg: Changed.

eval/listing.elv
xiaq: Yes.

xiaq:
More on loading a elv module when starting editor: later we can also move the default keybindings there.

xofyarg:
> It is no longer possible to have submode-specific keybindings. ... My suggestion is to preserve edit:binding[histlist loc lastcmd], and teach edit:listing-read to accept extra key bindings. When starting (e.g.) histlist pass $edit:binding[histlist] to it.

I'd prefer to let user providing an anonymous binding map which could override default binding temporarily. It's a more dynamic way to archive.

xofyarg:
Separated as a new mode and disabled by default.

PR #382, opened 2017-07-04
Fix FreeBSD build

myfreeweb:
FdSet.Bits is very weird here for some reason

https://golang.org/src/syscall/ztypes_freebsd_amd64.go

### DIFF
--- a/sys/select_32.go
+++ b/sys/select_32.go
@@ -1,4 +1,4 @@
-// +build darwin 386,freebsd arm,freebsd 386,linux arm,linux netbsd openbsd
+// +build darwin 386,linux arm,linux netbsd openbsd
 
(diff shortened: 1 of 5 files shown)


xiaq:
Thanks for the fix. The inconsistency is horrible.

It seems that we can simply use the `Select` implementation from golang.org/x/sys/unix (which is already a vendored dependency), have you looked at that? https://godoc.org/golang.org/x/sys/unix#Select

myfreeweb:
Looks like it's the same thing https://github.com/golang/sys/blob/94b76065f2d2081d0fef24a6e67c571f51a6408a/unix/ztypes_freebsd_amd64.go#L289

xiaq:
You are right, merging.

Issue #391, opened 2017-07-06
Elvish Dies with Specific Key Sequence


padietl:
`sh-4.3$ elvish
~> a = q =

runtime error: invalid memory address or nil pointer dereference
goroutine 1 [running]:
github.com/elves/elvish/sys.DumpStack(0x0, 0x0)
	/home/pdietl/go/src/github.com/elves/elvish/sys/dumpstack.go:7 +0x7b
github.com/elves/elvish/shell.rescue()
	/home/pdietl/go/src/github.com/elves/elvish/shell/shell.go:68 +0xde
panic(0xa2b880, 0xc82000a160)
	/usr/lib/go-1.6/src/runtime/panic.go:443 +0x4e9
github.com/elves/elvish/parse.(*Form).parse.func1(0x0)
	/home/pdietl/go/src/github.com/elves/elvish/parse/parse.go:207 +0x2c
github.com/elves/elvish/parse.(*Form).parse(0xc82019b1e0, 0xc82036a0c0)
	/home/pdietl/go/src/github.com/elves/elvish/parse/parse.go:213 +0x932
github.com/elves/elvish/parse.ParseForm(0xc82036a0c0, 0x100000061)
	/home/pdietl/go/src/github.com/elves/elvish/parse/boilerplate.go:102 +0x65
github.com/elves/elvish/parse.(*Pipeline).parse(0xc82018f260, 0xc82036a0c0)
	/home/pdietl/go/src/github.com/elves/elvish/parse/parse.go:120 +0x25
github.com/elves/elvish/parse.ParsePipeline(0xc82036a0c0, 0xc800000001)
	/home/pdietl/go/src/github.com/elves/elvish/parse/boilerplate.go:47 +0x69
github.com/elves/elvish/parse.(*Chunk).parse(0xc82036a120, 0xc82036a0c0)
	/home/pdietl/go/src/github.com/elves/elvish/parse/parse.go:70 +0x61
github.com/elves/elvish/parse.ParseChunk(0xc82036a0c0, 0x0)
	/home/pdietl/go/src/github.com/elves/elvish/parse/boilerplate.go:22 +0x65
github.com/elves/elvish/parse.Parse(0xaf0af0, 0xd, 0xc820123f90, 0x7, 0x1, 0x0, 0x0)
	/home/pdietl/go/src/github.com/elves/elvish/parse/parse.go:18 +0x193
[...]

PR #403, opened 2017-07-07
fix Elvish Dies with Specific Key Sequence (#391)

HeavyHorst:
this should fix issue #391

### DIFF
--- a/parse/parse.go
+++ b/parse/parse.go
@@ -40,6 +40,7 @@ var (
 	errShouldBeFilename       = newError("", "a composite term representing filename")
 	errShouldBeArray          = newError("", "spaced")
 	errStringUnterminated     = newError("string not terminated")
+	errMultipleAssignment     = newError("assigning a single value to multiple variables isn't supported")
 	errInvalidEscape          = newError("invalid escape sequence")
 	errInvalidEscapeOct       = newError("invalid escape sequence", "octal digit")
 	errInvalidEscapeHex       = newError("invalid escape sequence", "hex digit")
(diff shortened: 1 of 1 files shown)


xiaq:
Hi, thanks for uncovering and fixing this!

However, although this patch fixes the crash, it doesn't give a meaningful semantics to forms like `x = y = foo`. There could be two choices:

1. Make this a syntax error.

2. Assign both `x` and `y`.

I think 2. is quite useful, but for the sake of fixing the crash I would happily accept a PR for 1. as well.

HeavyHorst:
Yeah, you're right.
2. should work now.

```
~> a = b = c = d = e = 100
~> echo $a $b $e
100 100 100
```

xiaq:
Hi, that was pretty quick :)

However, this is not a good implementation of 2. Elvish supports assignments of multiple variables, like `a b = lorem ipsum`. To implement 2, it should be extended to support forms like `a b = x y = lorem ipsum`. You are only dealing with `a = b = lorem`, and you also made forms like `a b c = lorem`, which is illegal, legal.

It's fine if 2 is a bit tricky to get right and you only implement 1 for the sake of fixing the crash.

HeavyHorst:
Oh, haven't thought about these edge cases.
I will make it a syntax error for now.

We should perhaps create a new issue to support the other assignment cases.

xiaq:
Hi, I see that you have made this into a syntax error. However, the checking is better done when encountering the `=` instead of within `addLHS`.

xiaq:
Thanks for fixing this. Beware that GitHub does not send me notifications when you modify your PR, you need to add a comment when you have done a requested change.

Issue #413, opened 2017-07-08
Iterating over maps


silvasur:
If I'm not overlooking something, it is currently not possible to iterate over the items of a map. This makes it basically impossible to process some JSON data structures out there.

My suggestion would be to add a function that returns the keys of a map, so one can iterate over them (and as a consequence can also access the values for the given keys).

I implemented a version of this function (named `keys`) at my fork: silvasur/elvish@bcf959c0d15ade60ea39d2370db9ccd0aa550856

Here's an example of using it:

```
my_map = [&a=foo &b=bar &c=baz]
keys $my_map | each [k]{ echo "key: "$k", value: "$my_map[$k] }
```
Will print:
```
key: a, value: foo
key: b, value: bar
key: c, value: baz
```

If you think this is a useful addition, please let me know, and I'll do a proper pull request and add documentation / tests for the new function.

xiaq:
@silvasur you are right that we currently lack a way to iterating keys, and thanks for the issue and the implementation, I am open for a PR.

BTW, I peeked the code -- please use the `IterateKeyer` interface instead of `Map`.

PR #416, opened 2017-07-08
Add `keys` function

silvasur:
Yields the keys of a map value

See also #413. As @xiaq suggested there I now use `IterateKeyer` instead of `map` like I did in a previous cersion of this.

(I'm not too happy about the second test I added. Since the order of map keys are not guaranteed, I needed to sort the returned keys first with the `sort` utility.)


silvasur:
See also elves/elvish.io#8 for the documentation part

PR #428, opened 2017-07-11
Solved failure of buildup when Cgo is disabled

xchenan:
Applied a dummy implementation of daemon/service when Cgo is disabled.

Signed-off-by: Xuan Chen <xchenan@gmail.com>

### DIFF
--- a/daemon/service/service_nocgo.go
+++ b/daemon/service/service_nocgo.go
@@ -1,0 +1,22 @@
+// +build !cgo
+
+// Package service implements the daemon service for mediating access to the
+// storage backend.
+package service
+
+import (
+	"github.com/elves/elvish/util"
+	"syscall"
+)
+
+var logger = util.GetLogger("[daemon-dummy] ")
+
+// A dummy implementation of Serve for the environment of Cgo being disabled.
+func Serve(sockpath, dbpath string) {
+	logger.Println("pid is", syscall.Getpid())
+	logger.Println("going to listen", sockpath)
+
+	select {}
+
+	logger.Println("exiting")
+}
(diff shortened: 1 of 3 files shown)


daemon/service/service_nocgo.go
```diff
@@ -0,0 +1,9 @@
+// +build !cgo
+
+// Package service implements the daemon service for mediating access to the
+// storage backend.
+package service
+
+// A dummy implementation of Serve for the environment of Cgo being disabled.
+func Serve(sockpath, dbpath string) {
```
xiaq: Please put a `select { }` to block the goroutine, thus staying close to the behaviour of the non-dummy Serve implementation. Please also add a log here (using `logger`) to make it clear that this is a dummy implementation.

xiaq: Hmm, I am not sure what you mean. What is the problem you are trying to solve? Your current approach seems good enough to me.

xchenan: Sorry about the confusion. I don't fully understand your previous comments.

From my understanding, you are suggesting that I shall still keep the goroutine as in service.go, and put a `select {}` afterwards which blocks function Serve() till the goroutine receives SIGTERM or SIGINT. Please correct me if I am wrong.

xiaq: Ah. I was suggesting that you put `select {}` in your dummy Serve implementation just for the sake of making it behave more like the real Serve, and add a logging for easier debugging.

xchenan:
@xiaq Please feel free to let me know what you think of it. Thanks for the patience.

PR #431, opened 2017-07-12
Enable and disable Cgo in Travis

xchenan:
Signed-off-by: Xuan Chen <xchenan@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,6 @@ os:
     - osx
 script:
     make travis
+env:
+    - CGO_ENABLED=0
+    - CGO_ENABLED=1


xiaq:
1. The Travis build is broken because the `store` package is broken, and the travis target in Makefile will test all packages. A workaround is to exclude `store` in the Makefile when CGO_ENABLED=0.

1. The `upload` target should not upload binary when CGO_ENABLED=0.

1. We now have a 3 (Go versions) x 2 (OSes) x 2 (CGO_ENABLED) build matrix, which is quite slow.

I would suggest to simply wait until we get rid of the SQLite dependency, and we will always build with CGO_ENABLED=0. I don't foresee any additional cgo code in the short term, so any breakages of cgo-disabled building is unlikely.

xchenan:
@xiaq I understand the situation and agree. I will close the PR for now.

xiaq:
FYI, Elvish is now cgo-free, thanks to @xofyarg :D

Issue #398, opened 2017-07-07
Builtin to test element existence


xiaq:
Like the `in` operator in Python.

PR #432, opened 2017-07-12
added a has-key and has-value builtin to test the existence of an element.

### LINKED ISSUE #407: Builtin for testing existence of key in map

HeavyHorst:
This works with maps and every iterable type.

```
~/go/src/github.com/elves/elvish> contains [1 2 "3"] 2
▶ $true
~/go/src/github.com/elves/elvish> contains [1 2 "3"] 5
▶ $false
~/go/src/github.com/elves/elvish> contains [&foo=bar &lorem=ipsum] "lorem"
▶ $true
~/go/src/github.com/elves/elvish> contains "hallo" "o"
▶ $true

```

@xiaq what do you think?

fixes #407 #398

### DIFF
--- a/eval/builtin_fn.go
+++ b/eval/builtin_fn.go
@@ -112,6 +112,8 @@ func init() {
 		{"take", take},
 		{"range", rangeFn},
 		{"count", count},
+		{"has-key", hasKey},
+		{"has-value", hasValue},
 
 		// String
 		{"joins", joins},
@@ -778,6 +780,56 @@ func rangeFn(ec *EvalCtx, args []Value, opts map[string]Value) {
 	for i := lower; i < upper; i += step {
 		out <- String(fmt.Sprintf("%g", i))
 	}
+}
+
+func hasValue(ec *EvalCtx, args []Value, opts map[string]Value) {
+	TakeNoOpt(opts)
+
+	var container, value Value
+	var found bool
+
+	ScanArgs(args, &container, &value)
+
+	switch container := container.(type) {
+	case Iterable:
+		container.Iterate(func(v Value) bool {
+			found = (v == value)
+			return !found
+		})
+	case MapLike:
+		container.IterateKey(func(v Value) bool {
+			found = (container.IndexOne(v) == value)
+			return !found
+		})
+	default:
+		throw(fmt.Errorf("argument of type '%s' is not iterable", container.Kind()))
+	}
+
+	ec.ports[1].Chan <- Bool(found)
+}
+
+func hasKey(ec *EvalCtx, args []Value, opts map[string]Value) {
+	TakeNoOpt(opts)
+
+	var container, key Value
+	var found bool
+
+	ScanArgs(args, &container, &key)
+
+	switch container := container.(type) {
+	case HasKeyer:
+		found = container.HasKey(key)
+	case Lener:
+		// XXX(xiaq): Not all types that implement Lener have numerical indices
+		err := util.PCall(func() {
+			ParseAndFixListIndex(ToString(key), container.Len())
+		})
+		found = (err == nil)
+	default:
+		throw(fmt.Errorf("couldn't get key or index of type '%s'", container.Kind()))
+	}
+
+	ec.ports[1].Chan <- Bool(found)
 }
 
 func count(ec *EvalCtx, args []Value, opts map[string]Value) {
--- a/eval/eval_test.go
+++ b/eval/eval_test.go
@@ -292,6 +292,17 @@ var evalTests = []struct {
 	{`ord a`, strs("0x61"), nomore},
 	{`base 16 42 233`, strs("2a", "e9"), nomore},
 	{`wcswidth 你好`, strs("4"), nomore},
+	{`has-key [foo bar] 0`, bools(true), nomore},
+	{`has-key [foo bar] 0:1`, bools(true), nomore},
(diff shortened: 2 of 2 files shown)


xiaq:
This has the same problem with Python's `in` operator: it tests for presence of values for lists, and the presence of keys for maps. This former is more obvious, but the latter is not.

It's better to have a pair of builtins `has-key` and `has-value` to make this explicit. It would also be desirable if `has-key` handles slice index for lists. Examples:

```sh
~> li = [lorem ipsum foo bar]
~> map = [&lorem=ipsum &foo=bar]
~> has-key $li 0
$true
~> has-key $li 1:3 # slice for lists
$true
~> has-key $li 100
$false
~> has-key $m haha
$false
~> has-key $m lorem
$true
~> # maps don't have slice indices
```

You can use `ParseAndFixListIndex` to parse list indices, using `util.PCall` to catch the "exception".

The `has-value` builtin should hopefully be obvious.

HeavyHorst:
I've splitted the logic into has-key and has-value as requested.

```
~/go/src/github.com/elves/elvish> li = [lorem ipsum foo bar]
~/go/src/github.com/elves/elvish> map = [&lorem=ipsum &foo=bar]
~/go/src/github.com/elves/elvish> has-key $li 0
▶ $true
~/go/src/github.com/elves/elvish> has-key $li 1:3
▶ $true
~/go/src/github.com/elves/elvish> has-key $li 100
▶ $false
~/go/src/github.com/elves/elvish> has-key $map haha
▶ $false
~/go/src/github.com/elves/elvish> has-key $map lorem
▶ $true

```

eval/builtin_fn.go
```diff
@@ -780,6 +782,57 @@ func rangeFn(ec *EvalCtx, args []Value, opts map[string]Value) {
...
+			throw(fmt.Errorf("argument of type '%s' is not iterable", v.Kind()))
+		}
+	} else {
+		throw(errors.New("want 2 arguments"))
+	}
+
+	ec.ports[1].Chan <- Bool(found)
+}
+
+func hasKey(ec *EvalCtx, args []Value, opts map[string]Value) {
```
xiaq: Use `ScanArgs`.

eval/builtin_fn.go
```diff
@@ -780,6 +782,57 @@ func rangeFn(ec *EvalCtx, args []Value, opts map[string]Value) {
...
+	TakeNoOpt(opts)
+
+	var found bool
+	if len(args) == 2 {
+		v := args[0]
+		switch val := v.(type) {
+		case HasKeyer:
+			if val.HasKey(args[1]) {
```
xiaq: `if x { found = true }` is equivalent to `found = x`.

eval/builtin_fn.go
```diff
@@ -780,6 +782,57 @@ func rangeFn(ec *EvalCtx, args []Value, opts map[string]Value) {
...
+				found = true
+			}
+		case Lener:
+			if err := util.PCall(func() {
```
xiaq: Same; just use `found = ...`

eval/builtin_fn.go
```diff
@@ -780,6 +782,57 @@ func rangeFn(ec *EvalCtx, args []Value, opts map[string]Value) {
 	}
 }
 
+func hasValue(ec *EvalCtx, args []Value, opts map[string]Value) {
+	TakeNoOpt(opts)
```
xiaq: You can declare `var container, key Value` and use `ScanArgs` here. `ScanArgs` also check the number of arguments.

eval/builtin_fn.go
```diff
@@ -780,6 +782,57 @@ func rangeFn(ec *EvalCtx, args []Value, opts map[string]Value) {
...
+
+	var found bool
+	if len(args) == 2 {
+		v := args[0]
+		switch val := v.(type) {
+		case HasKeyer:
+			if val.HasKey(args[1]) {
+				found = true
+			}
+		case Lener:
```
xiaq: Hmm, this is actually a bit brittle as `Lener` does not always accept numerical indices (e.g. maps also implement `Lener`). But I cannot think of a better solution at the moment, please add a `XXX` comment here like `XXX(xiaq): Not all types that implement Lener have numerical indices`

eval/builtin_fn.go
```diff
@@ -780,6 +782,57 @@ func rangeFn(ec *EvalCtx, args []Value, opts map[string]Value) {
...
+		v := args[0]
+		if iterator, ok := v.(Iterable); ok {
+			iterator.Iterate(func(v Value) bool {
+				if v == args[1] {
+					found = true
+					return false
+				}
+				return true
+			})
+		} else {
```
xiaq: It would be great if `has-value` also works for `MapLike` types.

eval/builtin_fn.go
```diff
@@ -780,6 +782,57 @@ func rangeFn(ec *EvalCtx, args []Value, opts map[string]Value) {
...
+	ec.ports[1].Chan <- Bool(found)
+}
+
+func hasKey(ec *EvalCtx, args []Value, opts map[string]Value) {
+	TakeNoOpt(opts)
+
+	var found bool
+	if len(args) == 2 {
+		v := args[0]
+		switch val := v.(type) {
```
xiaq: You can reuse the name `v`: `v := v.(type)`

HeavyHorst:
Done. :)

Issue #228, opened 2016-10-08
Case insensitive completion


zhsj:

xiaq:
To use a case-insensitive prefix matcher, do

```sh
edit:-matcher[''] = [a]{ edit:match-prefix &ignore-case=$true $@a }
```

There is also "smart case" a la Vim -- case insensitive when the pattern is lower case, case sensitive otherwise:

```sh
edit:-matcher[''] = [a]{ edit:match-prefix &smart-case=$true $@a }
```

PR #433, opened 2017-07-13
Add user-defined 'matcher' interface

ALSchwalm:
This PR adds a user-defined 'matcher' interface. Generally, the idea is to implement a user-defined map of functions that are used as the filters for commands, arguments, variables, redirects, and indexes. Currently, there are two 'global' matchers, the default 'prefix matcher' and the 'subsequence matcher' that can be activated by setting `-use-subseq-matcher`. 

With this patch, you can, for example, allow case-independent completion for variables by setting:
    
    edit:-matcher[variable] = [pattern]{ each [cand]{
        put (re:match '(?i)^'$pattern $cand)
      }
    }
Note that the matcher is passed the pattern it match against as an argument. Completions can be read from the input stream. It must return a bool for each completion, indicating whether it is an acceptable match for `pattern`.

This interface is somewhat less intuitive than the callback being per-candidate, but has much better performance (especially for the builtin matcher). This PR should close #433, #228, and possibly #396

### DIFF
--- a/edit/candidate.go
+++ b/edit/candidate.go
@@ -81,15 +81,3 @@ func outputComplexCandidate(ec *eval.EvalCtx,
 
 	ec.OutputChan() <- c
 }
-
-func cookCandidates(raws []rawCandidate, pattern string,
-	match func(string, string) bool, q parse.PrimaryType) []*candidate {
-
-	var cooked []*candidate
-	for _, raw := range raws {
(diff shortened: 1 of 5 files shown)


xiaq:
You can pass candidates using the input channel instead of as arguments.

ALSchwalm:
Done. Updated the description.

PR #437, opened 2017-07-13
Add eval.ScanIntoOpts and minor changes to narrow mode

xofyarg:
eval.ScanIntoOpts is similar to eval.ScanOpts, instead of specifying
each option separately, it accepts an option struct and parse the
options according the struct internally.

Options organized as a group in a struct is more readable and could be
extend in the future if the calling function requires many.


xiaq:
This is pretty neat. Thanks!

PR #444, opened 2017-07-18
make matcher concurrent with completer

tw4452852:

### DIFF
--- a/edit/candidate.go
+++ b/edit/candidate.go
@@ -3,6 +3,7 @@ package edit
 import (
 	"fmt"
 	"os"
+	"sort"
 
 	"github.com/elves/elvish/edit/ui"
 	"github.com/elves/elvish/eval"
@@ -20,6 +21,12 @@ type rawCandidate interface {
 	text() string
 	cook(q parse.PrimaryType) *candidate
 }
+
+type rawCandidates []rawCandidate
+
+func (cs rawCandidates) Len() int           { return len(cs) }
+func (cs rawCandidates) Swap(i, j int)      { cs[i], cs[j] = cs[j], cs[i] }
+func (cs rawCandidates) Less(i, j int) bool { return cs[i].text() < cs[j].text() }
 
 // plainCandidate is a minimal implementation of rawCandidate.
 type plainCandidate string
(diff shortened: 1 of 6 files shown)


edit/candidate.go
```diff
@@ -101,39 +101,29 @@ func outputComplexCandidate(ec *eval.EvalCtx,
...
-	for i, value := range values {
-		if eval.ToBool(value) {
-			filtered = append(filtered, cands[i].cook(q))
+	for _, value := range values {
+		if rc, ok := value.(rawCandidate); ok {
+			filtered = append(filtered, rc.cook(q))
 		}
 	}
+
+	// TODO
```
xiaq: Please sort the candidates. Sorting (cooked) candidates by their `menu.text` should be good enough.

edit/candidate.go
```diff
@@ -101,39 +101,29 @@ func outputComplexCandidate(ec *eval.EvalCtx,
...
-	// directly to a channel and run concurrently with the matcher.
-	for _, cand := range cands {
-		input <- eval.String(cand.text())
-	}
-	close(input)
-
 	args := []eval.Value{eval.String(pattern)}
 	values, err := ec.PCaptureOutput(matcher, args, eval.NoOpts)
 	if err != nil {
+		logger.Printf("tw; %s\n", err)
```
xiaq: Please remove this.

edit/completers.go
```diff
@@ -111,29 +116,34 @@ func complVariable(n parse.Node, ev *eval.Evaler) (*complSpec, error) {
...
+	}()
 
-	var rawCands []rawCandidate
-	for _, entry := range entries {
-		rawCands = append(rawCands, noQuoteCandidate(entry))
-	}
-	cands, err := ev.Editor.(*Editor).filterAndCookCandidates(ev, "variable", nameHead,
+	cands, err := ev.Editor.(*Editor).filterAndCookCandidates(ev, matcher, nameHead,
 		rawCands, parse.Bareword)
+	<-collectDone
```
xiaq: Can you explain why this synchronization is necessary? If it is, consider using `close(collectDone)` above, it is the style used throughout the rest of the program.

edit/candidate.go
```diff
@@ -101,39 +101,29 @@ func outputComplexCandidate(ec *eval.EvalCtx,
...
 		return nil, err
-	} else if len(values) != len(cands) {
-		return nil, errIncorrectNumOfResults
 	}
 
 	var filtered []*candidate
-	for i, value := range values {
-		if eval.ToBool(value) {
-			filtered = append(filtered, cands[i].cook(q))
+	for _, value := range values {
```
tw4452852: But I think returning the filtering results is simpler and it doesn't add additional complexity to custom matcher writing. It just returns what they want instead of the boolean value of what they want. There is also another benefit: matcher could also return additional candidates according to the original input.

Also there is no way to obtain the original raw candidates here.

edit/completers.go
tw4452852: Just make sure completer goroutine exits to prevent leakage.

edit/candidate.go
tw4452852: Done.

edit/candidate.go
tw4452852: Done.

edit/candidate.go
xiaq: The matcher needs to be aware of the actual types of the inputs, as you have done in the type switch inside `wrapMatcher`. Right now user-defined functions cannot access the `text()` method of `rawCandidate`, and this makes custom matcher in elv no longer possible.

We could introduce a way to access the `text()` result from elv, for instance by adding an `get-text-for-matching` builtin, and rely on matchers to use it, but this results in a less flexible API. Exposing only the text to filter to the matcher creates more work for the caller, but it also gives more control to the caller.

For instance, we do not only want user-customizable matchers for completion candidates, but also for items in listing modes (e.g. directory history item, command history item). To make matchers work for listing items, we can simply extend `get-text-for-matching` to work on listing items as well; for instance, calling it on a directory history item will simply give the directory path. But what if we add some more complex listing modes, where a listing item has multiple fields that can be filtered on -- e.g. a "process listing mode", where each process has fields like user, tty, command name? Then we have to make a choice of *which* of the fields `get-text-for-matching` returns, and by doing this you take away the ability to match on all other fields. Instead, if we only give strings to the matcher, the caller can always decide which field to give to the matcher. This is what I mean by saying that supplying strings gives more control to the caller.

I cannot think of a good use case where allowing matchers to add new candidates is useful; do you have one in mind? It also sounds like a generally bad idea to allow matchers to do too many things.

The original raw candidates come from the `rawCands` channel, and you can add another goroutine that feeds the `text()` of those raw candidates and collect the output.

edit/candidate.go
```diff
@@ -101,39 +108,27 @@ func outputComplexCandidate(ec *eval.EvalCtx,
...
+		if rc, ok := value.(rawCandidate); ok {
+			filtered = append(filtered, rc.cook(q))
```
xiaq: Another thing -- sorry for not raising this earlier -- it's perhaps better to collect raw candidates, sort them by `text()` before cooking them, instead of sorting cooked candidates. Suppose that you have a directory containing files `a`, `x`, `a b`, `x y`, sorting cooked candidates will sort them by the quoted version, resulting in the ordering `'a b'`, `'x y'`, `a`, `x`; I think it's better to sort them by the unquoted version, namely `a`, `'a b'`, `x`, `'x y'`.

edit/candidate.go
tw4452852: Ah, it makes more sense to feed the matcher with string. What do you think of just returning filtered result instead boolean value for each input?

tw4452852: Discard my previous comment.

tw4452852: Done.

edit/candidate.go
tw4452852: Done.

xiaq:
I see that you needed to duplicate some code from eval.pcaptureOutput to capture output concurrently; there should be a concurrent version of PCaptureOutput, but your PR is good to merge. Thank you for bearing with me :)

PR #447, opened 2017-07-20
Enable build on more platforms

zhsj:
Since we get rid of sqlite3, now we can easily build elvish on more platforms.

This commit is tested by the following script:

https://gist.github.com/zhsj/74489e93b60ec35261f9877f769b4b8c

Tests results are also attatched.

For darwin,{arm,arm64}, go cross build failed with:

```
# github.com/elves/elvish
warning: unable to find runtime/cgo.a
/usr/lib/go-1.8/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: unrecognized option '-pagezero_size'
/usr/bin/ld: use the --help option for usage information
collect2: error: ld returned 1 exit status
```
So these platforms are unknown whether it can be built or not.

### DIFF
--- a/sys/select_32.go
+++ b/sys/select_32.go
@@ -1,4 +1,4 @@
-// +build darwin 386,linux arm,linux netbsd openbsd
+// +build darwin 386,linux arm,linux mips,linux mipsle,linux
 
 // The type of FdSet.Bits is different on different platforms.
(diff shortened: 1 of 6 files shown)


xiaq:
Let's not worry about iOS for now.

PR #450, opened 2017-07-21
Fix crash in concurrent chan operation for arg completer

xofyarg:
Wait for the producer to finish before closing the chan.

### DIFF
--- a/edit/arg_completers.go
+++ b/edit/arg_completers.go
@@ -138,16 +138,17 @@ func (bac *builtinArgCompleter) Call(ec *eval.EvalCtx, args []eval.Value, opts m
 		words[i] = string(s)
 	}
 
-	output := ec.OutputChan()
 	rawCands := make(chan rawCandidate)
-	defer close(rawCands)
+	var err error
 	go func() {
-		for rc := range rawCands {
-			output <- rc
-		}
+		defer close(rawCands)
+		err = bac.impl(words, ec.Evaler, rawCands)
 	}()
 
-	err := bac.impl(words, ec.Evaler, rawCands)
+	output := ec.OutputChan()
+	for rc := range rawCands {
+		output <- rc
+	}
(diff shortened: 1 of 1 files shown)


edit/arg_completers.go
```diff
@@ -138,17 +138,16 @@ func (bac *builtinArgCompleter) Call(ec *eval.EvalCtx, args []eval.Value, opts m
...
 
-	output := ec.OutputChan()
 	rawCands := make(chan rawCandidate)
-	defer close(rawCands)
 	go func() {
-		for rc := range rawCands {
-			output <- rc
-		}
+		defer close(rawCands)
+		maybeThrow(bac.impl(words, ec.Evaler, rawCands))
```
xiaq: You need to keep this on the main goroutine, otherwise it will crash this one.

xofyarg: May I get the reason if you don't mind?

xiaq: Throw uses panics under the hood, and panics are local to a goroutine.

xofyarg: Ah, thanks. Updated.

edit/arg_completers.go
```diff
@@ -138,16 +138,20 @@ func (bac *builtinArgCompleter) Call(ec *eval.EvalCtx, args []eval.Value, opts m
...
-	output := ec.OutputChan()
 	rawCands := make(chan rawCandidate)
-	defer close(rawCands)
+	var err error
 	go func() {
-		for rc := range rawCands {
-			output <- rc
+		defer close(rawCands)
+		err = bac.impl(words, ec.Evaler, rawCands)
+		if err != nil {
```
xiaq: Well, this is not needed...

xofyarg: My finger was not under control...

PR #458, opened 2017-07-23
Add a 'compileonly' option

ALSchwalm:
This PR just adds a `-compileonly` flag that will print any parsing/compilation errors but not execute anything. This is useful for writing tooling for elvish (for example, an emacs [flycheck](http://www.flycheck.org/en/latest/) mode). I'm not using the pretty printing facilities to make parsing the errors easier.


xiaq:
Looks good, but in future the error format should be customizable. For instance, the user should be able to request a JSON formatting of the error object.

PR #461, opened 2017-07-30
Changed default shortcut for location mode to Ctrl-/

muesli:
The previous default was Ctrl-L, which is commonly used for
clearing the terminal. As discussed with xiaq, Ctrl-/ is a
nice alternative which doesn't clash with any existing user
expectations.

### DIFF
--- a/edit/insert.go
+++ b/edit/insert.go
@@ -91,7 +91,7 @@ func init() {
 		{'N', ui.Ctrl}: "navigation:start",
 		{'R', ui.Ctrl}: "histlist:start",
 		{'1', ui.Alt}:  "lastcmd:start",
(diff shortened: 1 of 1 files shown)


xiaq:
Hi, thanks.

Interestingly, the way to send the key code 0x1f, what Elvish now considers as Ctrl-/ but better known as `^_` elsewhere, is inconsistent across platforms. I think on the Linux terminals I tested it was Ctrl-/, but in Terminal.app on macOS it is Ctrl-Dash. In conemu on Windows, there doesn't seem to be a way to send this keycode at all.

xiaq:
Closing because Ctrl-/ is not consistent across platforms. :(

PR #463, opened 2017-08-04
add string replace builtin

tw4452852:


xiaq:
Thanks. Functions in re: use the option name &max for the numerical option. Please also use that for consistency.

tw4452852:
Done.

PR #465, opened 2017-08-09
Updates and fixes to narrow.elv

zzamboni:
- Updated to account for recent syntax changes in Elvish
- Added hook capabilities to all three modes, documented in `narrow.elv`.
- Updated embedded module.
- Added `bind_keys` function to set key bindings.

### DIFF
--- a/eval/narrow.elv
+++ b/eval/narrow.elv
@@ -4,15 +4,15 @@
 #
 # Usage:
 #   use narrow
-#   narrow:bind_keys
+#   narrow:bind-trigger-keys
 #
-# narrow:bind_keys binds keys for location, history and lastcmd
+# narrow:bind-trigger-keys binds keys for location, history and lastcmd
 # modes. Without options, it uses the default bindings (same as the
 # default bindings for edit:location, edit:history and edit:lastcmd),
 # but different keys can be specified with the options. To disable a
 # binding, specify its key as "".
 # Example:
-#   narrow:bind_keys &location=Alt-l &lastcmd="" 
+#   narrow:bind-trigger-keys &location=Alt-l &lastcmd="" 
 
 # Hooks
 # Each hook variable is an array which must contain lambdas, all of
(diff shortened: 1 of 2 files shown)


eval/narrow.elv
```diff
@@ -1,78 +1,124 @@
...
+    put $@candidates
+  } [arg]{
+    edit:insert-at-dot $arg[content]
+    for hook $after-lastcmd { $hook }
+  } &modeline="[narrow] Lastcmd " &auto-commit=$true &bindings=[&M-1={ edit:narrow:accept-close }] &ignore-case=$true
 }
 
 
-# TODO: separate bindings from functions
+fn -bind_i [k f]{
```
xiaq: Consider calling this `-bind-insert`.

eval/narrow.elv
```diff
@@ -1,78 +1,124 @@
...
+  edit:insert:binding[$k] = $f
+}
 
-fn bind [m k f]{
-    edit:binding[$m][$k] = $f
+fn -bind_n [k f]{
```
xiaq: Consider calling this just `-bind`. When using this mode it will be `narrow:-bind`.

eval/narrow.elv
```diff
@@ -1,78 +1,124 @@
...
-bind insert C-l       narrow:location
-bind insert C-r       narrow:history
-bind insert M-1       narrow:lastcmd
+# Bind keys for location, history and lastcmd modes. Without
+# options, it uses the default bindings, but different keys
+# can be specified with the options. To disable a binding,
+# specify its key as "".
+# Example:
+#   narrow:bind_keys &location=Alt-l &lastcmd=""
+fn bind_keys [&location=C-l &history=C-r &lastcmd=M-1]{
```
xiaq: `bind-trigger-keys` seems better.

eval/narrow.elv
```diff
@@ -1,78 +1,124 @@
...
+  if (> (count $location) 0) { -bind_i $location narrow:location }
```
xiaq: I was going to say `if not-eq $location ""` is clearer here, then I found out that we don't have that `not-eq` builtin. I just added it, so that I can now say `if not-eq $location ""` is clearer. :)

zzamboni:
@xiaq Thanks for the feedback! I've addressed all the comments in my latest commit.

PR #469, opened 2017-08-25
Trim spaces from command before adding it to history

muesli:
This prevents accidental new-lines and spaces (e.g. from copy/paste actions) from unnecessarily polluting the history.

### DIFF
--- a/edit/history/fuser.go
+++ b/edit/history/fuser.go
@@ -1,6 +1,7 @@
 package history
 
 import (
+	"strings"
 	"sync"
 )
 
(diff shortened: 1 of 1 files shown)


muesli:
Looking into the failing tests right now.

muesli:
Ok, test failing tests looks unrelated to this PR.

xiaq:
What if the command is

```
echo `

```
? :)

muesli:
Not sure if GitHub breaks your paste or if that's supposed to a valid command. In other words: what are "valid" commands that end with a whitespace (and would break without it)?

If the whitespace is really quired, how about stripping just newlines? How about newlines with only whitespace?

xiaq:
@muesli That is a backtick followed by a newline, which functions as line continuation. Striping the newline makes it incomplete.

xiaq:
Closing because as the syntax of Elvish evolves, more cases where trailing whitespaces cannot be straightforwardly trimmed might appear. We might want to revisit this when the syntax matures a bit more.

PR #477, opened 2017-09-03
Use golang.org/x/sys in sys package

zhsj:


xiaq:
I will revisit this after I have tagged 0.10.

xiaq:
Looks valid. Let me just merge this and see if anything goes wrong at all.

PR #478, opened 2017-09-03
Use reflect to get nFdBits size

zhsj:


sys/select.go
```diff
@@ -1,8 +1,13 @@
...
 
 package sys
 
-import "syscall"
+import (
+	"reflect"
+	"syscall"
+)
+
+var nFdBits = (uint)(reflect.TypeOf(syscall.FdSet{}.Bits[0]).Size() * 8)
```
xiaq: You can use `unsafe.Sizeof`, which does not have the overhead of `reflect`. Also, in Go you do not need parentheses around the type name for casting. Use

```go
var nFdBits = uint(8 * unsafe.Sizeof(syscall.FdSet{}.Bits[0]))
```

zhsj: fixed.

zhsj:
we inline the index function to let go guess the type itself, so no need gccgo version again.

PR #487, opened 2017-09-20
Print a shorter traceback message in interactive mode

doppioandante:
Before:
```
~> l
Exception: search l: not found
Traceback:
  [interactive], line 1:
    l
~> test `
   (te`
   st)
Exception: search te: not found
Traceback:
  [interactive], line 2-3:
    (te`
    st)
~>
```

After:
```
~> l
Exception: search l: not found
line 1: l
~> test `
   (te`
   st)
Exception: search te: not found
line 2:3 (te`
         st)
~>
```

### DIFF
--- a/eval/exception.go
+++ b/eval/exception.go
@@ -38,12 +38,28 @@ func (exc *Exception) Pprint(indent string) string {
 	} else {
 		msg = "\033[31;1m" + exc.Cause.Error() + "\033[m"
 	}
-	fmt.Fprintf(buf, "Exception: %s\n", msg)
-	buf.WriteString(indent + "Traceback:")
+	fmt.Fprintf(buf, "Exception: %s", msg)
+
+	// In case of a simple error generated from only one line
+	// don't print the full traceback format by setting
+	// printContext to false.
+	printContext := false
+	if tb := exc.Traceback; tb != nil {
+		if tb.Name != "[interactive]" {
+			printContext = true
+		}
+		printContext = printContext || tb.Next != nil
+	}
+
+	sourceIndent := indent
+	if printContext {
+		sourceIndent += "    "
+		buf.WriteString("\n" + indent + "Traceback:")
+	}
 
 	for tb := exc.Traceback; tb != nil; tb = tb.Next {
-		buf.WriteString("\n" + indent + "  ")
-		tb.Pprint(buf, indent+"    ")
+		buf.WriteString("\n" + sourceIndent)
+		tb.PprintOption(buf, sourceIndent+" ", printContext)
 	}
 
 	if pipeExcs, ok := exc.Cause.(PipelineError); ok {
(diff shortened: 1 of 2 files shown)


util/source_context.go
```diff
@@ -43,14 +47,29 @@ func (sc *SourceContext) Pprint(w io.Writer, sourceIndent string) {
...
+
+	// Save line %d expansion on a string, in case
+	// its length is needed for indentation (printContext=false).
+	var culpritLines string
 	if beginLine == endLine {
-		fmt.Fprintf(w, "%s, line %d:\n", sc.Name, beginLine)
+		culpritLines = fmt.Sprintf("line %d:", beginLine)
 	} else {
-		fmt.Fprintf(w, "%s, line %d-%d:\n", sc.Name, beginLine, endLine)
+		culpritLines = fmt.Sprintf("line %d:%d", beginLine, endLine)
```
doppioandante: This is wrong, should be %d-%d:

eval/exception.go
```diff
@@ -38,12 +38,28 @@ func (exc *Exception) Pprint(indent string) string {
...
-	fmt.Fprintf(buf, "Exception: %s\n", msg)
-	buf.WriteString(indent + "Traceback:")
+	fmt.Fprintf(buf, "Exception: %s", msg)
+
+	// In case of a simple error generated from only one line
+	// don't print the full traceback format by setting
+	// printContext to false.
+	printContext := false
+	if tb := exc.Traceback; tb != nil {
+		if tb.Name != "[interactive]" {
```
doppioandante: Having "[interactive]" here doesn't look good, could it be turn into a constant?

eval/exception.go
```diff
@@ -38,12 +38,28 @@ func (exc *Exception) Pprint(indent string) string {
...
+			printContext = true
+		}
+		printContext = printContext || tb.Next != nil
+	}
+
+	sourceIndent := indent
+	if printContext {
+		sourceIndent += "    "
+		buf.WriteString("\n" + indent + "Traceback:")
```
doppioandante: Ops, increment of sourceIndent should go after the WriteString

eval/exception.go
xiaq: Add an additional field to `util.SourceContext` to indicate whether the source is an interactive session (say `tb.Interactive`).

I would prefer you invert the logics here - instead of keeping a `printContext` flag, keep a `abbrTraceback` flag:

```go
// Abbreviate the traceback when it only contains one entry and that entry is from the interactive session
abbrTraceback := exc.Traceback.Interactive && exc.Traceback.Next == nil
```

doppioandante:
@xiaq Thanks for the review, this was the path of least resistance to the result. I agree with the refactoring you propose, I will do it tomorrow.

xiaq:
Closed by accident.

@doppioandante, any difficulty with the refactoring?

doppioandante:
@xiaq I recall there was something that prevented the refactoring to be straightforward, but I had no time to continue afterwards. I still plan to finish this though.

xiaq:
I implemented the idea myself, so I'm closing this PR. Still, thank you for the inspiration! :)

Issue #279, opened 2016-10-14
add -version argument or $version to display commit hash


huiyiqun:
For debug purpose

xiaq:
Nice to have, tricky to implement.

PR #488, opened 2017-09-21
Add -version

zhsj:
People install with go get, will show version unknown
But if they install from prebuilt, make get, they will
known their version

Closes: #279

Signed-off-by: Shengjing Zhu <i@zhsj.me>


zzamboni:
This is nice - would be nicer if it showed you the git commit number when you install with `go get`.

Issue #503, opened 2017-10-15
Bad matcher cause completion to hang


xiaq:
Minimal example:

1. Do `edit:-matcher[''] = { bad command }`
2. In a directory with sufficiently many files, do `echo `<Tab>

The editor hangs.

tw4452852:
I think we could close this issue now.

xiaq:
FWIW, if you wrote "fix #503" or "fixes #503" this issue will be auto-closed: https://help.github.com/articles/closing-issues-using-keywords/

PR #517, opened 2017-11-22
edit: drain up input channel before filter returns

tw4452852:
fix issue #503


xiaq:
Thank you for fixing this!

I see you are also doing other fixes. Can you split the them into their own commits?

tw4452852:
@xiaq 
Done.

PR #518, opened 2017-11-23
edit: support custom location matcher 

tw4452852:

### DIFF
--- a/edit/location/location.go
+++ b/edit/location/location.go
@@ -31,17 +31,22 @@ type mode struct {
 	binding eddefs.BindingMap
 	hidden  vector.Vector
 	pinned  vector.Vector
+	matcher eval.Callable
 }
+
+var matchDirPatternBuiltin = eval.NewBuiltinFn("edit:location:match-dir-pattern", matchDirPattern)
 
 // Init initializes the location mode for an Editor.
 func Init(ed eddefs.Editor, ns eval.Ns) {
-	m := &mode{ed, eddefs.EmptyBindingMap, vals.EmptyList, vals.EmptyList}
+	m := &mode{ed, eddefs.EmptyBindingMap, vals.EmptyList, vals.EmptyList, matchDirPatternBuiltin}
 	ns.AddNs("location",
 		eval.Ns{
 			"binding": vars.FromPtr(&m.binding),
 			"hidden":  vars.FromPtr(&m.hidden),
 			"pinned":  vars.FromPtr(&m.pinned),
-		}.AddBuiltinFn("edit:location:", "start", m.start))
+			"matcher": vars.FromPtr(&m.matcher),
+		}.AddBuiltinFn("edit:location:", "start", m.start).
+			AddFn("match-pattern", matchDirPatternBuiltin))
 }
 
 func (m *mode) start() {
@@ -75,7 +80,7 @@ func (m *mode) start() {
 	// Drop the error. When there is an error, home is "", which is used to
 	// signify "no home known" in location.
 	home, _ := util.GetHome("")
-	ed.SetModeListing(m.binding, newProvider(dirs, home))
+	ed.SetModeListing(m.binding, newProvider(dirs, home, ed.Evaler(), m.matcher))
 }
 
 // convertListToDirs converts a list of strings to []storedefs.Dir. It uses the
@@ -105,13 +110,15 @@ func convertListsToSet(lis ...vector.Vector) map[string]struct{} {
 }
 
 type provider struct {
+	ev       *eval.Evaler
+	matcher  eval.Callable
 	home     string // The home directory; leave empty if unknown.
 	all      []storedefs.Dir
 	filtered []storedefs.Dir
 }
 
-func newProvider(dirs []storedefs.Dir, home string) *provider {
-	return &provider{all: dirs, home: home}
+func newProvider(dirs []storedefs.Dir, home string, ev *eval.Evaler, matcher eval.Callable) *provider {
+	return &provider{ev: ev, matcher: matcher, all: dirs, home: home}
 }
 
 func (*provider) ModeTitle(i int) string {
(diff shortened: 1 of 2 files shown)


edit/location.go
```diff
@@ -31,13 +31,14 @@ func init() {
 var PinnedScore = math.Inf(1)
 
 type location struct {
-	home     string // The home directory; leave empty if unknown.
-	all      []storedefs.Dir
-	filtered []storedefs.Dir
+	home            string // The home directory; leave empty if unknown.
+	all             []storedefs.Dir
+	filtered        []storedefs.Dir
+	caseInsensitive bool
```
xiaq: Use ignoreCase in accordance with the matcher flag `&ignore-case`. Same for below.

edit/location_test.go
```diff
@@ -12,16 +12,18 @@ var (
 		{"/pinned", PinnedScore},
 		{"/src/github.com/elves/elvish", 300},
 		{"/src/home/xyz", 233},
+		{"/deAd", 200},
 		{"/home/dir", 100},
 		{"/foo/\nbar", 77},
 		{"/usr/elves/elvish", 6},
-	}, "/home")
+	}, "/home", true)
```
xiaq: Please include tests for both case-sensitive and insensitive scenerios.

edit/location.go
```diff
@@ -215,3 +219,11 @@ var _ = RegisterVariable("loc-pinned", func() eval.Variable {
 func (ed *Editor) locPinned() eval.List {
 	return ed.variables["loc-pinned"].Get().(eval.List)
 }
+
```
xiaq: As said above, this should only be a temporary solution. Prefix the variable with a dash and make it `-loc-ignore-case`.

tw4452852:
@xiaq 
Indeed, so i just give it a shot, now user can set the location matcher with `edit:-matcher['-location']`, and it will leverage the same options as other matchers, like `$smart-case` and `$ignore-case`.

xiaq:
Hi, sorry for the delay. I feel there are two problems with the implementation:

* You are changing the representation of (filtered) dir entries from `storedefs.Dir` to a struct that satisfies the `candidate` interface, as well as introducing the `lesser` interface, so that you can use `filterAndCookCandidates`. This seems unnecessary: the matcher API is not tied to the `candidate` interface and can be used to filter arbitrary entries. Please keep the representation of `location.filtered` and implement a `filterDirs` helper function to filter `storedefs.Dir` slices instead.

* Your implementation of `locationMatch` seems to be different from the current matcher. In particular, when `s = "a/b/c/d"` and `p = "a/d"`, the current matcher considers that a match, while `locationMatch` doesn't. Consider implementing a matcher on top of the current `makeLocationFilterPattern` function, instead of relying on `wrapMatcher`.

Use convenience functions when they make things straightforward. If you need to make extra efforts to just use a convenience function, prefer working with the lower level interface directly. It is important that we don't introduce unnecessary indirections.

Again, sorry for the delay and the not-so-positive review comments and thank you very much for the contributions! I know it can be frustrating for you. In future, you may want to discuss the implementation approach a bit before writing the code (unfortunately this means more round-trip with me, and I have quite unpredictable latency these days, and I am trying to fix this). Also, if you are comfortable experimenting and iterating with code directly in this way, feel free to continue to do so.

Finally, a comment on the API: Please move the customization point to `$edit:loc:matcher` instead of `$edit:-matcher`; the latter is intended for completion matchers. I will move it to `$edit:completion:matcher` to make it clear that it is for completions.

tw4452852:
@xiaq 
I agree with your comments. Maybe I should think this over and let's just abandon my second commit.

BTW, you could cherry-pick my first commit. It's just for a better performance.

notramo:
@tw4452852 @xiaq 
Any progress?

xiaq:
@notramo I cannot speak for tw, but I am not working on this right now.

tw4452852:
@notramo I will try to finish this recently.

edit/location/location.go
```diff
@@ -31,17 +31,22 @@ type mode struct {
 	binding eddefs.BindingMap
 	hidden  vector.Vector
 	pinned  vector.Vector
+	matcher eval.Callable
 }
 
+var matchPattern = eval.NewBuiltinFn("edit:location:match-pattern", patternMatch)
```
xiaq: I would prefer `match-dir-pattern`.

Don't name one thing `AB` and another `BA` just to avoid name conflict :) You can call the function `matchDirPattern` and the variable `matchDirPatternBuiltin`

edit/location/location.go
```diff
@@ -149,10 +156,39 @@ func showPath(path, home string) string {
 
 func (p *provider) Filter(filter string) int {
 	p.filtered = nil
-	pattern := makeLocationFilterPattern(filter)
-	for _, item := range p.all {
-		if pattern.MatchString(showPath(item.Path, p.home)) {
-			p.filtered = append(p.filtered, item)
+
```
xiaq: This is mostly duplicate with `edit/completion/candidate.go`. It is an awkward one, not very easy to factor out.

But you can always document that fact. It can serve the interest of future contributors; at some point the evaler API might have changed in way to make this easier to factor out.

edit/location/location_test.go
```diff
@@ -16,7 +17,7 @@ var (
 		{"/home/dir", 100},
 		{"/foo/\nbar", 77},
 		{"/usr/elves/elvish", 6},
-	}, "/home")
+	}, "/home", eval.NewEvaler(), matchPattern)
```
xiaq: Please add a test using an alternative matcher. The alternative matcher can be a "toy" one, e.g. one that only matches directories starting with `/usr` regardless of the pattern.

edit/location/location.go
tw4452852: Done.

edit/location/location.go
tw4452852: Done.

edit/location/location_test.go
tw4452852: Done.

PR #519, opened 2017-11-26
try to set default executable path explicitly

### LINKED ISSUE #496: Elvish cannot find its binary to start daemon
It seems that if `elvish` is not in the path when starting Elvish for the first time, it cannot find its own binary to start the daemon. I have observed this problem on macOS 10.12.6. To reproduce:

- Install Elvish using Homebrew or Nix.
- Leave user's shell to the default of /bin/bash (or something other than Elvish)
- Make sure no other copies of elvish are already running (including the elvish daemon)
- Open Terminal.app and configure "Shells open with" to the path of the elvish binary (see screenshot for an example)
general

- Start a new tab

I see an error like the following:
```
warning: cannot start daemon: cannot find elvish: not found
cannot connect to daemon after 1s: dial unix /var/folders/h0/rh70xwd16913g_2lg0rx9_300320w3/T/elvish-3212163/sock: connect: no such file or directory
Failed to initialize command history. Disabled.
```

**Workaround:** Use the `-bin` option to specify the binary location, like this: `/Users/taazadi1/.nix-profile/bin/elvish -bin /Users/taazadi1/.nix-profile/bin/elvish`

tw4452852:
fix #496

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -5,16 +5,11 @@ package daemon
 import (
 	"errors"
 	"fmt"
-	"log"
 	"os"
 	"os/exec"
-	"path"
 	"path/filepath"
 	"strconv"
-	"strings"
 	"syscall"
-
-	"github.com/elves/elvish/util"
 )
 
 // Daemon keeps configurations for the daemon sub-program. It can be used both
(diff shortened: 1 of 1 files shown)


xiaq:
Thanks, I wasn't aware of `os.Executable` before. We should definitely use it.

However, instead of changing the flag itself, please modify the `(*Daemon).Spawn` method: https://github.com/elves/elvish/blob/2a579f0d0c91d3d122dfd6535da1913b1d8584a1/daemon/daemon.go#L110. You can get rid of the `getAbsBinPath` function as well.

tw4452852:
Done.

PR #522, opened 2017-12-07
fix alt key escape in combination

### LINKED ISSUE #181: Some keybindings do not work well in tmux
For example:
Alt+{Up,Down,Up,Down} results in:

```
Unbound: Ctrl-Alt-[
Unbound: Ctrl-Alt-[
Unbound: Ctrl-Alt-[
Unbound: Ctrl-Alt-[
~/go/src/github.com/elves/elvish> [A[B[A[B                                                                                                huiyiqun@edge
```

And Ctrl+{Left,Right} is parsed as {Left,Right}

tw4452852:
fix #181

### DIFF
--- a/edit/tty/reader.go
+++ b/edit/tty/reader.go
@@ -149,6 +149,17 @@ func (rd *Reader) readOne(r rune) {
 	switch r {
 	case 0x1b: // ^[ Escape
 		r2 := readRune()
+		// According to https://unix.stackexchange.com/a/73697, rxvt and derivatives
+		// prepend another ESC to a CSI-style or G3-style sequence to signal Alt.
+		// If that happens, remember this now; it will be later picked up when parsing
+		// those two kinds of sequences.
+		//
+		// issue #181
+		hasTwoLeadingESC := false
+		if r2 == 0x1b {
+			hasTwoLeadingESC = true
+			r2 = readRune()
+		}
 		if r2 == runeTimeout || r2 == runeReadError {
 			// Nothing follows. Taken as a lone Escape.
 			unit = Key{'[', ui.Ctrl}
@@ -248,6 +259,9 @@ func (rd *Reader) readOne(r rune) {
 				if k == (ui.Key{}) {
 					badSeq("bad CSI")
 				} else {
+					if hasTwoLeadingESC {
+						k.Mod |= ui.Alt
+					}
 					unit = Key(k)
 				}
 			}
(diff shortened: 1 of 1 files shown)


edit/tty/reader.go
```diff
@@ -148,7 +148,14 @@ func (rd *Reader) readOne(r rune) {
 
 	switch r {
 	case 0x1b: // ^[ Escape
+		var prefixMod ui.Mod
 		r2 := readRune()
+		// some terminals may also send Alt character which is same as ESC
```
xiaq: Elvish already handles the use of Alt-sends-Escape in some cases (near the end of this function, search "Something other than"), and this branch is specifically for handling double ESCs. I would rather you clarify this comment to make it read like:

> According to https://unix.stackexchange.com/a/73697, rxvt and derivatives prepend another ESC to a CSI-style or G3-style sequence to signal Alt. If that happens, remember this now; it will be later picked up when parsing those two kinds of sequences.

Also, I'd rather you make this a boolean flag `hasTwoLeadingESC`; a Mod seems needlessly general.

edit/tty/reader.go
```diff
@@ -196,7 +203,7 @@ func (rd *Reader) readOne(r rune) {
 				}
 				mod := mouseModify(int(cb))
 				unit = MouseEvent{
-					Pos{int(cy) - 32, int(cx) - 32}, down, button, mod}
+					Pos{int(cy) - 32, int(cx) - 32}, down, button, mod | prefixMod}
```
xiaq: Mouse events have very different escape sequences. Do you have evidence that double ESC can arise in this case? If not, I'd rather you remove this.

edit/tty/reader.go
```diff
@@ -239,7 +246,7 @@ func (rd *Reader) readOne(r rune) {
 				down := r == 'M'
 				button := nums[0] & 3
 				mod := mouseModify(nums[0])
-				unit = MouseEvent{Pos{nums[2], nums[1]}, down, button, mod}
+				unit = MouseEvent{Pos{nums[2], nums[1]}, down, button, mod | prefixMod}
```
xiaq: Ditto.

edit/tty/reader.go
tw4452852: Ok, i have just removed this.

edit/tty/reader.go
tw4452852: Done.

edit/tty/reader.go
tw4452852: Ditto.

Issue #527, opened 2017-12-07
Don't list $pwd in LOCATION mode


notramo:
The list in LOCATION mode shouldn't include $pwd.

xiaq:
Fix by #531.

PR #531, opened 2017-12-13
Removed $pwd from showing in Directory History listing in the CTRL+L tab(LOCATION) #527

adracea:
The make passed successfully with those changes and the behavior is the intended one . Awaiting comments and feedback as this is my first ***serious*** contribution to the github&open-source community .

### DIFF
--- a/edit/location.go
+++ b/edit/location.go
@@ -148,8 +148,10 @@ func locStart(ed *Editor) {
 	// Pinned directories are also blacklisted to prevent them from showing up
 	// twice.
 	black := convertListsToSet(ed.locHidden(), ed.locPinned())
-	pwd, _ := os.Getwd()
-	black[pwd] = struct{}{}
+	pwd, err := os.Getwd()
+	if err == nil {
+		black[pwd] = struct{}{}
+	}
 	stored, err := ed.daemon.Dirs(black)
 	if err != nil {
 		ed.Notify("store error: %v", err)
--- a/edit/location.go
+++ b/edit/location.go
@@ -139,20 +139,6 @@ func (loc *location) Accept(i int, ed *Editor) {
 	ed.mode = &ed.insert
 }
 
-//This removes the $pwd from the current directory-historyDue to the way the paths are handled before this operation , we have to get all paths evaluated to the homedir , which gives
-func remPwd(dirs []storedefs.Dir, home string) []storedefs.Dir {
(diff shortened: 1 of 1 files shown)


edit/location.go
```diff
@@ -139,6 +139,20 @@ func (loc *location) Accept(i int, ed *Editor) {
 	ed.mode = &ed.insert
 }
 
+//This removes the $pwd from the current directory-historyDue to the way the paths are handled before this operation , we have to get all paths evaluated to the homedir , which gives
```
xiaq: You can just add pwd to the blacklist on line 164. There is no need to convert the directory values to absolute paths either; the directory entries in storage always contain results from `os.Getwd` (c.f. https://github.com/elves/elvish/blob/master/eval/chdir.go).

xiaq:
Welcome to the open source community! :)

edit/location.go
adracea: I'll look into it tomorrow morning . Thanks for the warm welcome :smile: .

adracea: Okay , I figured out why on earth my initial commit wasn't working as intended when I started trying it out at home . So : util.Getwd() -- the function that returns just the string of the pwd , returns it as it gets displayed : as "~/go/src/elves/elvish" but ; os.Getwd() returns it as "/home/user/go/src/elves/elvish" . I was trying to put util.Getwd() in the blacklist while the directory listing had os.Getwd() history in it .
In the current form , I would overwrite the current blacklist/pinned list .
@xiaq How would one blacklist a folder ?
 I'll resubmit a full solution tomorrow morning .

edit/location.go
```diff
@@ -148,6 +148,8 @@ func locStart(ed *Editor) {
 	// Pinned directories are also blacklisted to prevent them from showing up
 	// twice.
 	black := convertListsToSet(ed.locHidden(), ed.locPinned())
+	pwd, _ := os.Getwd()
```
xiaq: Please handle the possible error.

If there was an error, you can simply not add `pwd` to the blacklist.

xiaq:
Thank you! :)

notramo:
Thanks.

adracea:
Thanks for including it ! :)
Glad I could contribute to elvish since I've been using it for a while now both at home(+phone) and at work ! I'll keep contributing however I can around here . :smiley:

PR #536, opened 2017-12-17
Changed command keybindings 0 as I and $ as A like vim.

themester:


xiaq:
As discussed on Telegram, this behaviour does not match Vim.

xiaq:
Suggested alternative to bind `I` and `A` is doing it with Elvishscript:

```
edit:binding:command[I] = { edit:move-dot-sol; edit:insert:start }
edit:binding:command[A] = { edit:move-dot-eol; edit:insert:start }
```

PR #541, opened 2017-12-19
Fixed narrow.elv and readline-binding.elv builtins to the new funcref syntax.

zzamboni:


xiaq:
Thanks! I also fixed it at roughly the same time though :)

PR #547, opened 2017-12-28
Added more install-function variations to epm

zzamboni:
Added:

- `epm:is-installed $pkg` to check if a package is already installed
- `epm:install-if-needed @pkgs` to install packages only if they are not
  already installed, be silent otherwise. This function makes it easier
  to install packages directly from the config file, but without
  printing noise if the packages are already there.
- `epm:install-or-upgrade @pkgs` to install packages if they are not
  installed, try to upgrade them otherwise.

### DIFF
--- a/eval/bundled_modules.go
+++ b/eval/bundled_modules.go
@@ -28,9 +28,14 @@ fn -get-url [pkg]{
     put https://$pkg
 }
 
+fn is-installed [pkg]{
+  dest = $-data-dir/lib/$pkg
+  put ?(test -e $dest)
+}
+
 fn -install-one [pkg]{
     dest = $-data-dir/lib/$pkg
-    if ?(test -e $dest) {
+    if (is-installed $pkg) {
         -error 'Package '$pkg' already exists locally.'
         return
     }
@@ -58,7 +63,7 @@ fn installed {
 
 fn -upgrade-one [pkg]{
     dest = $-data-dir/lib/$pkg
-    if (not ?(test -d $dest)) {
+    if (not (is-installed $pkg)) {
         -error 'Package '$pkg' not installed locally.'
         return
     }
@@ -74,6 +79,40 @@ fn upgrade [@pkgs]{
     for pkg $pkgs {
         -upgrade-one $pkg
     }
+}
+
+fn -install-or-upgrade-one [pkg]{
+  if (is-installed $pkg) {
+    -upgrade-one $pkg
+  } else {
+    -install-one $pkg
+  }
+}
+
+fn install-or-upgrade [@pkgs]{
+  if (eq $pkgs []) {
+    -error 'Must specify at least one package.'
+    return
+  }
+  for pkg $pkgs {
+    -install-or-upgrade-one $pkg
+  }
+}
+
+fn -install-if-needed-one [pkg]{
+  if (not (is-installed $pkg)) {
+    -install-one $pkg
+  }
+}
+
+fn install-if-needed [@pkgs]{
+  if (eq $pkgs []) {
+    -error 'Must specify at least one package.'
+    return
+  }
+  for pkg $pkgs {
+    -install-if-needed-one $pkg
+  }
 }
 
 fn -uninstall-one [pkg]{
@@ -123,7 +162,7 @@ fn uninstall [@pkgs]{
 # but different keys can be specified with the options. To disable a
 # binding, specify its key as "".
 # Example:
-#   narrow:bind-trigger-keys &location=Alt-l &lastcmd="" 
+#   narrow:bind-trigger-keys &location=Alt-l &lastcmd=""
 
 # Hooks
 # Each hook variable is an array which must contain lambdas, all of
--- a/eval/epm.elv
+++ b/eval/epm.elv
@@ -25,9 +25,14 @@ fn -get-url [pkg]{
     put https://$pkg
 }
 
+fn is-installed [pkg]{
+  dest = $-data-dir/lib/$pkg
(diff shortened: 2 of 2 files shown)


zzamboni:
@xiaq I'm thinking these should probably be options to `epm:install` instead of separate functions - what do you think?

xiaq:
Hi, thanks for the PR!

The concept of "installed" is a bit complicated now. A package can be

1. Completely absent.

2. Present in the filesystem (the `lib` directory), but not in the database(the `installed` file). This 
arises when the user has some local packages before starting to use `epm`.

3. Present in the database, but not in the filesystem. This arises when the user attempts to remove the package by removing the package without using `epm:uninstall`.

4. Present in both.

State 4 is definitely "installed", and it is a state that the `install` command should bring a package to. The implementation of the `installed` command admits 3 as installed but not 2; the `is-installed` you added admits 2 as installed but not 3.

To deal with the different states, we can

* Rename the `installed` command to `list-all-in-db`;

* Rename the `is-installed` command you are adding to `in-fs`;

* Add an `in-db` command. For completeness there should also be a `list-all-in-fs` command; however, it is a bit tricky to implement for a reason I'll come back to, so let's not worry about this;

The `install` and `upgrade` commands can work as follows:

*   The `install` command should be able to take a package that is in any of state 1, 2 or 3, and bring it into state 4. It is actually simpler than it may sound, it is just

    ```
    if (not (in-fs $pkg)) {
        info "cloning git repo"
        # clone git repo
    }
    if (not (in-db)) {
        info "recording in db"
        # record in db
    }
    ```

    We can add a `&quiet` option to `install` to suppress these messages.

*   In a similar spirit, we can implement `upgrade` simply as "install and update repo"; that is, it can take a package in any of state 1-4 and upgrade it.

zzamboni:
@xiaq thanks for the detailed analysis. Is there a reason to keep the database and not just use the filesystem as the source of truth? I think this would simplify the state checking and allow direct manipulation. The listing of installed modules would be as simple as `cd ~/.elvish/lib; ls -d */*/*`.

xiaq:
Sorry, I intended to explain the problem in my last post but forgot.

The idea of just using the directory as the database is a good one. The problem is that neither Elvish's module system nor `epm` cares about which directory level is the "package level". Let's say that the `lib` directory is level 0, then packages hosted on Github happen to be in level 3; this is because Github repo URLs look like `github.com/$user/$pkg`. If you host a package on your own website, the URL can look like `zzamboni.org/$pkg`, and then your package lives in level 2. It all depends on the setup of the Git server; you can even set up your server to serve a Git repo at the root level of subdomains, and your package URL can be just `$pkg.zzamboni.org`.

Hence, a naive directory-level-based system cannot work if we want to keep the URL <-> directory mapping (which I see as one of the major features of `epm`).

There are other ways to make it work though. One idea is to require packages to have some kind of metadata file. When listing packages, we recurse into the directory tree until we have found a metadata file; this has a drawback, however, if the user has a deeply nested directory tree without metadata files; then we'll be wasting time looking for metadata files in this tree.

One possible way to remedy this would be to create special markers in the parent directories of packages. For instance, when the user first installs `github.com/elves/sample-pkg`, we can create special marker files in `github.com` and `github.com/elves`. Yet another alternative is to create a metadata file only in `github.com`, which contains, possibly among other things, how many more levels you need to go to reach packages (for github.com, it is 2).

xiaq:
Hmm, keeping a metadata file in the `github.com` level makes more sense. It can record not just the package level, but also e.g. the version control system this domain uses (right now `epm` assumes git). Making the level explicit also makes the package metadata file optional. It has the slight drawback that, when configuring your own server, you have to serve packages at a fixed level, but that is good for consistency anyway.

My preferred design now looks like this:

*   Each domain has a domain config file called, say, `epm-domain.cfg`. For github.com, this file is `~/.elvish/lib/github.com/epm-domain.cfg`. It specifies the VCS this domain uses and the level packages are to be found. `epm` is bundled with domain configs for popular code hosting websites like GitHub, and populate the domain config automatically when the user first installs a package hosted there. The user can, of course, create a domain config manually.

    The domain config opens up many possibilities. For instance, when specifying the VCS, we can not only let the user choose from a predefined set (git, hg, ...), but also give the user complete control and specify the command to install and update packages; enabling the use of arbitrary VCS.

* When listing packages, `epm` looks at each directory in `~/.elvish/lib`; if a directory contains a domain config, it retrieves the package level number `$l` (for github.com, 2) and recurses exactly `$l` levels.

zzamboni:
Interesting idea - indeed opens up a lot of possibilities. One idea: make the domain file be in Elvish, e.g. `~/.elvish/lib/github.com/epm-domain.elv`, then we can use familiar syntax in it, and it can be loaded with `use github.com:epm-domain` both by `epm` and by the user, getting parsing for free.

zzamboni:
As I was trying to take a quick stab at this, I realized that you cannot use a variable in the argument to `use`, so my idea of having the domain config be an Elvish file is not very practical. Maybe it's better for it to be JSON, which can be read directly using `from-json`.

zzamboni:
@xiaq I've made an initial implementation of the method you described above, at https://github.com/zzamboni/dot_elvish/blob/master/lib/testing/epm.elv. For now it knows how to handle git, and has a built-in configuration for github.com. Handlers can be extended in the `-method-handler` variable (which specified both install and upgrade), and built-in domains in `-default-domain-config`.

Let me know what you think!

zzamboni:
Closing this PR, will open another one with the new code.

PR #563, opened 2018-01-14
File-based epm implementation

zzamboni:
Reimplementation of epm to be fully file-based, with per-domain configuration for different package installation methods. The scheme is described in https://github.com/elves/elvish/pull/547#issuecomment-354560975. In short:

- Packages are specified as `domain/package`. For example, `github.com/zzamboni/modules.elv`
- Packages are stored under `~/.elvish/lib` under their full path. For example, `~/.elvish/lib/github.com/zzamboni/modules.elv/`
- Each domain has a configuration file stored in `~/.elvish/lib/<domain>/epm-domain.cfg`. Currently epm has built-in domain configuration for the `github.com`, `gitlab.com` and `bitbucket.org` domains. This file is a JSON file which contains the following fields:

  - `method` specifies the method used to transfer the module files. Currently supported are `git` and `rsync`.
  - `levels` specifies how many directories under the domain directory are packages located. For example, the github.com domain has `levels` set to 2, which means the packages are directories stored 2 levels under `~/.elvish/lib/github.com`.
  - Other fields may be specified, depending on the method. For example, method `git` needs `protocol` (e.g. `https` or `git`) and method `rsync` needs `location` (the location from which to rsync the files).

For domains with built-in configuration (github.com, gitlab.com and bitbucket.org at the moment), the `epm-domain.cfg` file is created automatically the first time a package from that domain is installed. For others, you have to create the file by hand. For example, to have a "dev" domain which installs packages specified as `dev/foo` from `~/devel/foo`, you must create file `~/.elvish/lib/dev/epm-domain.cfg` containing the following:

```json
{
   "method" : "rsync",
   "location" : "/Users/taazadi1/Personal/devel/elvish/dev",
   "levels" : "1"
}
```

### DIFF
--- a/eval/bundled/epm.elv.go
+++ b/eval/bundled/epm.elv.go
@@ -26,10 +26,6 @@ debug-mode = $false
 -data-dir = ~/.elvish
 -lib-dir = $-data-dir/lib
 
-# Runtime state - records get copied from -default-domain-config
-# or read from epm-domain.cfg files as needed
--domain-config = [&]
-
 # General utility functions
 
 fn -debug [text]{
@@ -63,17 +59,24 @@ fn is-installed [pkg]{
 }
 
 fn -package-domain [pkg]{
-  splits / $pkg | take 1
+  splits &max=2 / $pkg | take 1
 }
 
 fn -package-without-domain [pkg]{
-  joins / [(splits / $pkg | drop 1)]
+  splits &max=2 / $pkg | drop 1 | joins ''
 }
 
 # Known method handlers. Each entry is indexed by method name (the
 # value of the "method" key in the domain configs), and must contain
 # two keys: install and upgrade, each one must be a closure that
 # received two arguments: package name and the domain config entry
+#
+# - Method 'git' requires the key 'protocol' in the domain config,
+#   which has to be `http` or `https`
+# - Method 'rsync' requires the key 'location' in the domain config,
+#   which has to contain the directory where the domain files are
+#   stored. It can be any source location understood by the rsync
+#   command.
 -method-handler = [
   &git= [
     &src= [pkg dom-cfg]{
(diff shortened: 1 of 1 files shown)


xiaq:
I took a brief look and it looks good! I’m still kind of busy ATM but I’ll get back with feedback soon.

zzamboni:
Thanks! No rush, I look forward to your comments.

eval/bundled/epm.elv.go
```diff
@@ -1,111 +1,279 @@
...
+# Internal configuration
 -data-dir = ~/.elvish
--installed = $-data-dir/epm-installed
+-lib-dir = $-data-dir/lib
 
-fn -info [text]{
-    print (edit:styled '=> ' green)
+# Runtime state - records get copied from -default-domain-config
+# or read from epm-domain.cfg files as needed
+-domain-config = [&]
```
xiaq: I think it's better to **not** cache domain configurations. Caching means that if the user edits a domain config, the change may or may not take effect, depending on whether that particular config has been cached. Just read them every time.

eval/bundled/epm.elv.go
```diff
@@ -1,111 +1,279 @@
...
-    }
-    for pkg $pkgs {
-        -upgrade-one $pkg
+# Return the filename of the metadata file for the given package
+# (regardless of whether it exists)
+fn -package-metadata-file [pkg]{
+  put (dest $pkg)/metadata.json
+}
+
+# Read the domain config file for a given domain. If the file does not
```
xiaq: Not caching means that this function can now be rewritten as one that outputs the config instead of storing it.

eval/bundled/epm.elv.go
```diff
@@ -1,111 +1,279 @@
...
+    cfg = $-domain-config[$dom]
+    method = $cfg[method]
+    if (has-key $-method-handler $method) {
+      if (has-key $-method-handler[$method] $what) {
+        $-method-handler[$method][$what] $pkg $cfg
+      } else {
+        fail "No handler for '"$what"' defined in method '"$method"'"
+      }
+    } else {
+      fail "No handler defined for method '"$method"', specified in in config file "(-domain-config-file $dom)
```
xiaq: This message can be made a bit terser to `Unknown method $method`, as the concept of handler is not very relevant here. Also, `in` appeared twice :)

eval/bundled/epm.elv.go
```diff
@@ -1,111 +1,279 @@
...
+      -package-op $pkg upgrade
     }
+  }
 }
-`
+
+# Uninstall is the same for everyone, just remove the directory
+fn uninstall [@pkgs]{
+  if (eq $pkgs []) {
+    fail 'Must specify at least one package.'
```
xiaq: Use "-error" to be consistent with other commands.

eval/bundled/epm.elv.go
```diff
@@ -1,111 +1,279 @@
...
+fn -package-domain [pkg]{
+  splits / $pkg | take 1
 }
 
-fn install [@pkgs]{
-    if (eq $pkgs []) {
-        -error 'Must specify at least one package.'
-        return
+fn -package-without-domain [pkg]{
+  joins / [(splits / $pkg | drop 1)]
```
xiaq: This can be: `splits / $pkg | drop 1 | joins /`

You can also avoid unnecessary splitting beyond the first / by

`splits &max=2 $pkg | drop 1 | joins ''`

You still need the last `joins ''` to correctly handle package names that do not contain any slashes (although I wonder whether that case is handled correctly everywhere else.)

eval/bundled/epm.elv.go
```diff
@@ -1,111 +1,279 @@
...
-    if ?(test -e $dest) {
-        -error 'Package '$pkg' already exists locally.'
-        return
-    }
-    -info 'Installing '$pkg
-    mkdir -p $dest
-    git clone (-get-url $pkg) $dest
-    add-installed $pkg
+fn -package-domain [pkg]{
+  splits / $pkg | take 1
```
xiaq: `splits &max=2 / $pkg | take 1`

eval/bundled/epm.elv.go
```diff
@@ -1,111 +1,279 @@
...
+######################################################################
+# Main user-facing functions
+
+# List installed packages
+fn installed {
+  e:ls $-lib-dir | each [dom]{
+    if ?(test -f (-domain-config-file $dom)) {
+      -read-domain-config $dom
+      lvl = $-domain-config[$dom][levels]
+      find $-lib-dir/$dom -type d -depth $lvl | each [pkg]{
```
xiaq: I would prefer that we do not rely on `find`, but it's OK for now.

eval/bundled/epm.elv.go
zzamboni: Thanks for the tip about `splits | ... | joins`!

The package-without-slashes case is handled correctly - I have locally an `rsync` package with a single level, package names like `dev/epm`, and they work fine. The minimum is `domain/package`.

eval/bundled/epm.elv.go
zzamboni: Any ideas about how to do this without `find`? I tried this, but it doesn't work (outputs the asterisks instead of expanding them), and it's not any shorter:
```
      put $-lib-dir/(repeat $lvl '*' | joins /) | each [pkg]{
        replaces $-lib-dir/ "" $pkg
      }
```
Another idea would be to `put $-lib-dir/**` and then filter by number of slashes (for the level) and by is-a-directory, but that seems even clunkier.

eval/bundled/epm.elv.go
zzamboni: Makes sense. `-error` for "regular" errors (usage) and `fail` for "internal" errors (lack of default config for a domain, etc.)

eval/bundled/epm.elv.go
zzamboni: Makes sense, working on it. I was also thinking: does it makes sense to create the file the first time with the default config, for known domains? If we don't, then we can just return the default config every time, which makes it easier to change the default configs in the module. It would also simplify the code: either read-and-return, or just return the default. No creating files.

xiaq: Aha, good point. I agree that we don't have to write out the default config.

eval/bundled/epm.elv.go
xiaq: I was thinking of root-level packages, like just `zzamboni.org`. It's a question whether such packages should be supported at all. If we don't, we should check that package names contain slashes.

If we know that the package name will always contain at least one slash, the trailing `| join ''` will not be necessary.

eval/bundled/epm.elv.go
xiaq: Globbing a generated pattern is not supported at the moment, so an implementation without using `find` will need to traverse the filesystem recursively:

```
fn subdirs [dir level]{
  if (eq $level 0) {
    put $dir
  } else {
    put $dir/*/ | each [subdir]{ subdirs $subdir (- $level 1) }
}
```

zzamboni:
@xiaq I've addressed all your comments - thanks for the feedback!

xiaq:
Thank you! :)

zzamboni:
Awesome! Next up I will work on updating the `epm` documentation.

PR #565, opened 2018-01-15
eval/compiler: fix builtin lookup

xofyarg:


xofyarg:
Declining since one could re-declare the module in local NS to make it work.

PR #575, opened 2018-01-21
Various epm updates

zzamboni:
Various epm improvements, including:

- Force output of `is-installed` to `$true` or `$false`, just for cleanliness
- The `metadata` function now returns not only package-provided metadata (if any) but also some base attributes such as package name, install status, source and destination (if installed), and install method.
- The `query` function uses the improved `metadata` function, and prettified output.
- Rudimentary dependency management:   If an installed package has a metadata file, and if the metadata has a    `dependencies` key, then it is expected to be a list of strings    indicating packages that must also be installed.    There is still no error handling for this. If the    installation of dependencies fails, should we just uninstall everything?    Feedback welcome.

### DIFF
--- a/eval/bundled/epm.elv.go
+++ b/eval/bundled/epm.elv.go
@@ -55,7 +55,7 @@ fn dest [pkg]{
 }
 
 fn is-installed [pkg]{
-  put ?(test -e (dest $pkg))
+  put (or (and ?(test -e (dest $pkg)) $true) $false)
 }
 
 fn -package-domain [pkg]{
--- a/eval/bundled/epm.elv.go
+++ b/eval/bundled/epm.elv.go
@@ -228,15 +228,18 @@ fn metadata [pkg]{
 # Print out information about a package
 fn query [pkg]{
   data = (metadata $pkg)
-  echo (edit:styled "Package "$pkg cyan)
-  echo (edit:styled "Source:" blue) (-package-method $pkg) (-package-op $pkg src)
-  if (is-installed $pkg) {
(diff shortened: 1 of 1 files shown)


eval/bundled/epm.elv.go
```diff
@@ -55,7 +55,7 @@ fn dest [pkg]{
 }
 
 fn is-installed [pkg]{
-  put ?(test -e (dest $pkg))
+  put (or (and ?(test -e (dest $pkg)) $true) $false)
```
xiaq: There is a `bool` [builtin](https://elvish.io/ref/builtin.html#bool) for this :)

eval/bundled/epm.elv.go
```diff
@@ -199,28 +199,47 @@ fn -package-op [pkg what]{
...
+  # Base metadata attributes
+  res = [
+    &name= $pkg
+    &method= (-package-method $pkg)
+    &src= (-package-op $pkg src)
+    &dst= (dest $pkg)
+    &installed= (is-installed $pkg)
+  ]
+  # Merge with package-specified attributes, if any
+  mdataf = (-package-metadata-file $pkg)
```
xiaq: Consider making it more verbose, like `metadata-filename`. Then, since this variable is local to `metadata` function, just call it `filename`.

eval/bundled/epm.elv.go
zzamboni: Duh - of course, thanks for the reminder! :)

zzamboni:
Thanks for the feedback! Addressed both points.

PR #606, opened 2018-02-23
sys: cleanup Termios.Lflags 64/32 bits type

zhsj:


xiaq:
I think the current approach is the lesser evil here. Losing type safety is a bigger loss than having to look up how big `Lflags` is on each platform.

Have you discovered an architecture not covered yet by the current two files?

zhsj:
> Have you discovered an architecture not covered yet by the current two files?
>

Not yet. Currently elvish can't build on other architectures due to
dependencies like boltdb.

PR #609, opened 2018-02-27
eval/builtin_fn_misc: Add environment manipulation commands

wkennington:
Sometimes we want to be able to reference environment variables by a key
that is specified at runtime. Using a builtin command instead of the E:
namespace makes that simple and very similar to other programming
languages.

### DIFF
--- a/eval/builtin_fn_misc.go
+++ b/eval/builtin_fn_misc.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"net"
+	"os"
 	"path/filepath"
 	"time"
 	"unicode/utf8"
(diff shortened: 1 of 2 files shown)


eval/builtin_fn_misc.go
```diff
@@ -60,6 +63,19 @@ func constantly(args ...interface{}) Callable {
 	)
 }
 
+func getenv(fm *Frame, key string) error {
```
xiaq: This function doesn't need to take a `Frame`, it can return `value`. Non-error return values are converted to output.

eval/builtin_fn_misc.go
```diff
@@ -60,6 +63,19 @@ func constantly(args ...interface{}) Callable {
...
+	value, exists := os.LookupEnv(key)
+	if !exists {
+		return errors.New("Non-existent environment variable")
+	}
+	fm.ports[1].Chan <- value
+	return nil
+}
+
+func setenv(fm *Frame, key, value string) error {
```
xiaq: Frame is also optional here. In fact, this entire function can be removed, just use os.Setenv.

eval/builtin_fn_misc.go
wkennington: Yeah, this makes sense

eval/builtin_fn_misc.go
wkennington: Done

wkennington:
Made the requested changes, please take a second look now

zzamboni:
This is very nice. One comment: there is a difference in behavior between `$E:VAR` and `get-env VAR` when the variable does not exist:
```
> get-env FOO
Exception: non-existent environment variable
[tty], line 1: get-env FOO
> put $E:FOO
▶ ''
```

I would suggest the exception is better to be consistent with other variable accesses in Elvish.

BTW: is there a way to check if a variable exists?

PR #628, opened 2018-03-09
Don't crash when git update fails

### LINKED ISSUE #626: epm:upgrade crashes if something in libs is not a git repo
should be handled gracefully

zzamboni:
When git update fails for any reason (i.e. not a git repo, unstaged changes, etc.), catch the exception and print an error, but don't crash.

Fixes #626

### DIFF
--- a/eval/bundled/epm.elv.go
+++ b/eval/bundled/epm.elv.go
@@ -111,7 +111,12 @@ fn tilde-expand [p]{
     &upgrade= [pkg dom-cfg]{
       dest = (dest $pkg)
       -info "Updating "$pkg
-      git -C $dest pull
+      is-ok = ?(out = [(git -C $dest pull)])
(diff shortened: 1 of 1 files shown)


eval/bundled/epm.elv.go
```diff
@@ -111,7 +111,12 @@ fn tilde-expand [p]{
     &upgrade= [pkg dom-cfg]{
       dest = (dest $pkg)
       -info "Updating "$pkg
-      git -C $dest pull
```
xiaq: what about using `try`:

```
try {
  git -C $dest pull
} except _ {
  -error "git pull failed"
}
```

zzamboni: Of course! Much simpler code. Thanks for the feedback - done.

Issue #520, opened 2017-12-01
Make `edit:styled` more high-level and allow nesting


xiaq:
Right now, `edit:styled` is tied to the terminal: There is a one-to-one relationship ([code](https://github.com/elves/elvish/blob/master/edit/ui/styled.go#L16)) mapping the supported styles to ANSI color sequences (`"\e[1;31m"`, where `1;31` can be any semicolon-separated list of numbers), and it is possible to specify a number (or multiple numbers separated by semicolons) as the "style".

This is an improvement over writing ANSI color sequences by hand, but it still has a few drawbacks:

1. It is tied to the terminal. In future, Elvish will support a web frontend where ANSI color sequences are not directly supported. It is possible to parse and convert them to CSS, but that is backwards. If we start with a high-level internal format, it can then support multiple presentations, including both ANSI color sequences and CSS.

2. It does not support nesting. For instance, the ANSI color sequence for reverse video is `"\e[7m"`. Two inverses cancel each other: so `echo (edit:styled (edit:styled lorem inverse) inverse)` should print `lorem` in normal style. However, without knowing the semantics of those different styles, it is not possible to compose them. Again, we can hack composing into the current implementation, but it is cleaner to do it from the source by having a semantics-based internal format.

Another minor thing is that since this function is useful outside editors, it is appropriate to put it in the builtin namespace, making the constructor `styled`.

Current design is as follows:

1.  A **styled text** is internally a list of **styled segments**; a segment is a string, combined with several attributes: `fg-color`, `bg-color`, `inverse`, `bold`, etc. It is possible to construct a segment with `styled-segment`, specifying the attributes as options, e.g. `styled-segment lorem &fg-color=red &bold=$true`. If the argument is already a styled segment, `styled-segment` copies it and modifies the specified attributes. Attributes of styled segments can be accessed like map elements, e.g. `$seg[bold]`.

2.  It is possible to concatenate plain strings, styled segments and styled texts together: they will form a styled text that combines them, e.g. `echo (styled-segment lorem &fg-color=red)(styled-segment ipsum &inverse=$true)foobar`. No implicit whitespace is inserted between concatenated elements.

[...]

PR #633, opened 2018-03-13
Make value concatenation more modular

fehnomenal:
While working on #520 I noticed the concatenation of values from the input might get more and more complex over time.
This is an attempt to make the process at least a little bit more modular.


fehnomenal:
After thinking about it, I belief having a central registry that maps an ordered tuple of types to a concatenating functions makes more sense.
I will come up with something tomorrow.

xiaq:
Sorry for the late feedback. I applaud your take at extensibility, but using reflect to do double dispatch looks overkill to me.

I think that Python's approach is sufficient: to overload the addition operator in `lhs + rhs`, Python supports a `__radd__` method in addition to `__add__`. If calling `lhs.__add__(rhs)` returns a special `NotImplemented` error Python will try `rhs.__add__(lhs)`. We can do virtually the same thing; define

```go
type Catter interface {
  Cat(interface{}) ({interface{}, error)
}

type RCatter interface {
  RCat(interface{}) ({interface{}, error)
}

var ErrCatNotImplemented = errors.New("cat not implemented")
```

xiaq:
BTW, there is no need for the mutex in the code as long as `Register` is not called in parallel with `Concat`. Concurrent reads of maps are safe.

Also, on second thought, the interface and methods should be called `Concatter`, `Concat`, `RConcat`, `RConcatter`. I was thinking to abbreviate the function name as well to just `Cat`, but that will be very awkward to expose as an Elvish builtin. Simply calling everything `concat` is better.

Also, please move the `Concat` function to `eval/vals`. There is no need to create another package.

fehnomenal:
Thank you for your response. Your proposal would essentially be what I tried in my first commit.
I think it has two drawbacks:
1. There still might be the need for type switches inside the implementations of `Concat`.
2. Builtin types like `string` need a wrapper and special handling.

eval/vals/concat.go
```diff
@@ -0,0 +1,51 @@
+package vals
+
+import (
+	"errors"
+	"fmt"
+)
+
+type (
```
xiaq: I would rather not use a `type` block, and have those types documented. But I'm not delaying this PR futher.

xiaq:
Yes you need to do a type switch. This is a style used in most of the functions in `eval/vals`. Go doesn't support advanced type dispatch mechanisms out of box, there is always some overhead involved if you need to do such things.

But I see you implemented this anyway :)

PR #635, opened 2018-03-14
Expand has-key functionality

### LINKED ISSUE #631: has-key does not work on namespaces
(mentioned in #492, but opening here for follow up)

In my tests, `keys` works fine on namespaces, but `has-key` does not:

```
> kind-of $edit:
▶ ns
> keys $edit:
▶ move-dot-eol~
▶ rprompt-persistent
▶ match-prefix~
▶ complex-candidate~
▶ selected-file
...
> has-key $edit: after-readline
Exception: couldn't get key or index of type 'ns'
```

fehnomenal:
Fixes #631


xiaq:
A good practical solution for now, but I would like `has-key` to eventually work with an interface; enumerating all keys in a namespace is not very efficient.

Issue #580, opened 2018-02-02
Preserve order of completions 


occivink:
Currently completions are sorted alphabetically before being shown. 

It'd be more powerful to preserve the order in which the completer output them, so that one could order them according to how relevant they are.

As potential usecases, I would like the filename completer to show directories first, and my `git add` completer to show files already tracked before the untracked ones.

Issue #525, opened 2017-12-07
Add custom command name completion


notramo:
Completion function is configurable for arguments, but not for executable/function names.
Elvish should have a command completion command, similar to the ones given in `edit:arg-completer`. I think `edit:command-completer` would be a good name for it.

zzamboni:
@notramo What would be a possible use for this? (I'm genuinely curious) One possibility I can think of is to have directory-specific completions, e.g. for project-specific scripts. Do you have others?

xiaq:
This sounds logical but it's not going to be very useful. Elvish has fixed rules for resolving commands and there is no customization point in changing that; this is a delibrate design choice. Since you cannot customize the rules for finding commands, customizing how to complete commands doesn't make much sense.

Do you have a particular use case in mind?

notramo:
I want to make a tool that completes command by recency, frequency, and $pwd.
I am coding a lot with Crystal, and I would like to see the compiler as the first suggestion. (When I type `c`, the suggestions fills my terminal, when I type `cr`, still there are lots of suggestion, but even with `cry` there are some)
`$pwd` based command name completion can also be handy if someone codes in e.g. Crystal and Rust. When `$pwd` is a Rust project dir, typing `c<Tab>` would put `cargo` in the first place, but when `$pwd` is a Crystal project dir, `c<Tab>` would give `crystal` in the first place.

My idea to solve the command resolution problem:
There should be a builtin Elvish command that lists all resolved commands (functions, external executables, variables, etc.).
Then the user could write a function that converts it e.g. JSON, or build a command line with the candidates. In these examples, the `edit:list-commands` function lists all the possible commands.
```elvish
fn my-custom-command-completer [command]{ edit:list-commands | to-json | completer-executable $command }
```
or
```elvish
fn my-custom-command-completer [command]{ completer-executable --current $command --all ( edit:list-commands | each [i]{ print $i' ' ) }
```

xiaq:
That is a very good use case, I like it.

Although instead of allowing the user to customize the entire completion for commands, I prefer to support this customization by adding an additional sorting phase and supporting customization for that. Elvish's completion is already divided into a candidate-generation phase and a matching phase, both of which can be customized separately. The motivation is orthogonality and flexibility for customization; adding a sorting phase (and relevant customization API) is coherent with the design.

Issue #456, opened 2017-07-22
Add number types


alexherbo2:
Considering bare numbers as numbers and quoted numbers as strings.

xiaq:
Number type is OK, but number literals are problematic.

You don't want number literals (`1`, `1.0` or `-1`) to be an entirely different thing from strings, because they could well be filenames or command arguments. If naming a file `-1` sounds rare, beware that [there](https://manpages.debian.org/unstable/wget/wget.1.en.html) [are](https://manpages.debian.org/unstable/curl/curl.1.en.html) [numerous](https://manpages.debian.org/stretch/iproute2/ip.8.en.html) [commands](https://manpages.debian.org/unstable/netcat-openbsd/nc.1.en.html) that accept `-4` and `-6` as switches, and [man](https://manpages.debian.org/unstable/man-db/man.1.en.html) accepts an optional section number argument.

To accomodate such usages, you inevitably need to give number literals a double identity of both number and string. This double identity leads to (among other things) problems in equality semantics: are `1` and `1.0` equal? They are equal as numbers, but not as strings; how do you tell which one the user intends? If we were to support hexadecimal number literals, are `10` and `0xa` equal?

Distinct types and fuzzy equality semantics are not a good mix. We already know this from the `==` operator in JavaScript. Elvish now solves this problem by not having number literals at all, and let a set of arithmetic commands parse their arguments into numbers. This is a less magical and less confusing approach.

alexherbo2:
### Comparing bare numbers (Numbers)

``` elvish
== 1 1.0 → true
```

### Comparing quoted numbers (Strings)

``` elvish
== '1' '1.0' → false
```

xiaq:
For an external command `x`, should `x 1` and `x 1.0` behave the same?

This is an open design question, but I strongly prefer that they should behave differently, since `1` and `1.0` can be different filenames or otherwise have different meanings, and they are common enough that I don't want to require them to be quoted.

Then we end up with the situation where `1` and `1.0` are considered equal (`== 1 1.0` outputs `$true`), but do not always behave the same, which contradicts the fact that they are equal. Technically it can be made to work, but it's bad design.

alexherbo2:
We could draw inspiration from [String interpolation][].

[Ruby / Crystal][] automatically convert objects to their string representation.

### Converting numbers to strings (Elvish)

``` elvish
str 1.0 → "1.0"
```

[String interpolation]: https://en.wikipedia.org/wiki/String_interpolation
[...]

PR #642, opened 2018-03-16
edit/completion: add sorter

tw4452852:
Just got inspiration from sort package, user can setup a comparion function (like the `Less` method in sort),
and its definition is: `func isBefore (seed, a, b string) bool`, its result determines whether candidate `a` should be placed before candidate `b`. 

For issue #580 , a simple directory first sorter would be: `[seed a b]{ put ?(test -d $a) }`,  you can also create any sophisticated sorter for sure.

BTW, now custom sort is a little slow, I will try to figure out the potention reason and improve it.

Thanks.

### DIFF
--- a/edit/completion/sorter.go
+++ b/edit/completion/sorter.go
@@ -1,0 +1,66 @@
+package completion
+
+import (
+	"errors"
+	"os"
+	"sort"
+
+	"github.com/elves/elvish/eval"
+	"github.com/elves/elvish/eval/vals"
+	"github.com/xiaq/persistent/hashmap"
+)
+
+var (
+	errSorterMustBeFn = errors.New("sorter must be a function")
+)
+
+type rawCandidates []rawCandidate
+
+func (cs rawCandidates) Len() int           { return len(cs) }
+func (cs rawCandidates) Swap(i, j int)      { cs[i], cs[j] = cs[j], cs[i] }
+func (cs rawCandidates) Less(i, j int) bool { return cs[i].text() < cs[j].text() }
+
+func lookupSorter(m hashmap.Map, name string) (eval.Callable, bool) {
+	key := name
+	if !hashmap.HasKey(m, key) {
+		// Use fallback sorter
+		if !hashmap.HasKey(m, "") {
+			return nil, false
+		}
+		key = ""
+	}
+	value, _ := m.Index(key)
+	sorter, ok := value.(eval.Callable)
+	return sorter, ok
+}
+
+func sortRawCandidates(ev *eval.Evaler, customSorter eval.Callable, seed string, rs []rawCandidate) ([]rawCandidate, error) {
+	// default sorter
+	if customSorter == nil {
+		sort.Sort(rawCandidates(rs))
+		return rs, nil
+	}
+
+	ports := []*eval.Port{
+		eval.DevNullClosedChan,
+		{}, // Will be replaced when capturing output
+		{File: os.Stderr},
+	}
+	ec := eval.NewTopFrame(ev, eval.NewInternalSource("[editor sorter]"), ports)
+	sort.SliceStable(rs, func(i, j int) bool {
+		args := []interface{}{seed, rs[i].text(), rs[j].text()}
+		rets, err := ec.CaptureOutput(customSorter, args, eval.NoOpts)
+		if err != nil {
+			logger.Printf("sorter: %v", err)
+			return false
+		}
+		// just return one bool
+		if n := len(rets); n != 1 {
+			logger.Printf("sorter: return value count %d != 1", n)
+			return false
+		}
+		return vals.Bool(rets[0])
+	})
+
+	return rs, nil
+}
--- a/edit/completion/candidate.go
+++ b/edit/completion/candidate.go
@@ -3,7 +3,6 @@ package completion
 import (
 	"fmt"
 	"os"
-	"sort"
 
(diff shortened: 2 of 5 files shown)


edit/completion/sorter.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+		key = ""
+	}
+	value, _ := m.Index(key)
+	sorter, ok := value.(eval.Callable)
+	return sorter, ok
+}
+
+func sortRawCandidates(ev *eval.Evaler, customSorter eval.Callable, seed string, rs []rawCandidate) ([]rawCandidate, error) {
+	// default sorter
```
tw4452852: Actually, I prefer to remove this special case, but seems it is extremely fast compared to the following.

occivink:
I have concerns over this approach. 

The proposed directory directory sorting involves calling `test` n*log(n) times which is bound to be slow, especially with a lot of candidates. Even if there was a builtin, it's that many `stat()` calls that are esentially redundant with what the completer is doing.

In addition, the `git add` sorting would be tricky to implement, since the information of whether the file is tracked or not is lost after the completer.

SolitudeSF:
why cant implicit sorting be disabled and sorting routine just would be explicitly implemented in each completer? is there really a need to add another layer?

tw4452852:
@occivink  Thanks for your concerns. The proposed directory sorting just a example, i think stat() calls is necessary.  For `git add`, sorter's input is just the completer's result.

tw4452852:
@SolitudeSF , For now we just use the default sorter to sort  the candidates, anyway we could add a option to disable sorting entirely. What do you think @xiaq ?

And the sorting routine could be set per completer like the matcher.

xiaq:
Thank you, @tw4452852, but I share @occivink's concern.

@SolitudeSF, decoupling the sorter from the completer is useful. For instance, you can customize the ordering of filename completions without touching the filename completer.

The example you have given is an extreme case, but in general, if you can make do with calling an Elvish function O(1) times you should not call it O(n) times. You can usually rely on the pipeline for communication to reduce the number of calls; the matcher API is an example of how this principle is applied. Instead of calling the matcher for each candidate, it calls the matcher once and feeds the text of all candidates using the pipeline. Granted, to implement such a matcher in Elvish, it's likely that you still need to use `each` to iterate the input and thus end up calling some function O(n) times anyway. But that's an overhead we can optimize out later.

In our earlier discussion in the user group we have outlined two possible approaches to the sorter API:

* Feed the sorter function all candidates and use its output as the final candidate list. This approach has the downside that it makes the sorter overly powerful and can potentially be misused.

* Feed the sorter function all candidates (same as above) and expect it to output a numeric score for each candidate, and use that score for sorting. This API is much simpler, but it makes some simple things hard. For instance, how would you sort things simply by lexicographical order? To do this you need a string -> number mapping that preserves order. You can certainly do this by treating the string as a base-256 number (UTF-8 preserves ordering of codepoints), but that's crazy.

I have just thought of a variant of the second approach that can work, though, and it seems to be the best so far. The sorter still outputs a "score" for each candidate, but it doesn't have to be a numeric score. It can be anything, and the candidates are sorted according to the "score". For instance, if you want candidates to be sorted first by whether they are directories and then by the name, you can use this sorter:

```
{ each [name]{ put [?(test -d $name) $name] } }
```

There is one problem with this approach: Elvish doesn't support the notion of comparing, nor sorting anything other than strings and numbers. You can implement something specifically for the sorter of course. For instance, instead of relying on the default list type, you can implement a special `sort-key` type that has a well-defined ordering. Like a numerical score, it is used purely for scoring the candidates. Then the sorter above will look like

```
{ each [name]{ sort-key ?(test -d $name) $name } }
```

Another open question is how to distinguish numerical ordering and lexicographical ordering (Elvish doesn't have separate number types; #456), but that is also relatively easy if you have a dedicated `sort-key` builtin anyway. You can do something like `sort-key [number (get-modified-timestamp $name)] $name`, for example. (This is not a very good example because now you cannot use lists directly as arguments `sort-key`, but you get the idea. A good API for `sort-key` requires more thought.)

All these said, the completion pipeline still lacks something to make the sorter API more useful. In particular, you almost always want a specific sorter to only apply to a certain "type" of candidates. We do have a notion of "completion type" now, which includes `command`, `index`, `filename`, `redir`, and `argument`. The sorter API is already useful, for instance, if you can define a custom ordering for the `command` type. In fact this is what @notramo asked for in #525.

But the `argument` type is problematic; it should be finer-grained concept. For instance, the completion candidates for `git` can include flags, branch names, filenames, etc. And we want at least two things:

* In the arg-completer API, we should make it possible for the `git` completer to tag each candidate with types like `git-branch`, `git-flag`, `filename`, etc.

* You should be able to define sorters for different types.

This can be solved by making the "completion type" part of the candidate structure. Right now it is a property of the whole completion.

I hope I have made my vision clear. If you (still) want to tackle it, I recommend that you start with the last alternative of the sorter API I suggested and implement some version of `sort-key`. You can leave the support for finer-grained completion types to me.

krader1961:
@xiaq Ping? This is now almost two years old and seems unlikely to be merged. If it is still an interesting idea perhaps it should be converted into an enhancement issue. Should this PR be closed as rejected?

xiaq:
Thank you, @krader1961. I have filed #915 for the feature and am closing this PR.

PR #657, opened 2018-03-29
Add &numberify option to to-json builtin

VictorLowther:
I have a usecase that involves taking JSON from a remote source,
modifying one or more fields in it, and updating the remote source
with the new JSON fields.

Elvish internally translates all the numeric fields in the recieved
JSON into strings, which causes the JSON I need to send back to be
invalid and rejected by the remote service.  Until elvish gains
support for ints and floats as native internal data types, &numberify
is a hack to turn any strings that parse as valid ints or floats into
actual ints or floats.


eval/builtin_fn_io_test.go
```diff
@@ -19,17 +15,19 @@ func TestBuiltinFnIO(t *testing.T) {
 		That(`print "a\nb" | slurp`).Puts("a\nb"),
 		That(`print "a\nb" | from-lines`).Puts("a", "b"),
 		That(`print "a\nb\n" | from-lines`).Puts("a", "b"),
-		That(`echo '{"k": "v", "a": [1, 2]}' '"foo"' | from-json`).Puts(
```
xiaq: Did you remove this test case by mistake?

eval/builtin_ns.go
```diff
@@ -13,6 +14,7 @@ var builtinNs = Ns{
 	"ok":    vars.NewRo(OK),
 	"true":  vars.NewRo(true),
 	"false": vars.NewRo(false),
+	"nil":   vars.NewRo(vals.Nil{}),
```
xiaq: I don't want to introduce a nil value for Elvish. Call this `-json-nil`.

eval/builtin_fn_io.go
```diff
@@ -122,13 +125,48 @@ func toLines(fm *Frame, inputs Inputs) {
...
+		return res
+	default:
+		return src
+	}
+}
+
 // toJSON converts a stream of Value's to JSON data.
-func toJSON(fm *Frame, inputs Inputs) {
+func toJSON(fm *Frame, rawOpts RawOptions, inputs Inputs) {
+	opts := struct{ Numberify bool }{false}
```
xiaq: Call the option `&-numberify` by using a field tag `` `name="-numberify"` ``. Leading dashes mark things that will likely go away.

eval/vals/nil.go
```diff
@@ -0,0 +1,5 @@
+package vals
+
+// Represents the nil value, without actually being nil
+// which otherwise causes problems doing equality testing.
+type Nil struct{}
```
xiaq: Call this `JsonNil` and document that it is only used for JSON interop.

eval/vals/repr.go
```diff
@@ -32,6 +32,8 @@ type Reprer interface {
 // format "<unknown %v>".
 func Repr(v interface{}, indent int) string {
 	switch v := v.(type) {
+	case Nil:
```
xiaq: This is not a builtin type and thus doesn't need special-casing. Implement `Reprer` for `Nil` (or rather `JsonNil`, the name I suggested).

runtime/runtime.go
```diff
@@ -22,6 +22,7 @@ import (
 )
 
 var logger = util.GetLogger("[runtime] ")
+var ExtraNS = map[string]eval.Ns{}
```
xiaq: This is unrelated.

runtime/runtime.go
```diff
@@ -97,6 +98,9 @@ func InitRuntime(binpath, sockpath, dbpath string) (*eval.Evaler, string) {
 		ev.InstallModule("store", storemod.Ns(client))
 		ev.InstallModule("daemon", daemonmod.Ns(client, spawner))
 	}
+	for k, v := range ExtraNS {
```
xiaq: Ditto.

eval/builtin_fn_io.go
```diff
@@ -122,13 +125,48 @@ func toLines(fm *Frame, inputs Inputs) {
 	})
 }
 
+// Exported so that other packages can refer to it for proper nil
```
xiaq: * This doesn't need to be exported; it is not used in another other package.
* This function is only needed when "numberify" is enabled. Just name it `numberify` and avoid calling it when the option is not given.
* Instead of only handling marshaling of `Nil` here, implement [`Json.Marshaler`](https://godoc.org/encoding/json#Marshaler) for the `Nil` type so that it can be marshaled into JSON using the standard JSON encoding functions.

eval/builtin_fn_io_test.go
VictorLowther: No, the test case I added made it redundant.

eval/builtin_ns.go
VictorLowther: Can do.

eval/builtin_fn_io.go
VictorLowther: OK, will do.

eval/vals/repr.go
VictorLowther: Will do.

runtime/runtime.go
VictorLowther: You want me to pull these into a seperate PR?

eval/builtin_fn_io.go
VictorLowther: I call it from Go functions that are implemented in a namespace I inject using the ExtraNS stuff -- that namespace is responsible for implementing in-shell access to a JSON baseed API that does not deal well with numbers and nulls magically transmuting to strings.  If this is not a good place for it, is there a better location?

eval/builtin_fn_io_test.go
xiaq: I don't see what you mean; you haven't added a test case against the output of `from-json`.

eval/builtin_fn_io.go
xiaq: Elvish does not have a stable Go API. Helper functions like these are especially subject to change. If you want to use Elvish from another project, it is a better idea to duplicate those code in the consumer package. The other APIs you use are also subject to change, but less so.

runtime/runtime.go
xiaq: Please see my comment above about Go API; if you are building an alternative runtime, consider duplicating some code from the runtime package.

eval/builtin_fn_io_test.go
VictorLowther: Does the test case that tests that going from json, then back to the same json not cover it?

runtime/runtime.go
VictorLowther: My intent is not to build an alternative runtime, it is to make it easy to add new namespaces that are implemented in go. Doing that in a way that does not involve just forking elvish means I need to at least be able to have the runtime register new go foncs in their own namespace.

xiaq: Sorry if I wasn't clear.

Where do you intend to add new namespaces implemented in Go?

* Are you going to add them in the Elvish codebase? Then the current approach of calling `ev.InstallModule` works well.

* Are going to add them outside the Elvish codebase? Then that is what I meant by "an alternative runtime". Doing so means that you will be inevitably relying on Elvish's Go API. I am not ready to commit to a stable Go API, so I don't intend to support such use cases.

eval/builtin_fn_io_test.go
xiaq: No. That only tests that `from-json` and `to-json` are inverse operations (under certain conditions). If you both `from-json` and `to-json` are simply identity functions, the test you added would pass.

eval/builtin_ns.go
xiaq: Sorry for the late feedback, but now that this is only used for JSON, `json-null` is actually a more sensible choice...

xiaq:
Thanks for your interest and contribution, but this is no longer necessary as we have a design for explicitly number types in #816.

PR #674, opened 2018-05-16
Add `styled` and `styled-segment` builtins

fehnomenal:
These commits add two new builtins `styled` and `styled-segment` that are modelled after @xiaq's design in #520. It allows nesting, querying for style attributes and defining custom style transformers.

Currently the new functions are not used anywhere. The problem is that `ui.Styled` and `ui.Styles` are used to render the editor (completion matching, history, ...). This is done for the individual parts of the UI.
Probably there should be something like a renderer interface that is implemented by `shell.Shell` and `web.Web` and allows rendering of more abstract things like a selection list, notifications and so on.

This is probably a rather large refactoring that needs to be coordinated and unfortunately I'm very short on time at the moment. So I decided to publish at least my current progress.


fehnomenal:
With the last commit `$edit:styled~` now calls the new function. It seems to work but I haven't tested it intensively.

xiaq:
Hi! Thank you so much for your contribution. Really impressed by the progress you brought.

I will take a look later this week - a bit occupied lately.

xiaq:
Hi - I haven't been able to look at this last week. I will definitely (TM) take a look this week!

styled/colors.go
```diff
@@ -0,0 +1,107 @@
+package styled
+
+import "fmt"
+
+type Color uint
```
xiaq: Type safety and memory compactness aside, it seems to be much more straightforward if we just use `string` to represent colors and translate them to SGR codes when printing.

It will also be easier to extend more complex color representations (e.g. hex `#123456`) if we just use `string`.

styled/styled.go
```diff
@@ -0,0 +1,140 @@
+package styled
+
+import (
+	"fmt"
+
+	"github.com/elves/elvish/eval/vals"
+)
+
+type TextStyle struct {
+	bold       *bool
```
xiaq: Is `nil` (lack of an attribute) and `false` (explicit false) a useful distinction? I think not - just make the fields `bool` and the code can be simplified quite a bit.

eval/builtin_fn_styled.go
```diff
@@ -0,0 +1,155 @@
...
+
+func init() {
+	addBuiltinFns(map[string]interface{}{
+		"styled-segment": styledSegment,
+		"styled":         StyledBuiltin,
+	})
+
+	transformers := make(map[interface{}]interface{})
+	for k, v := range styled.SegmentTransformers {
+		transformers[k] = NewBuiltinFn("style-transform "+k, v)
```
xiaq: These builtins are not actually accessible from the `builtin:` namespace. Give them more funny names, like `<style transformer xxx>`.

Speaking of which, I thought a bit more about the API, and I think that exposing transformers as a writable map is likely a bad idea: this is essentially a global variable, and it is hard to coordinate access from different modules written by different people. For instance, one module may install a transformer named `foo` and use it later, while another may install a transformer *with the same name* `foo`. One of the two modules will use a transformer that it does not intend to use.

My amended proposal is to not expose this transformer map at all and keep it strictly internal. This part of the code can be removed.

In fact, after removing the transformer map, the builtin transformers do not even need to live in a map; the `styled` command can also parse the transformer programmatically.

fehnomenal:
Thank you for your comments. I will have a more in depth look at them this evening.
My comment about the lack of time was targeted at a greater refactoring (w.r.t. having a framework that handles rendering differently for the shell and the web frontend) but I want to finish these changes.

Thinking about this: Do we really need to distinguish these two parts?
I understand the web frontend is/will be more flexible to display hints, errors, completions, ... But in the end it would also display the output of external programs which expect to print to a shell and thus most probably use ANSI sequences themselves.
These sequences can only be displayed properly by the web frontend if the output is parsed and enriched with css classes.
This means we need this parsing anyway and the web frontend can parse our styles (that we are outputting, not external programs) as well.

styled/styled.go
xiaq: You are right. `no-` is probably a better prefix.

xiaq:
Re. distinguishing terminal and HTML/CSS outputs: you are right that for a web interface to work we will need to convert ANSI sequences to HTML/CSS anyway, so it makes sense to just implementation styled -> ANSI conversion and let ANSI -> HTML/CSS do the rest. But I see some disadvantages to this approach.

1) The range of possible styling supported by HTML/CSS is considerably larger than that of ANSI sequences. We will want `styled` to be able to represent them, losslessly. The conversion path of `styled` -> ANSI -> HTML/CSS will be lossy.

2) If you consider the problem of ANSI -> HTML/CSS conversion, you will first need to parse ANSI into some internal representation - and it makes perfect sense to used `styled` as that representation! Now that we have a `styled` -> HTML/CSS converter, all you need to do now is to parse ANSI into `styled`.

eval/builtin_fn_styled.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+
+			if n := len(vs); n != 1 {
+				return nil, fmt.Errorf("style transformers must return a single styled segment; got %d", n)
+			}
+
+			switch transformedSegment := vs[0].(type) {
+			case styled.Segment:
+				text[i] = transformedSegment
+			case *styled.Segment:
```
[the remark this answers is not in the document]
fehnomenal: This applied to `styledSegment` as well and was needed because sometimes they would get a pointer and sometimes not. The error was that I was passing the non-pointer segment to the user supplied function. It is gone in the refactored code.

fehnomenal:
OK, I rebased to the latest master and adressed all of your comments.
Available transformers are:
- `(bg-)?<color>`: Sets the foreground or background color (if `color` is valid according to `styled.IsValidColorString`)
- `(no-)?(bold|dim|italic|underlined|blink)`: Sets or unsets the respective attribute
- `inverse`: __Toggles__ the inverse attribute
- `force-(no-)?inverse`: Sets or unsets the inverse attribute

Probably there could be even more tests.

styled/styled.go
```diff
@@ -0,0 +1,101 @@
...
+
+func EmptyStyle() Style {
+	return Style{
+		Foreground: "default",
+		Background: "default",
+	}
+}
+
+func (s *Style) ImportFromOptions(options map[string]interface{}) error {
+	assignColor := func(key string, assign func(string)) error {
```
xiaq: The second argument can be a `*string`.

styled/styled.go
```diff
@@ -0,0 +1,101 @@
...
+		return err
+	}
+	if err := assignBool("inverse", func(b bool) { s.Inverse = b }); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func IsValidColorString(col string) bool {
```
xiaq: This does not need to be exported. It is also probably better called `isValidColorName`.

I am a little torn about how you are encoding a string set using a switch statement. It is indeed the most concise way though; let's keep it as is for now.

styled/styled.go
```diff
@@ -0,0 +1,101 @@
...
+		if c, ok := options[key]; ok {
+			if c, ok := c.(string); ok && IsValidColorString(c) {
+				assign(c)
+			} else {
+				return fmt.Errorf("value to option '%s' must be a valid color string", key)
+			}
+		}
+		return nil
+	}
+	assignBool := func(key string, assign func(bool)) error {
```
xiaq: Idem.

styled/segment.go
```diff
@@ -0,0 +1,104 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+
+	"github.com/elves/elvish/eval/vals"
+)
+
+// A styled Segment is a string that has some style applied to it.
```
xiaq: Docstrings for exported symbols should start with the symbol itself (https://golang.org/doc/effective_go.html#commentary). In this case, simply remove "A styled", which is apparent from the package it is in.

styled/segment.go
```diff
@@ -0,0 +1,104 @@
...
+type Segment struct {
+	Style
+	Text string
+}
+
+func (Segment) Kind() string { return "styled-segment" }
+
+// Returns the representation of this Segment. The string can be used to construct
```
xiaq: Start the docstring with `Repr`, so it reads "Repr returns ...". Same for docstrings below.

styled/styled.go
```diff
@@ -0,0 +1,101 @@
+package styled
+
+import (
+	"fmt"
+)
+
```
xiaq: Add docstring.

Generally, exported symbols should have docstrings, the only exception being docstrings are optional for methods implementing interface.

styled/styled.go
```diff
@@ -0,0 +1,101 @@
...
+	Foreground string
+	Background string
+	Bold       bool
+	Dim        bool
+	Italic     bool
+	Underlined bool
+	Blink      bool
+	Inverse    bool
+}
+
```
xiaq: This can be a variable. "`DefaultStyle`" is a better name; "empty" is normally understood to be `StructName{}` in Go.

Speaking of which, it probably makes sense to use the empty string for default colors; this makes `Style{}` a valid value, which is considered good practice in Go. But I am not requiring you to do this in this PR.

styled/segment.go
fehnomenal: Alright I will rework and add missing docstrings.

styled/styled.go
fehnomenal: I also thought about defining empty strings to be the default color but wasn't sure how to handle the style transformer names. They should still be called `default` and `bg-default ` but I had the thought that it could be inconsistent...

Should the representation of segments contain the default color?

styled/styled.go
fehnomenal: I knew there has to be a non awkward way to do this. Thx!

styled/styled.go
fehnomenal: You're absolutely right on both points...

fehnomenal:
Again I will work on this this evening after work. While doing this I could also add more (and better named) transformers. 

I will make them uniform so that for each attribute there will be setting, unsetting and toggling.

fehnomenal:
Done.

zzamboni:
@xiaq @fehnomenal this is an awesome change! I look forward to exploring all the new capabilities.

If I may suggest, removing `edit:styled` altogether is a bit sudden in my opinion. I normally don't mind backward-incompatible changes, but would it be possible to add `edit:styled` as an alias of the new `styled` for some time? Just to give us a chance to update all those modules which use it... :)

PR #689, opened 2018-06-01
Suppress Background Notification

### LINKED ISSUE #683: Suppress "finished" message from a backgrounded job
Is there any way to suppress the `job ... finished` message after a backgrounded job completes?

iwoloschin:
Fixes #683 

Allow the suppression of the `job ... finished` message after a background process finishes.  Message suppression is controlled by the new builtin variable `suppress-background`.  A value of "true" will suppress any background notification messages.  Errors are always printed regardless of how `suppress-background` is set.

### DIFF
--- a/eval/compile_op.go
+++ b/eval/compile_op.go
@@ -141,7 +141,7 @@ func (op *pipelineOp) Invoke(fm *Frame) error {
 			if err != nil {
 				msg += ", errors = " + err.Error()
 			}
-			if suppressBackground.Get() != "true" {
+			if suppressBackground.Get() != "true" || err != nil {
 				if fm.Editor != nil {
 					fm.Editor.Notify("%s", msg)
 				} else {
--- a/eval/compile_op.go
+++ b/eval/compile_op.go
@@ -137,11 +137,10 @@ func (op *pipelineOp) Invoke(fm *Frame) error {
 			wg.Wait()
 			msg := "job " + op.source + " finished"
 			err := ComposeExceptionsFromPipeline(errors)
-			suppressBackground := vars.FromPtr(fm.Evaler.suppressBackground)
 			if err != nil {
 				msg += ", errors = " + err.Error()
 			}
-			if suppressBackground.Get() != true || err != nil {
+			if fm.Evaler.notifyBgJobSuccess != true || err != nil {
 				if fm.Editor != nil {
 					fm.Editor.Notify("%s", msg)
 				} else {
--- a/eval/compile_op.go
+++ b/eval/compile_op.go
@@ -140,7 +140,7 @@ func (op *pipelineOp) Invoke(fm *Frame) error {
 			if err != nil {
 				msg += ", errors = " + err.Error()
 			}
-			if fm.Evaler.notifyBgJobSuccess != true || err != nil {
+			if fm.Evaler.notifyBgJobSuccess || err != nil {
 				if fm.Editor != nil {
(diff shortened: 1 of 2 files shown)


fehnomenal:
Why don't you use `bool` for the variable? In elvish it would be set with `$true` or `$false`.

eval/compile_op.go
```diff
@@ -137,13 +137,16 @@ func (op *pipelineOp) Invoke(fm *Frame) error {
...
 			err := ComposeExceptionsFromPipeline(errors)
+			suppressBackground := vars.FromPtr(fm.Evaler.suppressBackground)
 			if err != nil {
 				msg += ", errors = " + err.Error()
 			}
-			if fm.Editor != nil {
-				fm.Editor.Notify("%s", msg)
-			} else {
-				fm.ports[2].File.WriteString(msg + "\n")
+			if suppressBackground.Get() != "true" || err != nil {
```
xiaq: There is no need to construct a `Var` just to access a struct field. Just use `fm.Evaler.suppressBackground`.

eval/eval.go
```diff
@@ -36,13 +36,15 @@ const (
...
 	initIndent         = vals.NoPretty
+	defaultSuppressBackground = "false"
 )
 
 // Evaler is used to evaluate elvish sources. It maintains runtime context
 // shared among all evalCtx instances.
 type Evaler struct {
 	evalerScopes
 	valuePrefix  string
+	suppressBackground string
```
xiaq: Use a `bool`.

Prefer using positive-meaning names for options. In this case, call the field `notifyBgJobSuccess`, give it a default value of `true`, and call the Elvish variable `$notify-bg-job-success`

eval/eval.go
```diff
@@ -36,13 +36,15 @@ const (
 const (
 	defaultValuePrefix = "▶ "
 	initIndent         = vals.NoPretty
+	defaultSuppressBackground = "false"
```
xiaq: This constant belongs together with `defaultValuePrefix`, so move it up one line.

iwoloschin:
Ha, didn't even realize it could be a bool, was thinking in terms of environmental variables.  Other changes fixed too.

eval/compile_op.go
```diff
@@ -140,10 +140,12 @@ func (op *pipelineOp) Invoke(fm *Frame) error {
 			if err != nil {
 				msg += ", errors = " + err.Error()
 			}
-			if fm.Editor != nil {
-				fm.Editor.Notify("%s", msg)
-			} else {
-				fm.ports[2].File.WriteString(msg + "\n")
+			if fm.Evaler.notifyBgJobSuccess != true || err != nil {
```
xiaq: Since we reverted the meaning of the flag, it should simply be

`if fm.Evaler.notifyBgJobSuccess || err != nil`

(note that even if you want `!= true`, you would always write `!flag` instead of `flag != true`)

eval/eval.go
```diff
@@ -34,19 +34,21 @@ const (
 )
 
 const (
-	defaultValuePrefix = "▶ "
-	initIndent         = vals.NoPretty
+	defaultValuePrefix        = "▶ "
+	defaultNotifyBgJobSuccess = false
```
xiaq: Default to `true` :)

iwoloschin: I should really not try and code with a toddler and infant awake...

PR #691, opened 2018-06-02
Make complex-candidate indexable

### LINKED ISSUE #684: Allow indexing of complex-candidate objects
It would be nice to be able to extract components from complex-candidate objects.
```
[~]─> a = (edit:complex-candidate something &code-suffix='' &display-suffix='' &style=red)
[~]─> put $a
▶ (edit:complex-candidate something &code-suffix='' &display-suffix='' &style=red)
[~]─> keys $a
Exception: map cannot have its keys iterated
[tty], line 1: keys $a
[~]─> put $a[code-suffix]
Exception: not indexable
[tty], line 1: put $a[code-suffix]
```

zzamboni:
Makes it possible to access the fields of a complex-candidate object. Fixes #684.

### DIFF
--- a/edit/completion/candidate.go
+++ b/edit/completion/candidate.go
@@ -67,6 +67,24 @@ type complexCandidate struct {
 	style         ui.Styles // Used in the menu.
 }
 
+func (c *complexCandidate) Index(k interface{}) (interface{}, bool) {
+	switch k {
+	case "stem":
+		return c.stem, true
+	case "code-suffix":
+		return c.codeSuffix, true
+	case "display-suffix":
+		return c.displaySuffix, true
+	case "style":
+		return c.style.String(), true
+	}
+	return nil, false
+}
+
+func (c *complexCandidate) IterateKeys(f func(interface{}) bool) {
+	vals.Feed(f, "stem", "code-suffix", "display-suffix", "style")
+}
+
 func (c *complexCandidate) Kind() string { return "map" }
 
 func (c *complexCandidate) Equal(a interface{}) bool {


xiaq:
Awesome, thank you!

Can you add some tests as well? This can be a little bit involved, so here is some more guidance :)

In general with Go, you write test cases by writing functions named `Test*` in a file `*_test.go`, and those functions should accept one argument of type `*testing.T`. That is, the functions look like `func TestSomething(t *testing.T) { ... }` (the argument name `t` is convention).

To test the code you added does indeed implement indexing for complex candidates, you want to run some Elvish code and see if they behave as expected. For instance, you want to test that `put (edit:complex-candidate lorem)[stem]` puts the string `lorem`. Elvish has an internal mini-framework for testing Elvish code in `eval/testutils.go`: https://github.com/elves/elvish/blob/1ab103aceed10eca03d7a46359fd09edd1014f41/eval/testutils.go#L1-L14

Since the `edit` module is not built-in, this will require some setup. One such example is https://github.com/elves/elvish/blob/1ab103aceed10eca03d7a46359fd09edd1014f41/eval/re/re_test.go#L10-L13

Where the setup is simply adding an additional `re` namespace to the Evaler. In this case, it is easier to construct a namespace on the fly and add `complex-candidate` to it, so you would add something like this to `candidate_test.go`:

```go
func TestComplexCandidate(t *testing.T) {
	setup := func(ev *eval.Evaler) {
		ev.Builtin.AddNs("edit", eval.NewNs().AddBuiltinFn(
			"edit:", "complex-candidate", makeComplexCandidate))
	}
	That := eval.That 
	eval.TestWithSetup(t, setup,
		That("put (edit:complex-candidate lorem)[stem]").Puts("lorem"),
		// Add more similar test cases here
	)
}
```

Good luck!

zzamboni:
@xiaq I've added some test cases - thanks for the pointers!

PR #692, opened 2018-06-03
Add variable to track number of background jobs

iwoloschin:
Pretty much what it says in the subject, added in a variable to track the number of background jobs.  The most obvious use for something like this is to create a prompt notification about background jobs but it could potentially have other uses as well, such as warning the user if they try to exit the shell with jobs still running in the background.


xiaq:
Ideally this should have some tests, but right now it's kind of hard to test.

PR #696, opened 2018-06-11
Added 'edit:move-dot-left-small-word' and 'edit:move-dot-right-small-word'

muesli:
Following the examples of 'kill-small-word-left' and the 'move-dot-*-word'
functions, I added two new functions to move the cursor left or right,
stopping at the next "small word".

This would typically be used for Ctrl+Left & Right, e.g.:

edit:insert:binding[Ctrl+Left]=$edit:move-dot-left-small-word~
edit:insert:binding[Ctrl+Right]=$edit:move-dot-right-small-word~

### DIFF
--- a/edit/edcore/insert.go
+++ b/edit/edcore/insert.go
@@ -30,14 +30,16 @@ func initCoreFns(ed *editor, ns eval.Ns) {
 		"kill-rune-left":       ed.killRuneLeft,
 		"kill-rune-right":      ed.killRuneRight,
 
-		"move-dot-left":       ed.moveDotLeft,
-		"move-dot-right":      ed.moveDotRight,
-		"move-dot-left-word":  ed.moveDotLeftWord,
-		"move-dot-right-word": ed.moveDotRightWord,
-		"move-dot-sol":        ed.moveDotSOL,
-		"move-dot-eol":        ed.moveDotEOL,
-		"move-dot-up":         ed.moveDotUp,
-		"move-dot-down":       ed.moveDotDown,
+		"move-dot-left":             ed.moveDotLeft,
+		"move-dot-right":            ed.moveDotRight,
+		"move-dot-left-word":        ed.moveDotLeftWord,
+		"move-dot-right-word":       ed.moveDotRightWord,
+		"move-dot-left-small-word":  ed.moveDotLeftSmallWord,
+		"move-dot-right-small-word": ed.moveDotRightSmallWord,
+		"move-dot-sol":              ed.moveDotSOL,
+		"move-dot-eol":              ed.moveDotEOL,
+		"move-dot-up":               ed.moveDotUp,
+		"move-dot-down":             ed.moveDotDown,
 
 		"insert-last-word": ed.insertLastWord,
 		"insert-key":       ed.insertKey,
@@ -210,26 +212,26 @@ func (ed *editor) moveDotRight() {
 	ed.dot += w
 }
 
-func (ed *editor) moveDotLeftWord() {
+func (ed *editor) moveDotLeftSepFunc(fn func(rune) bool) {
 	if ed.dot == 0 {
 		return
 	}
 	space := strings.LastIndexFunc(
-		strings.TrimRightFunc(ed.buffer[:ed.dot], unicode.IsSpace),
-		unicode.IsSpace) + 1
+		strings.TrimRightFunc(ed.buffer[:ed.dot], fn),
+		fn) + 1
 	ed.dot = space
 }
 
-func (ed *editor) moveDotRightWord() {
-	// Move to first space
-	p := strings.IndexFunc(ed.buffer[ed.dot:], unicode.IsSpace)
+func (ed *editor) moveDotRightSepFunc(fn func(rune) bool) {
+	// Move to first matching rune
+	p := strings.IndexFunc(ed.buffer[ed.dot:], fn)
 	if p == -1 {
 		ed.dot = len(ed.buffer)
 		return
 	}
 	ed.dot += p
-	// Move to first nonspace
-	p = strings.IndexFunc(ed.buffer[ed.dot:], notSpace)
+	// Move to first non-matching rune
+	p = strings.IndexFunc(ed.buffer[ed.dot:], func(r rune) bool { return !fn(r) })
 	if p == -1 {
 		ed.dot = len(ed.buffer)
 		return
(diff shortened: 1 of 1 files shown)


edit/edcore/insert.go
```diff
@@ -210,35 +212,47 @@ func (ed *editor) moveDotRight() {
 	ed.dot += w
 }
 
-func (ed *editor) moveDotLeftWord() {
+func (ed *editor) moveDotLeftIndexFunction(fn func(rune) bool) {
```
xiaq: `moveDotLeftSepFunc` would be a better name.

edit/edcore/insert.go
```diff
@@ -210,35 +212,47 @@ func (ed *editor) moveDotRight() {
...
 	space := strings.LastIndexFunc(
-		strings.TrimRightFunc(ed.buffer[:ed.dot], unicode.IsSpace),
-		unicode.IsSpace) + 1
+		strings.TrimRightFunc(ed.buffer[:ed.dot], fn),
+		fn) + 1
 	ed.dot = space
 }
 
-func (ed *editor) moveDotRightWord() {
+func (ed *editor) moveDotRightIndexFunction(fn func(rune) bool) {
```
xiaq: Ditto; `moveDotRightSepFunc`.

edit/edcore/insert.go
```diff
@@ -210,35 +212,47 @@ func (ed *editor) moveDotRight() {
...
+func (ed *editor) moveDotLeftWord() {
+	ed.moveDotLeftIndexFunction(unicode.IsSpace)
+}
+
+func (ed *editor) moveDotRightWord() {
+	ed.moveDotRightIndexFunction(unicode.IsSpace)
+}
+
+func (ed *editor) moveDotLeftSmallWord() {
+	ed.moveDotLeftIndexFunction(isAlnum)
```
xiaq: There is something curious here; the internal functions are given a function that returns whether a rune is a separator above, but here it is the opposite: they are given a function that returns whether a rune is **not** a separator.

Both would in fact work, but for clarify, please define a function `isNotAlnum` and use that instead.

xiaq:
Friendly ping. There is just one change required now, namely inverting the predicate used in the small word movement functions.

zzamboni:
@muesli Another friendly ping - any chance to make the last changes @xiaq asked for? I'd love to have this merged ;)

muesli:
Yeah, still interested in this, but I must admit that I'm a bit lost as to what behavior is desired now. Might have to go through the IRC logs again.

edit/edcore/insert.go
```diff
@@ -210,35 +212,47 @@ func (ed *editor) moveDotRight() {
...
+func (ed *editor) moveDotLeftWord() {
+	ed.moveDotLeftSepFunc(unicode.IsSpace)
+}
+
+func (ed *editor) moveDotRightWord() {
+	ed.moveDotRightSepFunc(unicode.IsSpace)
+}
+
+func (ed *editor) moveDotLeftSmallWord() {
+	ed.moveDotLeftSepFunc(isAlnum)
```
xiaq: Instead of passing `isAlnum`, pass the negation of `isAlnum`, i.e.

```go
ed.moveDotLeftSepFunc(func(r rune) { return !isAlnum(r) })
```

Ditto below

kwshi:
I'm inclined to argue that using sep-functions (as in `moveDotLeftSepFunc(sepFunc func(rune) bool)`) for small-word movement does not fit our definition of *small words*.  

The sep-function approach assumes that each character either *is in a word* or *is not in a word*.

In the case of "big" (i.e. not small) words, this works, because any whitespace character *is not* in a word and used as a word *sep*arator, and any non-whitespace character *is* in a word.  Thus any sequence of characters can be converted to a string of `a` (not-word) and `_` (word) characters and navigated as such (with word boundaries marked below as `^` (start) / `$` (end)):

```
sudo systemctl list-unit-files; cd /etc/system/system.d/
aaaa_aaaaaaaaa_aaaaaaaaaaaaaaaa aa aaaaaaaaaaaaaaaaaaaaa
^  $ ^       $ ^              $ ^$ ^                   $
```

In the case of *small* words, the distinction is not a simple true/false partition: alphanumeric and non-alphanumeric characters are both word-characters, and delimiting adjacent words becomes not a matter of word/non-word but rather word-alphanumeric/word-nonalphanumeric/non-word (mapped to `a`/`*`/`_`) (word boundaries marked `^` (start), `$` (end), `.` (single-character-word)):

```
sudo systemctl list-unit-files; cd /etc/system/system.d/
aaaa_aaaaaaaaa_aaaa*aaaa*aaaaa*_aa_*aaa*aaaaaa*aaaaaa*a*
^  $ ^       $ ^  $.^  $.^   $. ^$ .^ $.^    $.^    $...
```

If we applied a simple sep-func approach to partitioning by small-words, specifically implemented here as alphanumeric => word (`a`), non-alphanumeric => not-word (`_`), we'd end up instead with

```
sudo systemctl list-unit-files; cd /etc/system/system.d/
aaaa_aaaaaaaaa_aaaa_aaaa_aaaaa__aa__aaa_aaaaaa_aaaaaa_a_

^  $ ^       $ ^  $.^  $.^   $. ^$ .^ $.^    $.^    $... (desired word boundaries)
^  $ ^       $ ^  $ ^  $ ^   $  ^$  ^ $ ^    $ ^    $ .  (actual word boundaries)
```

You can see that this is effectively not partitioning by *small-words*, but rather by *alphanumeric words*.  In this scheme, there is no distinction between non-alphanumeric characters and whitespace-haracters:

```
mkdir -- --file (original string)
aaaaa_**_**aaaa (desired small-word interpretation)
aaaaa______aaaa (interpretation under this scheme)
^   $ ^$ ^$^  $ (desired word boundaries)
^   $      ^  $ (actual word boundaries)
```

This is inconsistent with the notion of small-words used in `kill-left-small-word`, and also with other editors' (namely, vim's) notion of small-words.  I'm not against incorporating this feature, but I think we'd be better off naming it `moveDotLeftAlphanumericWord`, which more accurately fits its actual behavior.  

For small-word navigation behavior consistent with the current definitions, I propose my implementation in #721 (which also fixes some other issues).

xiaq:
@kwshi Your comment about n-category is spot-on. I don't think it is any over-abstraction, it is a good model of how people expect things to work. I was thinking at one point about exposing a generic word-movement function that would simply take a list of categories. For instance, (big) word movement can work like:

```
edit:move-dot-left-generic-word ["\n\t "]
```

The set of characters not in any category form an implicit category. In this case, only the whitespace category is explicitly given and the rest form an implicit (big) word category.

Small word movement is slightly more complex as the alphanumeric set is pretty large if you consider Unicode. The API needs to accept predicates in addition to just strings:

```
edit:move-dot-left-generic-word ["\n\t " [r]{ or (unicode:is-letter $r) (unicode:is-number $r) }]
```

(This also depends on a nonexistent `unicode` module...)

xiaq:
I am going to accept the PR as is. It is not perfect, but it is an oversight on the design side (which I take responsibility) instead of implementation (which @muesli deserves the credit for).

@kwshi Let's iterate this in your #721.

xiaq:
Hmm, I overlooked again the fact that whitespaces are not symmetrical to word characters. Whitespaces should always be skipped, and word movements should always stop at the beginning of a run of word characters.

The return value of the categorizer needs to have a meaning: when it is 0, it signals "things that should be just skipped". When it is > 0, the actual value doesn't matter.

PR #707, opened 2018-06-27
Add support for -norc flag

iwoloschin:
Basically what it says in the title, adding support for a `-norc` flag to the Elvish shell.  Intended to be used for debugging/troubleshooting, or really anything that requires testing against a "stock" instance of Elvish without any customization that would normally be in a user's `rc.elv` file.

### DIFF
--- a/program/program.go
+++ b/program/program.go
@@ -34,7 +34,7 @@ type flagSet struct {
 
 	Help, Version, BuildInfo, JSON bool
 
-	CodeInArg, CompileOnly, norc bool
+	CodeInArg, CompileOnly, NoRc bool
 
 	Web  bool
 	Port int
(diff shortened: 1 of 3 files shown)


program/program.go
```diff
@@ -34,7 +34,7 @@ type flagSet struct {
 
 	Help, Version, BuildInfo, JSON bool
 
-	CodeInArg, CompileOnly bool
+	CodeInArg, CompileOnly, norc bool
```
xiaq: Call it `NoRc` for consistency with other fields.

program/shell/interact.go
```diff
@@ -29,7 +29,7 @@ func interact(ev *eval.Evaler, dataDir string) {
 	defer ed.Close()
 
 	// Source rc.elv.
-	if dataDir != "" {
+	if norc == false && dataDir != "" {
```
xiaq: `!norc`

program/shell/shell.go
```diff
@@ -21,14 +21,15 @@ type Shell struct {
 	DbPath      string
 	Cmd         bool
 	CompileOnly bool
+	norc        bool
```
xiaq: Call it `NoRc`.

Issue #702, opened 2018-06-24
readline-bindings alt+backspace, alt+d delete word


kwshi:
It would be nice to have Alt+backspace (for deleting a word to the left) and Alt+d (for deleting a word to the right) be included in the readline keybindings.  I use them regularly in other shells and think it would be a fitting addition.

kwshi:
Resolved by #721.

PR #721, opened 2018-07-16
edit/edcore: Implement functions for word-based movement.

kwshi:
Implemented `kill-word-right`, `kill-small-word-right`, `move-dot-left-small-word`, and `move-dot-right-small-word`.

**Oops:** I had overlooked the fact that @muesli had already implemented the last two functions (`move-dot-left-small-word` and `move-dot-right-small-word`); I originally set out to implement the `kill-` functions, saw that those two happened to be missing from the source, and accidentally did those too.  

*Note*, however, that these two implementations behave differently.  @muesli's implementation seems to always move to the next non-alphanumeric character, as though it only considers alphanumeric runs as valid small-words.  This doesn't exactly match the current definition (as described by @xiaq in the comment before `kill-small-word-left`), which also considers non-alphanumeric runs as valid (but separate) small-words.  This implementation tries to follow the current definition exactly.


kwshi:
By the way, this is part of an attempt to address #702, since it introduces the functions that perform the kill feature that #702 demands.  If this is accepted, then all we need to do is simply attach the keybindings in `eval/bundled/binding.elv.go`.

xiaq:
Maybe rebase the work when #696 gets merged and apply the idea you proposed in #722?

kwshi:
Resolves #722.

kwshi:
Resolves #702.

edit/edcore/insert.go
```diff
@@ -131,30 +136,78 @@ func (ed *editor) commandStart() {
...
 }
 
-func (ed *editor) killLineRight() {
-	eol := util.FindFirstEOL(ed.buffer[ed.dot:]) + ed.dot
-	ed.buffer = ed.buffer[:ed.dot] + ed.buffer[eol:]
+func (ed *editor) applyKill(move func(string, int) int) func() {
+	return func() {
+		index := move(ed.buffer, ed.dot)
+		
+    if index < ed.dot { // delete left
```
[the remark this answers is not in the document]
kwshi: Strange. I'm using the automatic indentation with emacs's go-mode.  I'll be sure to run gofmt next time...

xiaq:
Ah, I just saw that you have pushed more changes. I will review later today.

GitHub doesn't send out emails on pushes to PR, you would need to remind me to do another review (or if you don't mind, you can also rely on my occasional browsing of PRs ;-) )

edit/edcore/insert.go
```diff
@@ -133,136 +136,160 @@ func (ed *editor) commandStart() {
...
+// function, which returns for each character an integer representing
+// the category of the integer.  The return value `0` of the
+// categorizer function is used as a special category representing
+// non-word, or "always-skip" characters.  All other return values are
+// equivalent, used only to distinguish between different character
+// classes.
+//
+// See `categorizeWord`, `categorizeSmallWord`, and
+// `categorizeAlphanumeric` for examples.
+func moveDotLeftCategoryFunc(categorize func(rune) int) func(string, int) int {
```
xiaq: Can we make this a low-order function `moveDotLeftCategoryFunc(categorize func(rune) int, buffer string, dot int) int`? This will make the definition of the `moveDotLeft*` functions slightly more complex (did you know that `gofmt` actually doesn't mind if you define a series functions as one-liners without empty lines between them?), but it is slightly clearer IMO. Ditto for the `moveDotRightCategoryFunc`.

edit/edcore/insert.go
```diff
@@ -133,136 +136,160 @@ func (ed *editor) commandStart() {
...
+		return 1
+	default:
+		return 0
+	}
 }
 
-func (ed *editor) moveDotRightSmallWord() {
-	ed.moveDotRightSepFunc(func(r rune) bool { return !isAlnum(r) })
-}
+var (
```
xiaq: Can we have some unit tests for these functions? :)

kwshi:
@xiaq, I have made the changes you suggested in your review.  Let me know what you think.  Cheers!

:yum:

xiaq:
Wonderful!

PR #769, opened 2018-10-20
Add NixOS/nix installation to download section

SitiSchu:
Signed-off-by: SitiSchu <admin@sitischu.com>


SitiSchu:
While I'm at it do you want me to make the section a bit more consistent?
For example, some parts use a comment to say its the latest version(macos and now nixos) and some have it in the message before (OpenBSD).

PR #771, opened 2018-10-28
Bump Go versions and use '.x' to always get latest patch versions

HaraldNordgren:


HaraldNordgren:
@xiaq Yup. It's an unfortunate consequence of the Go's release versioning.

1.10 is the only "1.10.0" version, e.g. `eval "$(gimme "1.10.0")"` doesn't works since that version does not exist.

PR #792, opened 2019-01-28
sys: use unix.IoctlGetWinsize and unix.Winsize

tklauser:
Use the IoctlGetWinsize function and Winsize type from
golang.org/x/sys/unix in sys.GetWinsize instead of duplicating it.


xiaq:
Thank you!

PR #794, opened 2019-02-01
support option argument completion

tw4452852:
Signed-off-by: Tw <wei.tan@intel.com>

### DIFF
--- a/edit/completion/compl_getopt.go
+++ b/edit/completion/compl_getopt.go
@@ -20,6 +20,7 @@ func complGetopt(fm *eval.Frame, elemsv, optsv, argsv interface{}) {
 	)
 	desc := make(map[*getopt.Option]string)
 	argdesc := make(map[*getopt.Option]string)
+	optCompleters := make(map[*getopt.Option]eval.Callable)
 	// Convert arguments.
 	err := vals.Iterate(elemsv, func(v interface{}) bool {
 		elem, ok := v.(string)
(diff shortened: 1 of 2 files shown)


edit/completion/compl_getopt.go
```diff
@@ -91,6 +92,22 @@ func complGetopt(fm *eval.Frame, elemsv, optsv, argsv interface{}) {
 		if s, ok := getStringField("arg-desc"); ok {
 			argdesc[opt] = s
 		}
+
+		getCallableField := func(k string) (eval.Callable, bool) {
```
xiaq: Can you move the definition to just under `getStringField`?

edit/completion/compl_getopt.go
```diff
@@ -91,6 +92,22 @@ func complGetopt(fm *eval.Frame, elemsv, optsv, argsv interface{}) {
...
+			if !ok {
+				return nil, false
+			}
+			if vb, ok := v.(eval.Callable); ok {
+				return vb, true
+			}
+			throwf("%s should be fn, got %s", k, vals.Kind(v))
+			panic("unreachable")
+		}
+		if f, ok := getCallableField("handler"); ok {
```
xiaq: Can we call the field `completer` instead of `handler`?

edit/completion/compl_getopt.go
```diff
@@ -141,6 +158,18 @@ func complGetopt(fm *eval.Frame, elemsv, optsv, argsv interface{}) {
 		out <- c
 	}
 
+	callHandler := func(fn eval.Callable, args []string) {
```
xiaq: Can we call the function `callCompleter`?

edit/completion/compl_getopt.go
tw4452852: I also put all three functions (`getCallableField`,`getCallableField` and `getBoolField`) together, hope to make it more clear.

tw4452852:
@xiaq Of course, I have updated reference manual.

PR #801, opened 2019-03-14
Convert to Go modules

kenshaw:
Converts Gopkg.{toml,lock} to go.{mod,sum} and makes minor changes to
the travis-ci configuration.


.travis.yml
```diff
@@ -4,9 +4,9 @@ os:
   - osx
 sudo: false
 go:
+  - "1.12.x"
```
xiaq: Please change the Travis configuration in another followup PR.

.travis.yml
```diff
@@ -4,9 +4,9 @@ os:
...
   - "1.11.x"
-  - "1.10.x"
-env: CGO_ENABLED=0
+env: CGO_ENABLED=0 GO111MODULE=on
```
xiaq: Before `GO111MODULE` defaults to on, we should support both GOPATH-based and module-based workflows. So we will need to test two cases, one where GO111MODULE is not set (and because of the setup in our travis script, the GOPATH workflow will be used) and where it is set to be `on`.

kenshaw:
Apologies on the delay on submitting the revisions. I added the additional `env` matrix in the travis config, and made some other minor edits to the travis config. All of the env matrix variants for Go 1.11 and 1.12 pass. Hoping this can get rebased onto master, so that I can continue with the other dep -> mod conversions I'm undertaking. Thanks!

.travis.yml
kenshaw: I've changed it here, as this won't work properly with Go 1.10. It's an "all or nothing" with regards to converting over to modules.

kenshaw:
Also -- a quick note: it's not really possible to change the `env` matrix for Go 1.10, hence this PR necessitates also changing the travis config, as I don't see how they can be separate PRs.

xiaq:
I wish that you could split the change of the Travis config into another PR, but I don't feel too strongly about it. It's not worth yet another round trip over this detail.

I am going to merge, thanks for the contribution!

PR #802, opened 2019-03-16
eval: move restore functionality into assignmentOp

### LINKED ISSUE #532: Temporary assignment to nonexistent variable leaves it equal to ''
```
~/go/src/github.com/elves/elvish> put $y
Compilation error: variable $y not found
  [interactive], line 1:
    put $y
~/go/src/github.com/elves/elvish> y=1 echo hi
hi
~/go/src/github.com/elves/elvish> put $y
▶ ''
~/go/src/github.com/elves/elvish>       
```

Is this a known issue? I think `y` should be undefined as before.

xofyarg:
This change makes assignmentOp mutable only on temporary assignment,
otherwise the outer Op needs to know how to deal with it.
Theoretically, lvaluesOp should tell which variable is new, but it
requires more changes to the underlying structures.

This fixes #532.

### DIFF
--- a/eval/compile_effect.go
+++ b/eval/compile_effect.go
@@ -141,20 +141,17 @@ func (op *pipelineOp) invoke(fm *Frame) error {
 }
 
 func (cp *compiler) form(n *parse.Form) effectOpBody {
-	var saveVarsOps []lvaluesOp
 	var assignmentOps []effectOp
+
 	if len(n.Assignments) > 0 {
-		assignmentOps = cp.assignmentOps(n.Assignments)
 		if n.Head == nil && n.Vars == nil {
 			// Permanent assignment.
-			return seqOp{assignmentOps}
+			return seqOp{cp.assignmentOps(n.Assignments, false)}
 		}
-		for _, a := range n.Assignments {
-			v, r := cp.lvaluesOp(a.Left)
-			saveVarsOps = append(saveVarsOps, v, r)
-		}
-		logger.Println("temporary assignment of", len(n.Assignments), "pairs")
 	}
+
+	assignmentOps = cp.assignmentOps(n.Assignments, true)
+	logger.Println("temporary assignment of", len(n.Assignments), "pairs")
 
 	// Depending on the type of the form, exactly one of the three below will be
 	// set.
@@ -213,7 +210,7 @@ func (cp *compiler) form(n *parse.Form) effectOpBody {
 			argsOp.end = argOps[len(argOps)-1].end
 		}
 		spaceyAssignOp = effectOp{
-			&assignmentOp{varsOp, restOp, argsOp},
+			&assignmentOp{varsOp, restOp, argsOp, false, false, nil, nil},
 			n.Range().From, argsOp.end,
 		}
 	}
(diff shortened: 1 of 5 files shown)


eval/compile_effect.go
```diff
@@ -358,23 +328,35 @@ func allTrue(vs []interface{}) bool {
...
 	restOp      lvaluesOp
 	valuesOp    valuesOp
+	temporary   bool
+	assigned    bool
+	origVars    []vars.Var
+	origVals    []interface{}
 }
 
 func (op *assignmentOp) invoke(fm *Frame) (errRet error) {
+	if op.temporary && op.assigned {
```
xiaq: This feels like fitting a square peg into a round whole: since `assignmentOp` can now do both assignment and restoration, `effectOpBody` is not an expressive enough interface for it. For instance, you can define a `revertableEffectOpBody` which can look like:

```go
type revertableEffectOpBody interface {
    invoke(fm *Frame) (revert effectOp, err error)
}
```

Or maybe you want `assignmentOp` to continue satisfying the `effectOpBody` interface (which is useful when doing permanent assignments as you don't need to revert), and let it satisfy another interface in addition; in that case you would give the method another name:

```go
type revertableEffectOpBody interface {
    invokeRevertable(fm *Frame) (revert effectOp, err error)
}
```

I'm not sure which option will work the best; you might want to play with those options and maybe be creative.

eval/compile_effect.go
```diff
@@ -358,23 +328,35 @@ func allTrue(vs []interface{}) bool {
 	return true
 }
 
-func (cp *compiler) assignment(n *parse.Assignment) effectOpBody {
+func (cp *compiler) assignment(n *parse.Assignment, temporary bool) effectOpBody {
```
xiaq: Overall, what you are doing here is not quite enough; because Elvish checks for variable existence statically, code like `x=foo echo $x; echo $x` should be a **compilation error**. This means that deleting the variable from the runtime Ns is not sufficient, it also has to be deleted from the compiler's scope after compiling the first form `x=foo echo $x`, so that the second form `echo $x` does not compile at all.

Essentially what you have to do is to revert the side effect `registerVariableSet` call [here](https://github.com/elves/elvish/blob/71d050b31c1578b763cf9d6cdae50ca6fcc2956a/eval/compile_lvalue.go#L81), for variables that did not exist before the temporary assignment. Doing so is non-trivial (which is why I did not implement it in the first place)

xofyarg: Ah, I see what you mean. I felt I'm digging a rabbit hole when I started. Should we make an annotation on the temporary variable and let the compiler know when it tries to get the variable from up scope? SInce reverting `registerVariableSet` in compiler will cause runtime failure.

xiaq: I feel that you took "reverting `registerVariableSet`" a bit too literally; I don't mean removing that call, but something a bit more complex:

1. When doing a temporary assignment, you need to record which new variables have been created in the scope as a side effect of that `registerVariableSet` call [here](https://github.com/elves/elvish/blob/71d050b31c1578b763cf9d6cdae50ca6fcc2956a/eval/compiler.go#L138)

2. After the entire form has been compiled, remove those new variables from the scope.

Step 2 is actually a parallel process to how you delete the name from the `Ns` when reverting the temporary assignment. The compilation process needs to mirror what happens in the runtime.

I don't think it's undoable, but because the reversion in the compilation process needs to happen in the `form` method, and `registerVariableSet` is buried several call stacks deep under, letting the `form` method know which variables are newly set and should be reverted involves either a) passing that information all the way up (meh), or b) recording it somewhere in the compiler (meh and worse). Both are kind of painful, but approach a is the lesser evil here.

krader1961:
@xiaq Another PR that seems to have fallen through the cracks. It's been eleven months since the last update and thus appears to have been abandoned. If there is merit to the proposal it should probably be turned into an enhancement request for further discussion.

xiaq:
The relevant issue is #532. I am going to close this PR; @xofyarg if you wish to continue to work on this, feel free to reopen.

PR #806, opened 2019-03-20
Namespace db paths incase $HOME is the same for multiple users

wlan0:
On my ubuntu machine, the $HOME is the same for `root` as well as the `ubuntu` user. When I switch to another user from `ubuntu` 

```sh
$ whoami
ubuntu
$ sudo -s
Cannot connect to daemon: unexpected RPC error on socket /tmp/elvish-0/sock: timeout
Daemon-related functions will likely not work.
Failed to initialize command history; disabled.
$ whoami
root
```

I looked into the codebase, and the issue was because of a timeout while trying to open the database. The database for the `root` user pointed to the database for the `ubuntu` user because of the default configuration of my ubuntu server installation. Both users had the same home_dir. Boltdb tries to acquire a lock on the existing dbfile, so that only one writer/reader can access it any time. It waits in-definitely in this case leading to the timeout.

By namespacing the db path to the uid, we sidestep this issue for multiple users sharing the same home dir. 

Tested.

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -59,7 +59,7 @@ func InitRuntime(binpath, sockpath, dbpath string) (*eval.Evaler, string) {
 		fmt.Fprintln(os.Stderr, "warning: cannot create data directory ~/.elvish")
 	} else {
 		if dbpath == "" {
-			dbpath = filepath.Join(dataDir, "db")
+			dbpath = filepath.Join(dataDir, fmt.Sprintf("%d-db",os.Getuid()))
 		}
 	}
 


xiaq:
This feels like solving the wrong problem. It seems to me that it is the setup of sharing home between root and your normal user that is broken.

wlan0:
Its a standard practice in cloud and server installations. Whenever I provision Ubuntu on the cloud, the home dir for root and Ubuntu user are the same. 

In addition, some applications run as their own user/group. (Like Postgres) These useers will not have a home dir set, or it’ll be set to the another user’a home dir. 

In some cases, users are created without a home directory(which Linux allows). Elvish should not assume that each user has a home directory. 

If we ever needed to switch to such a user, elvish would return the same error messages there too.

xiaq:
Thanks, but given the current context, I am only more convinced that the solution here is really broken and because there is no correct thing to do when two users share a home directory. Maybe you can convince me otherwise if there is more context.

> Its a standard practice in cloud and server installations. Whenever I provision Ubuntu on the cloud, the home dir for root and Ubuntu user are the same.

The word "standard" strongly depends on personal opinions; as far as I know there is no industrial standard on whether two users are supposed to share a home directory on a Unix system.

Running an interactive shell session on a cloud host is also a pretty rare thing to do; if you are starting an interactive shell to make some ad-hoc changes, this breaks the reproducibility of your deployment. One valid use case I can think of is to just play with whatever runtime the cloud host has and inspect some things. I will mention workarounds to the end of the comment.

> In addition, some applications run as their own user/group. (Like Postgres) These useers will not have a home dir set, or it’ll be set to the another user’a home dir.

If these users don't have a home directory, where should the DB be? There isn't a correct answer to this question as Elvish does not even know whether there exists a directory in the entire filesystem that the user has access to.

If these users use another user's home directory - what is the more correct thing to do, sharing the DB or segmenting them? Again, the answer depends on **why** those users share the home directory in the first place and there is no correct answer and I don't want Elvish to make that decision for the user.

Worse - if those users are used in a su(1) or sudo(1) setting, that means they inherit whichever use ran `su` or `sudo`, meaning that they will have multiple DBs depending on what the previous user. That doesn't sound like the right thing to do either.

Sharing home directories is broken in other scenarios too. For instance, some programs create files on their first startup (bash creates .bash_history, vim creates .viminfo). And if those programs are first run under `root`, the other user wouldn't be able to write those files.

If you have to work with systems where multiple users share the home directory, consider either

* Specifying an alternative `HOME`;

* Using a wrapper script that calls Elvish with explicit `-db` and `-sock` arguments to specify the path of the database and socket. This also enables the use case of sharing the same data store for multiple users, if you only specify `-sock` and let it point to the socket of another user (typically `/tmp/elvish-$uid/sock`).

There is inherent tension between the convenience of day-to-day interactive usage, and more "cloudy" or "sys-adminy" use cases that requires more rigor. In the former case it's quite reasonable to assume that no users share home directories, and in the latter case, you are better off being explicit (specifying `HOME`, or passing `-sock` and `-db`) anyway.

krader1961:
I can't speak to whether having accounts share the same home dir is common on cloud deployments. But it definitely is not common on traditional servers. And while many accounts not intended to be used interactively may share the same home dir (e.g., */var/empty* on macOS) having non-root accounts share the same home dir as the root account is a huge mistake. Don't do it.

P.S., This is also a good example for why any change in behavior should first be discussed in an issue before you go to the trouble of creating a pull request.

xiaq:
Since I have yet to hear any more arguments supporting the idea behind this PR, I am closing it.

PR #807, opened 2019-03-21
add missing newline escape

edouard-lopez:
install command for ubuntu/debian was incorrect on one command.


xiaq:
Have you run into any shell where this is an actual problem? All of `bash`, `dash` and `zsh` support trailing pipelines as a line continuation marker.

edouard-lopez:
I'm using `fish` shell

xiaq:
It sounds like you should open an issue to fish to ask for support for trailing pipes, but fair enough - I like how we can make the code snippet compatible with fish.

PR #817, opened 2019-04-07
Updating dependencies

kenshaw:
Second half to PR #801, this updates to latest module dependencies.


xiaq:
Please also updated the versions of the vendored dependencies.

kenshaw:
@xiaq originally, I did forget to do `go mod vendor`, however I immediately did a second commit, and squashed it into the original before force pushing to the PR. If you check the actual commit -- https://github.com/elves/elvish/commit/87ebf94fad988cd5d9a01c715ce5a2f23a930423 -- the changes to the vendor directory are present.

Also, for what it's worth: https://travis-ci.org/elves/elvish/builds/517006223 -- the failure on Go 1.12 seems to be due to a problem on Travis, not with this commit or changes. If you restart those builds on Travis, they should go green. I would do that myself, but of course since this is a PR, I do not have the permission on Travis to restart a build.

xiaq:
Thanks, I had a stale tab open from when I first received the PR :)

I have restarted the build, let's see how it goes.

xiaq:
Interesting - the restarted builds failed again for the same failure, so it seems to be genuine.

kenshaw:
I am not sure what the issue is. Since it's on Darwin, it will be difficult for me to track this down (I do have access to a MacBook, but I do most primary development on Linux). 

It could be a problem introduced in the newer 1.12.{3,4} versions of the Go compiler (since it varies between with the `1.12.x` name in the travis config), or it could be a change in one of the C dependencies or a combination of the two. Other, past similar errors on Darwin seemed to be related to the CGO bindings on Darwin. The Darwin platform for Go has definitely had the most troublesome errors over the last few years, especially with CGO.

kenshaw:
The problem seems to be with the `golang.org/x/sys` package. The fix is to do a `replace` in `go.mod` to force the older version. Sometime in the last 4 months, it introduced this problem on Darwin.

kenshaw:
That fixed the build/test issue on Darwin. Please see the updated (forced-pushed) change. I also had rebased it off of `master` prior to fixing the issue.

xiaq:
Hmm, can you just keep the old version of the package, and avoid doing a `replace`?

It is also weird that golang.org/x/sys should have a bug, this is a package maintained by the official Go team.

Elvish always builds with CGO disabled, and golang.org/x/sys does not have any CGO dependency. It always uses the underlying syscall interface without using the libc wrappers.

xiaq:
Also, an optional thing - if you would like to avoid your contrib stats being messed up by those vendoring commits, you can submit under a different email address than your usual one.

xiaq:
Also, the offending test code is a very simple call to `unix.Pipe`. Should it really be a bug of the library, the following code should produce a crash (when compiled with Go 1.12.x, with CGO_ENABLED=0):

```go
package main

import "golang.org/x/sys/unix"

func main() {
    var p [2]int
    unix.Pipe(p[:])
}
```

Unfortunately I don't have a macOS system at hand now...

kenshaw:
@xiaq the `replace` in the `go.mod` is the way to pin using go modules, because multiple the third party dependencies may update the version requirement. I misspoke about the CGO issue, I wasn't exactly paying attention to this and was simply sharing anecdotal information from previous breakages on Darwin that I've encountered. But yes, even though the package causing problems is supported by the Go team, this is clearly a bug. I'm sure will get worked out at some point in the future -- the fix for now is to simply pin to the existing versions (as I've done here).

With regards to stats, it doesn't really bother me. Now that it's using Go modules, I would even argue it's worth just dropping the `vendor/` directory completely, since dependency/module version resolution is handled by the Go tool itself. Yes, this was not the recommended path previously, but it is going to be the recommended route moving forward. 

I'm happy to make any specific changes requested -- let me know specifically what is necessary, and I'll go ahead and make them. Alternately, feel free to do a soft reset on this, and put the stats under your  user identity as it's not exactly code that I wrote myself. 

My goals / involvement with this project has simply been to get this moved over to Go modules (which is done), so that I can continue my work on using project so that I can fix the dependency issues in that project.

kenshaw:
For what it's worth, feel free to close this PR and do a `go get -u` in the `master` branch directly, along with the removal of  the `/vendor/` folder (I'd highly recommend this), and the `replace ...` line in the `go.mod` file. The point here was to help with getting the module dependencies up-to-date, and I believe this PR has all the information necessary to backport it to the `master` branch directly now.

xiaq:
Thanks for the explanation on the use of `replace`. This is good to merge now. I'll see if I can reproduce the issue with golang.org/x/sys.

Re. the `vendor` folder, Elvish has a policy of always supporting whichever Go versions are still supported (which is always the last 2 versions). Since Go still supports a GOPATH-based workflow (as opposed to a module-based one), Elvish will continue to support that as well. In other words, Elvish does not break any user's workflow before Go officially does. That's why I am still keen on maintaining the `vendor` directory. It will go away when module defaults to be on (as opposed to auto) for 2 versions of Go.

PR #842, opened 2019-07-19
Removed ineffectual assignments

muesli:
I assume we don't want to handle the potential error returned by readFileUTF8
intentionally. Otherwise we should add a proper error check there.

### DIFF
--- a/edit/completion/completion_mode.go
+++ b/edit/completion/completion_mode.go
@@ -389,7 +389,6 @@ func (c *completion) ListRender(width, maxHeight int) *ui.Buffer {
 			col.WriteSpaces(remainedWidth, styleForCompletion.String())
 		}
 		bb.ExtendRight(col.Buffer(), 0)
-		remainedWidth = 0
 	}
 	return bb.Buffer()
 }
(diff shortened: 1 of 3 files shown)


program/shell/interact.go
```diff
@@ -97,7 +97,7 @@ func sourceRC(ev *eval.Evaler, dataDir string) error {
 		}
 		return fmt.Errorf("cannot get full path of rc.elv: %v", err)
 	}
-	code, err := readFileUTF8(absPath)
```
xiaq: Actually, the code here should handle this error and print a warning.

edit/completion/completion_mode.go
```diff
@@ -389,7 +389,6 @@ func (c *completion) ListRender(width, maxHeight int) *ui.Buffer {
 			col.WriteSpaces(remainedWidth, styleForCompletion.String())
 		}
 		bb.ExtendRight(col.Buffer(), 0)
-		remainedWidth = 0
```
xiaq: This code was here to maintain that `remainedWidth` always reflect the width that remains, so that if more code is added below, `remainedWidth` would have the correct value.

Please add a comment there instead of removing this assignment.

PR #854, opened 2019-08-27
eval: add query commands to store module

xofyarg:
This change adds two query commands for querying command history from
underlying store. With the index from the previous result, one could
use "del-cmd" to delete it then.


xiaq:
Finally getting to this... :)

Exposing the `Cmds` and `NextCmd` methods kind of works, but the result is not good Elvish API: `Cmds` outputs a slice, but typically Elvish functions outputs multiple values; `NextCmd` outputs a string and a number, but a typical Elvish function should output a map or map-like value.

I will work on a more idiomatic way to expose these methods.

Issue #760, opened 2018-09-29
Support JSON output in -compileonly


notramo:
It is complicated to parse the output of the `-compileonly` flag if there are compilation errors.
I want to implement syntax checking into the Kakoune plugin: https://gitlab.com/notramo/elvish.kak
It would be better if this flag could support the `-json` flag.
It should also be able to output physical line and column number.

xiaq:
Letting `-json -compileonly` write JSON is a brilliant idea.

LSP has a [Diagnostic](https://microsoft.github.io/language-server-protocol/specification#diagnostic) format, we should use it.

PR #858, opened 2019-10-16
[#760] Support JSON output in -compileonly

jiujieti:
Added the support for `-json` flag and transform the errors into array JSON format.

### DIFF
--- a/program/shell/errors_to_json.go
+++ b/program/shell/errors_to_json.go
@@ -40,7 +40,7 @@ func errorToJSON(err error) []byte {
 	}
 	jsonError, errMarshal := json.Marshal(e)
 	if errMarshal != nil {
-		return []byte(`[{"message":"Unable to convert the errors to JSON format"}]`)
+		return []byte(`[{"message":"Unable to convert the errors to JSON"}]`)
 	}
 	return jsonError
 }
--- a/program/shell/errors_to_json.go
+++ b/program/shell/errors_to_json.go
@@ -1,0 +1,47 @@
+package shell
+
+import (
+	"encoding/json"
+
+	"github.com/elves/elvish/eval"
+	"github.com/elves/elvish/parse"
+)
+
+// ErrorInJSON is the data structure for self-defined error
+type ErrorInJSON struct {
+	FileName string `json:"fileName"`
+	Start    int    `json:"start"`
+	End      int    `json:"end"`
+	Message  string `json:"message"`
+}
+
+// SimplifiedErrorInJSON is the data structure for errors which contain limited info
+type SimplifiedErrorInJSON struct {
+	Message string `json:"message"`
+}
+
+// ErrorToJSON converts the error into JSON format
+func ErrorToJSON(err error) string {
+	var e interface{}
+	switch err := err.(type) {
+	case *eval.CompilationError:
+		e = []interface{}{ErrorInJSON{err.Context.Name, err.Context.Begin, err.Context.End, err.Message}}
+	case *parse.MultiError:
+		e = processMultiError(err)
+	default:
+		e = []interface{}{SimplifiedErrorInJSON{err.Error()}}
+	}
+	jsonError, er := json.Marshal(e)
+	if er != nil {
+		return `[{"message":"Unable to convert the errors to JSON format"}]`
+	}
+	return string(jsonError)
+}
+
+func processMultiError(e *parse.MultiError) []ErrorInJSON {
+	var errArr []ErrorInJSON
+	for _, v := range e.Entries {
+		errArr = append(errArr, ErrorInJSON{v.Context.Name, v.Context.Begin, v.Context.End, v.Message})
+	}
+	return errArr
+}
--- a/program/shell/errors_to_json.go
+++ b/program/shell/errors_to_json.go
@@ -20,8 +20,8 @@ type simpleErrorInJSON struct {
 	Message string `json:"message"`
 }
 
-// ErrorToJSON converts the error into JSON format.
-func ErrorToJSON(err error) []byte {
+// Converts the error into JSON.
+func errorToJSON(err error) []byte {
 	var e interface{}
 	switch err := err.(type) {
 	case *eval.CompilationError:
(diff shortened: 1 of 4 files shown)


program/shell/errors_to_json.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/elves/elvish/eval"
+	"github.com/elves/elvish/parse"
+)
+
+// ErrorInJSON is the data structure for self-defined error
+type ErrorInJSON struct {
```
xiaq: This type does not need to be exported.

Suggest naming it `errorWithContextInJSON`.

The doc comment is too vague; suggested alternative:

```go
// An auxiliary struct for converting errors with diagnostics information to JSON.
```

program/shell/errors_to_json.go
```diff
@@ -0,0 +1,47 @@
...
+	FileName string `json:"fileName"`
+	Start    int    `json:"start"`
+	End      int    `json:"end"`
+	Message  string `json:"message"`
+}
+
+// SimplifiedErrorInJSON is the data structure for errors which contain limited info
+type SimplifiedErrorInJSON struct {
```
xiaq: This type does not need to be exported.

It can be named `simpleErrorInJSON`.

The doc comment is too vague; suggested alternative:

```go
// An auxiliary struct for converting errors with only a message to JSON.
```

program/shell/errors_to_json.go
```diff
@@ -0,0 +1,47 @@
...
+	Message string `json:"message"`
+}
+
+// ErrorToJSON converts the error into JSON format
```
xiaq: Please make doc comment full sentences.

program/shell/errors_to_json.go
```diff
@@ -0,0 +1,47 @@
...
+func ErrorToJSON(err error) string {
+	var e interface{}
+	switch err := err.(type) {
+	case *eval.CompilationError:
+		e = []interface{}{ErrorInJSON{err.Context.Name, err.Context.Begin, err.Context.End, err.Message}}
```
xiaq: Break into two lines.

program/shell/errors_to_json.go
```diff
@@ -0,0 +1,47 @@
...
+	case *parse.MultiError:
+		e = processMultiError(err)
```
xiaq: You can inline the implementation of `processMultiError` here.

program/shell/errors_to_json.go
```diff
@@ -0,0 +1,47 @@
...
+	default:
+		e = []interface{}{SimplifiedErrorInJSON{err.Error()}}
+	}
+	jsonError, er := json.Marshal(e)
+	if er != nil {
+		return `[{"message":"Unable to convert the errors to JSON format"}]`
+	}
+	return string(jsonError)
```
xiaq: There is no need to convert this to a string; returning a `[]byte` is fine. You would need to convert the literal string above to a `[]byte` though.

program/shell/errors_to_json_test.go
```diff
@@ -0,0 +1,47 @@
...
+			`{"fileName":"file2","start":15,"end":16,"message":"ERR2"}]`,
+	},
+	{
+		errors.New("ERR"),
+		`[{"message":"ERR"}]`,
+	},
+}
+
+func TestErrorsToJSON(t *testing.T) {
+	for i, test := range errorsJSONTests {
```
xiaq: We have a library for table-based tests against functions; see https://github.com/elves/elvish/blob/c4189114dda0666caa37457af214e23db47a49e7/eval/vals/string_test.go#L11 for an example.

program/shell/errors_to_json.go
jiujieti: I am definitely not the biggest fan of inlining all "simple" functions. Even though `processMultiError` is not called elsewhere, the function name could still help improve the readability.

Anyway here is just a loop, will do if you ask so

program/shell/errors_to_json.go
jiujieti: sure

program/shell/errors_to_json.go
jiujieti: sure

program/shell/errors_to_json.go
jiujieti: will do

program/shell/errors_to_json.go
jiujieti: will do

program/shell/errors_to_json.go
jiujieti: sure

program/shell/errors_to_json.go
```diff
@@ -0,0 +1,46 @@
...
+		var errArr []errorInJSON
+		for _, v := range err.Entries {
+			errArr = append(errArr,
+				errorInJSON{v.Context.Name, v.Context.Begin, v.Context.End, v.Message})
+		}
+		e = errArr
+	default:
+		e = []interface{}{simpleErrorInJSON{err.Error()}}
+	}
+	jsonError, er := json.Marshal(e)
```
xiaq: Variables that keep errors are always called `err` or some name starting with `err`. Since `err` is taken in the function, call the second variable `errMarshal`.

program/shell/errors_to_json.go
```diff
@@ -0,0 +1,46 @@
...
+	if er != nil {
+		return []byte(`[{"message":"Unable to convert the errors to JSON format"}]`)
```
xiaq: "Unable to convert the errors to JSON" is sufficient, no need for "format"

program/shell/errors_to_json.go
```diff
@@ -0,0 +1,46 @@
...
+	Message  string `json:"message"`
+}
+
+// An auxiliary struct for converting errors with only a message to JSON.
+type simpleErrorInJSON struct {
+	Message string `json:"message"`
+}
+
+// ErrorToJSON converts the error into JSON format.
+func ErrorToJSON(err error) []byte {
```
xiaq: Actually this function does not need to be exported either.

xiaq:
Thank you! :D

Issue #820, opened 2019-05-07
else body not recognized in while loop


Crestwave:
According to the [docs](https://elv.sh/ref/language.html#conditional-loop:-while), this should execute if the condition is initially false, but it results in a compilation error.

```elvish
~> while $false { echo a } else { echo b }
Compilation error: too many arguments
[tty], line 1: while $false { echo a } else { echo b }
```

0x005c:
This issue is solved by #863. Should be closed.

Crestwave:
@0x005c You can mention one of `[close{,s,d}, fix{,es,ed}, resolve{,s,d}]`, then the issue number in the pull request description to automatically close the issue when merged. E.g., "resolves #863".

PR #863, opened 2019-10-27
Add else part for while statement

0x005c:
As shown in #820, `while $false { echo a } else { echo b }` emits error.
So, I fixed builtin_special.go to handle it, and added tests for that.
But, I did't add highlights for else part ( I don't know how to add a new highlight )

( I'm sorry I sent this PR with no comments by mistake, so I edited this comment )

### DIFF
--- a/eval/builtin_special.go
+++ b/eval/builtin_special.go
@@ -447,17 +447,27 @@ func compileWhile(cp *compiler, fn *parse.Form) effectOpBody {
 	args := cp.walkArgs(fn)
 	condNode := args.next()
 	bodyNode := args.nextMustLambda()
-	args.mustEnd()
+	elseNode := args.nextMustLambdaIfAfter("else")
+	args.mustEnd();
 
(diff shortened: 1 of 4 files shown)


eval/builtin_special.go
```diff
@@ -447,27 +447,41 @@ func compileWhile(cp *compiler, fn *parse.Form) effectOpBody {
...
 
 type whileOp struct {
-	condOp, bodyOp valuesOp
+	condOp, bodyOp, elseOp valuesOp
 }
 
 func (op *whileOp) invoke(fm *Frame) error {
 	body := op.bodyOp.execlambdaOp(fm)
+	elseFn := op.elseOp.execlambdaOp(fm)
+	isFirstLoop := true
```
xiaq: Can you use the same pattern as in the implementation of `for`, keeping track of whether any iteration has been executed at all, and execute the `else` body if no iteration has been executed?

0x005c:
Thank you for reviewing.
I fixed func `invoke`, added highlights for else part of `while` and test for highlights.

PR #866, opened 2019-10-27
Fix `%` not to crash when divisor is 0

0x005c:
Executing `% 1 0` on elvish cause runtime error and elvish crash.
As one solution, I implemented mod function which returns ErrArgs if divisor is 0.

### DIFF
--- a/eval/builtin_fn_num.go
+++ b/eval/builtin_fn_num.go
@@ -139,7 +139,7 @@ func divide(fm *Frame, prod float64, nums ...float64) {
 	out <- vals.FromGo(prod)
 }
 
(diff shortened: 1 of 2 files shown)


eval/builtin_fn_num.go
```diff
@@ -139,6 +139,13 @@ func divide(fm *Frame, prod float64, nums ...float64) {
 	out <- vals.FromGo(prod)
 }
 
+func mod(fm *Frame, a, b int) (int, error) {
```
xiaq: The `*Frame` argument here is superfluous; remove it.

0x005c:
Thank you for reviewing! I fixed it now :D

Issue #859, opened 2019-10-18
Support named FD in redirections


xiaq:
Elvish supports the POSIX syntax for FD redirections, like `echo 1>&2`. This means "redirect stdout (fd 1) to stderr (fd 2)". Since `echo` writes to FD 1, the whole command will write to FD 2 as a result.

The syntax is succinct, but not very readable. One way to improve it is to support using names instead of FD numbers. For instance, the former can be written as `echo stdout>&stderr`.

Note: neither the `1` nor the `stdout` before `>` is actually needed in the examples, because stdout is the default source of output redirections, but they are spelled out for the purpose of examples.

jiujieti:
I will try to work on this one

PR #868, opened 2019-10-29
[elves#859]Support named FD in redirections

jiujieti:
Fixes the issue #859


eval/unwrap.go
```diff
@@ -102,7 +108,20 @@ func (u ValueUnwrapper) FdOrClose() (int, error) {
 	if err == nil && s == "-" {
 		return -1, nil
 	}
-	return u.NonNegativeInt()
+	switch s {
```
xiaq: You can delegate all this work to the `Fd` method.

Also put the `Fd` method just above or under this method as they are closely related.

eval/unwrap.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/elves/elvish/eval/vals"
 )
 
+const (
```
xiaq: These are not exported symbols, so don't give them names starting with a capital letter.

They should be only be used in one place and are very common names for the 3 standard files, so it's simpler to remove the constants and use string literals in the functions below,

PR #869, opened 2019-10-29
[#859]Support named FD in redirections

jiujieti:
this fixes the issue #859

### DIFF
--- a/eval/unwrap.go
+++ b/eval/unwrap.go
@@ -123,8 +123,8 @@ func (u ValueUnwrapper) FdOrClose() (int, error) {
 	if err == nil && s == "-" {
 		return -1, nil
 	}
-	fd, errFd := u.Fd()
-	if errFd != nil {
+	fd, err := u.Fd()
+	if err != nil {
 		return 0, u.ctx.errorpf(u.begin, u.end, "redirection source must be standard stream name or integer; got %s", s)
 	}
 	return fd, nil
--- a/eval/unwrap.go
+++ b/eval/unwrap.go
@@ -97,12 +97,37 @@ func (u ValueUnwrapper) NonNegativeInt() (int, error) {
 	return i, u.err
 }
 
+func (u ValueUnwrapper) Fd() (int, error) {
+	s, err := u.String()
+	if err != nil {
+		return 0, err
+	}
+	switch s {
+	case "stdin":
+		return 0, nil
+	case "stdout":
+		return 1, nil
+	case "stderr":
+		return 2, nil
+	default:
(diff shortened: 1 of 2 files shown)


eval/unwrap.go
```diff
@@ -97,12 +97,37 @@ func (u ValueUnwrapper) NonNegativeInt() (int, error) {
...
+	case "stdin":
+		return 0, nil
+	case "stdout":
+		return 1, nil
+	case "stderr":
+		return 2, nil
+	default:
+		i, errInt := u.NonNegativeInt()
+		if errInt != nil {
+			return 0, fmt.Errorf("Fd must be standard stream names or integer; got %s", s)
```
xiaq: Can you use `u.errorpf` to build the error, like in the `error` method above? This way the error will have the correct information about the range.

eval/unwrap.go
```diff
@@ -97,12 +97,37 @@ func (u ValueUnwrapper) NonNegativeInt() (int, error) {
...
+
 func (u ValueUnwrapper) FdOrClose() (int, error) {
 	s, err := u.String()
 	if err == nil && s == "-" {
 		return -1, nil
 	}
-	return u.NonNegativeInt()
+	fd, errFd := u.Fd()
+	if errFd != nil {
+		return 0, fmt.Errorf("redirection source must be standard stream names or integer; got %s", s)
```
xiaq: There will be no need to build this new error if you address my comment above.

eval/unwrap.go
```diff
@@ -97,12 +97,37 @@ func (u ValueUnwrapper) NonNegativeInt() (int, error) {
...
+	}
+	switch s {
+	case "stdin":
+		return 0, nil
+	case "stdout":
+		return 1, nil
+	case "stderr":
+		return 2, nil
+	default:
+		i, errInt := u.NonNegativeInt()
```
xiaq: You can reuse the name `err` here. It is common pattern in Go unless you have to store multiple errors.

eval/unwrap.go
```diff
@@ -97,12 +97,37 @@ func (u ValueUnwrapper) NonNegativeInt() (int, error) {
...
+	}
+}
+
 func (u ValueUnwrapper) FdOrClose() (int, error) {
 	s, err := u.String()
 	if err == nil && s == "-" {
 		return -1, nil
 	}
-	return u.NonNegativeInt()
+	fd, errFd := u.Fd()
```
xiaq: You can reuse the name `err` here. It is common pattern in Go unless you have to store multiple errors.

Issue #873, opened 2019-12-04
Inconsistent return for -compileonly -json


sblundy:
If the input has a compilation error, a missing variable for instance, `elvish -compileonly -json` outputs JSON objects with information about the error's location in separate fields. But when it's a parse error, the objects only have the message field.

Example:

**val-missing.elv**
```
if $x {
	echo 'yes'
}
```

**bracket-missing.elv**
```
if $true {
	echo 'yes'

```

```bash
$ elvish -compileonly -json val-missing.elv
[{"fileName":"/val-missing.elv","start":3,"end":5,"message":"variable $x not found"}]
Exception: elvish exited with 2
[tty], line 1: elvish -compileonly -json val-missing.elv
$ elvish -compileonly -json bracket-missing.elv
[{"message":"parse error: 23-23 in bracket-missing.elv: should be '}'"}]
Exception: ./elvish exited with 2
[tty], line 1: elvish -compileonly -json bracket-missing.elv
```

The first has `fileName`, `start`, `end`, and `message` fields. The second just has `message`

xiaq:
Fixed.

PR #874, opened 2019-12-04
[#873] fix json output of -compileonly

sblundy:
The [`errorToJSON()`](https://github.com/elves/elvish/blob/master/program/shell/errors_to_json.go#L31) method checks whether the error returned is a `*parse.MultiError`. However, rather than a pointer to a `MultiError`, an instance was being passed in. 

Updated the the instance methods of `MultiError` to take pointers as receivers to be in line with what was expected. This further required updating some other references.

### DIFF
--- a/cliedit/highlight/highlight.go
+++ b/cliedit/highlight/highlight.go
@@ -26,7 +26,7 @@ func highlight(code string, dep Dep, lateCb func(styled.Text)) (styled.Text, []e
 
 	n, errParse := parse.AsChunk("[interactive]", code)
 	if errParse != nil {
-		for _, err := range errParse.(parse.MultiError).Entries {
+		for _, err := range errParse.(*parse.MultiError).Entries {
 			if err.Context.Begin != len(code) {
 				errors = append(errors, err)
 			}
--- a/edit/edcore/edit.go
+++ b/edit/edcore/edit.go
@@ -273,7 +273,7 @@ func atEnd(e error, n int) bool {
 	switch e := e.(type) {
 	case *eval.CompilationError:
 		return e.Context.Begin == n
-	case parse.MultiError:
+	case *parse.MultiError:
 		for _, entry := range e.Entries {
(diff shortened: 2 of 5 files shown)


xiaq:
Hmm, what about just changing the type switch in errotToJSON?

sblundy:
> Hmm, what about just changing the type switch in errotToJSON?

Consistency mostly. Pointer receivers are the norm, instance ones only when necessary. `MultiError` had a mix of both. Unless there's a reason that I missed, inconsistency tends to foster bugs.

Thinking about it more, the most idiomatic way would be with an interface that both `eval.CompilationError` and `parse.MultiError` would implement, but that'd be an even bigger change.

Something like this:

```golang
func errorToJSON(err error) []byte {
	var e interface{}
	type hasParseErrors interface {
		ParseErrors() []*parse.Error
	}
	switch err := err.(type) {
	case hasParseErrors:
		var errArr []errorInJSON
		for _, v := range err.ParseErrors() {
			errArr = append(errArr,
				errorInJSON{v.Context.Name, v.Context.Begin, v.Context.End, v.Message})
		}
		e = errArr
	default:
		e = []interface{}{simpleErrorInJSON{err.Error()}}
	}
	jsonError, errMarshal := json.Marshal(e)
	if errMarshal != nil {
		return []byte(`[{"message":"Unable to convert the errors to JSON"}]`)
	}
	return jsonError
}
```

xiaq:
Ah, this type has methods of both value and pointer receivers. I agree that that is bad and your PR is in the correct direction, which I will merge.

But I am not convinced on the larger point that that pointer receivers are the norm; I don't think there is such consensus in the Go community. Value receivers are useful from time to time.

Regarding how to refactor this further: `eval.CompilationError` and `parse.Error` have the same fields, just slightly different implementations of how they print themselves, because they have different types. We can consolidate the difference by introducing a more general Error type with an extra "Type" field. The `diag` package is a good place for this new Error type.

Issue #891, opened 2020-01-02
panic in location mode when no path matched


zhsj:
```
~/go/src/github.com/elves/elvish>                                                                                                                                            zhsj@zhsj-debian
 LOCATION  xxx
(0x97dae0,0xc000262060)
goroutine 1 [running]:
github.com/elves/elvish/pkg/sys.DumpStack(0x97dae0, 0xc000262060)
        github.com/elves/elvish/pkg/sys/dumpstack.go:10 +0x9d
github.com/elves/elvish/pkg/program/shell.rescue()
        github.com/elves/elvish/pkg/program/shell/shell.go:66 +0xa4
panic(0x97dae0, 0xc000262060)
        /usr/lib/go-1.13/src/runtime/panic.go:679 +0x1b2
github.com/elves/elvish/pkg/cli/addons/location.Start.func3(0xa7fb00, 0xc000125e60, 0x0)
        github.com/elves/elvish/pkg/cli/addons/location/location.go:97 +0x232
github.com/elves/elvish/pkg/cli.(*listBox).Accept(0xc000210510)
        github.com/elves/elvish/pkg/cli/listbox.go:390 +0x74
github.com/elves/elvish/pkg/cli.(*listBox).Handle(0xc000210510, 0xa79f40, 0xc00024c1b8, 0x441c74)
        github.com/elves/elvish/pkg/cli/listbox.go:264 +0x174
github.com/elves/elvish/pkg/cli.(*comboBox).Handle(0xc00020ee00, 0xa79f40, 0xc00024c1b8, 0x0)
        github.com/elves/elvish/pkg/cli/combobox.go:62 +0x63
github.com/elves/elvish/pkg/cli.(*app).handle(0xc0001cc000, 0x954700, 0xc00024c1b8)
        github.com/elves/elvish/pkg/cli/app.go:164 +0x1e8
github.com/elves/elvish/pkg/cli.(*loop).Run(0xc0001b23c0, 0xc0001c2001, 0xc0001d4180, 0xc0001cc000, 0xc00024c0a0)
        github.com/elves/elvish/pkg/cli/loop.go:129 +0x230
github.com/elves/elvish/pkg/cli.(*app).ReadCode(0xc0001cc000, 0x0, 0x0, 0x0, 0x0)
        github.com/elves/elvish/pkg/cli/app.go:331 +0x4eb
[...]

PR #892, opened 2020-01-02
pkg/cli: fix index out of range when selecting state items

zhsj:
Fixes: #891

### DIFF
--- a/pkg/cli/listbox_test.go
+++ b/pkg/cli/listbox_test.go
@@ -452,3 +452,32 @@ func TestListBox_Select_CallOnSelect(t *testing.T) {
 	w.Select(func(ListBoxState) int { return 0 })
 	verifyOnSelect(0)
 }
+
+func TestListBox_Accept_IndexCheck(t *testing.T) {
+	tests := []struct {
+		name         string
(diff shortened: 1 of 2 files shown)


zhsj:
Fixed.

pkg/cli/listbox_test.go
```diff
@@ -452,3 +452,31 @@ func TestListBox_Select_CallOnSelect(t *testing.T) {
...
+	}{
+		{"index in range", 1, 0},
+		{"index exceeds left boundary", 1, -1},
+		{"index exceeds right boundary", 0, 0},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			w := NewListBox(ListBoxSpec{
+				OnAccept: func(it Items, i int) {
+					if i >= it.Len() || i < 0 {
```
xiaq: Avoid repeating the logic of the code under test.

Instead, introduce a `shouldAccept` field in the test cases, and use `!shouldAccept` as the condition.

zhsj: make sense!

Issue #524, opened 2017-12-07
Add incognito mode


notramo:
I would like to see that Elvish has an incognito mode.
I mean not saving command/dir history under *incognito mode*.
E.g. some programs require to pass them passwords in the command line, and it is bad if these commands are saved to history.

xiaq:
You can prepend a space to the command to prevent it from going into the history, but there is no analog in dir history. An "incognito mode" sounds good.

krader1961:
I have never liked the "prepend a space to a command" idiom used by legacy POSIX shells to keep the command from being persisted to the command history. It is too easy to use mistakenly and too hard to discover by new users. I would much prefer an explicit *incognito* mode that deals with commands, directory changes, and anything else considered sensitive. The question, it seems to me, is whether this should be something that can be toggled or should require running elvish with an option that enables icognito mode. Similar to how web browsers like Chrome operate. I favor the latter approach. That is having to do something like `elvish -incognito`. Possibly removing the command space prefix hack.

iwoloschin:
There should be a [variable](https://elv.sh/ref/builtin.html#variables) that shows if incognito mode is active, it'd be very useful for prompts to display mode.

krader1961:
> There should be a [variable](https://elv.sh/ref/builtin.html#variables) that shows if incognito mode is active, it'd be very useful for prompts to display mode.

Not only that but we now have several examples of variables that can be queried but also modify Elvish behavior when assigned to; e.g., https://elv.sh/ref/unix.html#unix:rlimits and https://elv.sh/ref/unix.html#unix:umask. An `edit:incognito` variable could function in a similar fashion. It would default to false but you could do `set edit:incognito = $true` to enable incognito mode. Something you could wrap in a function or abbreviation if you used it a lot and wanted something shorter to type when enabling/disabling incognito mode. Personally, I would prefer such a mechanism because it would be easier to discover by new users and less likely to result in surprising behavior if a user accidentally begins a command line with a space.

P.S., Is the leading space to keep a command from being appended to the history actually documented? I tried really hard to find such documentation and failed. Which sort of makes my point even if it is documented. And if it isn't documented I don't see an obvious way to document it in a manner that makes it easy to discover. It's also notable that Bash, by default, does not recognize a leading space as special. You have to `HISTCONTROL=ignorespace`. So apparently the Bash maintainers also concluded that "don't record commands with a leading space" is not desirable by default.

krader1961:
[...]

PR #895, opened 2020-01-11
pkg/edit: Introduce before-add-cmd hook

xofyarg:
edit:before-add-cmd hook let its functions check the command about to save
and output a stirng. If the string is non-empty, it will be passed to
the next function in the hook chain as input and finally saved to db;
otherwise the command will be ignored.

Edit: this variable has been renamed to `edit:add-cmd-filters` as suggested by @xiaq. A `filter` is a bit different than a `hook`, it outputs a boolean value to indicate whether the argument to it passed the the check or not.

### DIFF
--- a/pkg/edit/config_api.go
+++ b/pkg/edit/config_api.go
@@ -49,6 +49,16 @@ func initAfterReadline(appSpec *cli.AppSpec, ev *eval.Evaler, ns eval.Ns) {
 	})
 }
 
+//elvdoc:var add-cmd-filters
+//
+// List of filters to run before adding a command to history.
+//
+// A filter is a function that takes a command as argument and outputs
+// a boolean value. If any of the filters outputs `$false`, the
+// command is not saved to history, and the rest of the filters are
+// not run. The default value of this list contains a filter which
+// ignores command starts with space.
+
 func callHooks(ev *eval.Evaler, name string, hook vals.List, args ...interface{}) {
 	i := -1
 	for it := hook.Iterator(); it.HasElem(); it.Next() {
@@ -68,6 +78,44 @@ func callHooks(ev *eval.Evaler, name string, hook vals.List, args ...interface{}
 		fm := eval.NewTopFrame(ev, parse.Source{Name: name}, ports)
 		fn.Call(fm, args, eval.NoOpts)
 	}
+}
+
+func callFilters(ev *eval.Evaler, name string, filters vals.List, args ...interface{}) bool {
+	i := -1
+	for it := filters.Iterator(); it.HasElem(); it.Next() {
+		i++
+		name := fmt.Sprintf("%s[%d]", name, i)
+		fn, ok := it.Elem().(eval.Callable)
+		if !ok {
+			// TODO(xiaq): This is not testable as it depends on stderr.
+			// Make it testable.
+			diag.Complainf(os.Stderr, "%s not function", name)
+			continue
+		}
+		// TODO(xiaq): This should use stdPorts, but stdPorts is currently
+		// unexported from eval.
+		ports := []*eval.Port{
+			eval.DevNullClosedChan, {File: os.Stdout}, {File: os.Stderr}}
+		fm := eval.NewTopFrame(ev, parse.Source{Name: name}, ports)
+		out, err := fm.CaptureOutput(func(fm *eval.Frame) error { return fn.Call(fm, args, eval.NoOpts) })
+		if err != nil {
+			diag.Complainf(os.Stderr, "%s return error", name)
+			continue
+		}
+		if len(out) != 1 {
+			diag.Complainf(os.Stderr, "filter %s should only return $true or $false", name)
+			continue
+		}
+		p, ok := out[0].(bool)
+		if !ok {
+			diag.Complainf(os.Stderr, "filter %s should return bool", name)
+			continue
+		}
+		if !p {
+			return false
+		}
+	}
+	return true
 }
 
 func newIntVar(i int) vars.PtrVar            { return vars.FromPtr(&i) }
--- a/pkg/edit/editor.go
+++ b/pkg/edit/editor.go
@@ -39,6 +39,9 @@ type notifier interface {
 
 // NewEditor creates a new editor from input and output terminal files.
 func NewEditor(tty cli.TTY, ev *eval.Evaler, st store.Store) *Editor {
+	// Declare the Editor with a nil App first; some initialization functions
+	// require a notifier as an argument, but does not use it immediately.
+	ed := &Editor{ns: eval.Ns{}, excList: vals.EmptyList}
 	appSpec := cli.AppSpec{TTY: tty}
 
 	fuser, err := histutil.NewFuser(st)
@@ -47,17 +50,21 @@ func NewEditor(tty cli.TTY, ev *eval.Evaler, st store.Store) *Editor {
 	}
 
 	if fuser != nil {
+		ignoreLeadingSpace := eval.NewGoFn("<ignore-cmd-with-leading-space>",
+			func(s string) bool { return !strings.HasPrefix(s, " ") })
+		filters := newListVar(vals.MakeList(ignoreLeadingSpace))
+		ed.ns["add-cmd-filters"] = filters
+
 		appSpec.AfterReadline = []func(string){func(code string) {
-			if code != "" && !strings.HasPrefix(code, " ") {
+			if code != "" &&
+				callFilters(ev, "$<edit>:add-cmd-filters",
+					filters.Get().(vals.List), code) {
 				fuser.AddCmd(code)
 			}
 			// TODO(xiaq): Handle the error.
 		}}
 	}
 
-	// Declare the Editor with a nil App first; some initialization functions
-	// require a notifier as an argument, but does not use it immediately.
-	ed := &Editor{ns: eval.Ns{}, excList: vals.EmptyList}
 	initHighlighter(&appSpec, ev)
 	initConfigAPI(&appSpec, ev, ed.ns)
 	initInsertAPI(&appSpec, ed, ev, ed.ns)
--- a/pkg/edit/editor_test.go
+++ b/pkg/edit/editor_test.go
@@ -27,14 +27,95 @@ func TestEditor_DoesNotAddEmptyCommandToHistory(t *testing.T) {
 	testCommands(t, f.Store /* no commands */)
 }
 
-func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
-	f := setup()
+func TestEditor_TestAddCmdFilters(t *testing.T) {
+	cases := []struct {
+		name        string
+		rc          string
+		input       string
+		wantHistory []string
+	}{
+		// TODO: Enable the following two tests once error output can
+		// be tested.
+		// {
+		// 	name:        "non callable item",
+		// 	rc:          "edit:add-cmd-filters = [$false]",
+		// 	input:       "echo\n",
+		// 	wantHistory: []string{"echo"},
+		// },
+		// {
+		// 	name:        "callback outputs nothing",
+		// 	rc:          "edit:add-cmd-filters = [[_]{}]",
+		// 	input:       "echo\n",
+		// 	wantHistory: []string{"echo"},
+		// },
+		{
+			name:        "callback outputs true",
+			rc:          "edit:add-cmd-filters = [[_]{ put $true }]",
+			input:       "echo\n",
+			wantHistory: []string{"echo"},
+		},
+		{
+			name:        "callback outputs false",
+			rc:          "edit:add-cmd-filters = [[_]{ put $false }]",
+			input:       "echo\n",
+			wantHistory: nil,
+		},
+		{
+			name:        "false-true chain",
+			rc:          "edit:add-cmd-filters = [[_]{ put $false } [_]{ put $true }]",
+			input:       "echo\n",
+			wantHistory: nil,
+		},
+		{
+			name:        "true-false chain",
+			rc:          "edit:add-cmd-filters = [[_]{ put $true } [_]{ put $false }]",
+			input:       "echo\n",
+			wantHistory: nil,
+		},
+		{
+			name:        "positive",
+			rc:          "edit:add-cmd-filters = [[cmd]{ ==s $cmd echo }]",
+			input:       "echo\n",
+			wantHistory: []string{"echo"},
+		},
+		{
+			name:        "negative",
+			rc:          "edit:add-cmd-filters = [[cmd]{ ==s $cmd echo }]",
+			input:       "echo x\n",
+			wantHistory: nil,
+		},
+		{
+			name:        "default value",
+			rc:          "",
+			input:       " echo\n",
+			wantHistory: nil,
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			f := setup(rc(c.rc))
+			defer f.Cleanup()
+
+			feedInput(f.TTYCtrl, c.input)
+			f.Wait()
+
+			testCommands(t, f.Store, c.wantHistory...)
+		})
+	}
+}
+
+func TestEditor_AddCmdFiltersHasShortCircuit(t *testing.T) {
+	f := setup(rc(
+		`called = $false`,
+		`@edit:add-cmd-filters = [_]{ put $false } [_]{ called = $true; put $true }`,
+	))
 	defer f.Cleanup()
 
-	feedInput(f.TTYCtrl, " echo\n")
+	feedInput(f.TTYCtrl, "echo\n")
 	f.Wait()
-
-	testCommands(t, f.Store /* no commands */)
+	testCommands(t, f.Store)
+	testGlobal(t, f.Evaler, "called", false)
 }
 
 func testCommands(t *testing.T, store store.Store, wantCmds ...string) {


xofyarg:
Use case: do not save command which contains an auth token.
```
fn ignore-auth-token [cmd]{
   if (re:match "X-AUTH" $cmd) { nop } else { put $cmd }
}

edit:before-add-cmd = [$ignore-auth-token~]
```

pkg/edit/config_api.go
```diff
@@ -69,6 +69,44 @@ func callHooks(ev *eval.Evaler, name string, hook vals.List, args ...interface{}
...
+		// unexported from eval.
+		ports := []*eval.Port{
+			{File: os.Stdin}, {File: os.Stdout}, {File: os.Stderr}}
+		fm := eval.NewTopFrame(ev, eval.NewInternalGoSource(name), ports)
+		out, err := fm.CaptureOutput(fn, args, eval.NoOpts)
+		if err != nil {
+			diag.Complainf("%s return error", name)
+			continue
+		}
+		if len(out) != 1 {
```
krader1961: I don't see how this enforces the requirement that the filter output a single `$true` or `$false`. And, in fact, your single test case outputs `$false`. What happens if the filter outputs `argle`, `1`, `(float64 3.14)`?

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,16 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
 	testCommands(t, f.Store /* no commands */)
 }
 
+func TestEditor_DoesCommandFilterBeforeAddToHistory(t *testing.T) {
```
xiaq, line 40: TestEditor_FiltersCommandBeforeAddingToHistory

pkg/edit/config_api.go
xiaq: Line 98 checks the type of the output. There should be a test for this but this is not testable at the moment as the warning message can only be written to stderr.

I plan to refactor `Evaler` a bit, including adding a `ports` field to it (now only `Frame` has ports), which will make this testable in future.

pkg/edit/config_api.go
```diff
@@ -69,6 +69,44 @@ func callHooks(ev *eval.Evaler, name string, hook vals.List, args ...interface{}
...
+			// Make it testable.
+			diag.Complainf("%s not function", name)
+			continue
+		}
+		// TODO(xiaq): This should use stdPorts, but stdPorts is currently
+		// unexported from eval.
+		ports := []*eval.Port{
+			{File: os.Stdin}, {File: os.Stdout}, {File: os.Stderr}}
+		fm := eval.NewTopFrame(ev, eval.NewInternalGoSource(name), ports)
+		out, err := fm.CaptureOutput(fn, args, eval.NoOpts)
```
xiaq, line 89: The signature of `CaptureOutput` has been changed, do this instead:

```go
fm.CaptureOutput(func(fm *Frame) error { return fn.Call(fm, args, eval.NoOpts) })
```

pkg/edit/config_api.go
```diff
@@ -69,6 +69,44 @@ func callHooks(ev *eval.Evaler, name string, hook vals.List, args ...interface{}
...
+		if !ok {
+			// TODO(xiaq): This is not testable as it depends on stderr.
+			// Make it testable.
+			diag.Complainf("%s not function", name)
+			continue
+		}
+		// TODO(xiaq): This should use stdPorts, but stdPorts is currently
+		// unexported from eval.
+		ports := []*eval.Port{
+			{File: os.Stdin}, {File: os.Stdout}, {File: os.Stderr}}
```
xiaq, line 87: This function should not read inputs, use `eval.DevNullClosedChan` as the first FD.

pkg/edit/config_api.go
```diff
@@ -69,6 +69,44 @@ func callHooks(ev *eval.Evaler, name string, hook vals.List, args ...interface{}
...
+		fm := eval.NewTopFrame(ev, eval.NewInternalGoSource(name), ports)
```
xiaq, line 88: The `NewInternalGoSource` function has been removed, use `parse.Source{Name: name}` here.

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,16 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
...
+	f := setup(rc("edit:add-cmd-filters = [[_]{ put $false }]"))
```
xiaq, line 41: The test passes for the following incorrect implementations:

* Suppress history as long as `$edit:add-cmd-filters` is not empty
* Suppress history as long as function in `$edit:add-cmd-filters` has any output at all

Use a more complex filter function that can return true or false for different inputs.

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,16 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
...
+	defer f.Cleanup()
+
+	feedInput(f.TTYCtrl, "echo x\n")
+	f.Wait()
+
+	testCommands(t, f.Store /* no commands */)
+}
+
```
xiaq, line 49: Add a test case for the logical-and semantics of `$edit:add-cmd-filters.`

pkg/edit/config_api.go
```diff
@@ -69,6 +69,44 @@ func callHooks(ev *eval.Evaler, name string, hook vals.List, args ...interface{}
 	}
 }
 
+func callFilters(ev *eval.Evaler, name string, filters vals.List, args ...interface{}) bool {
```
xiaq, line 72: Please add an elvdoc for `$edit:add-cmd-filters`, see https://github.com/elves/elvish/blob/master/CONTRIBUTING.md#documenting-builtin-functions-and-variables.

krader1961:
I vote no for merging this change even if it is updated to resolve prior feedback from @xiaq. This appears to be intended to implement a very subtle security mechanism. So subtle that it isn't obvious if the security mechanism is in effect or not. This is a dangerous pseudo-security mechanism. The incognito mode discussed in issue #524 is a better, meaning safer, approach to the problem. Force the user to enable incognito mode if they don't want their commands to be recorded.

P.S., It's not obvious what is meant by

> ...otherwise the command will be ignored...

Is an exception thrown? Presumably the literal meaning of that phrase is not true. That is, the behavior is as if the command ran with a success status even though the command was not actually run.

xofyarg:
> intended to implement a very subtle security mechanism

I'm not sure if I understand this correctly, do you mean the use case I showed above? It might be my "so subtle that it isn't obvious" issue, but it's quite different than disabling command history for a whole session.

This PR intents to give user ability to decide whether or not a command should be added to history. Several use cases I can think of:
- Ignore a sensitive command which contains a token, a password, etc. The hook function can try to match a special keyword in such commands, or like bash, give command a space as prefix to make it special.
- Don't save meaningless command if it contains only "ls", "cd" for instance.
- Implement the "incognito mode" by adding a hook function which always return `$false`.

I'm pretty sure people have their own workflow, the point is to have a common, flexible interface, which can then be extended via modules.

zzamboni:
I think having this hooks would be useful in the general case. @krader1961: the use case of "not saving to history if it contains a password" is only one potential use case, but I can imagine others. This is separate from incognito mode since it could be used for other things (e.g. modifying or normalizing commands in some way before saving to history).

I'm a big fan of hooks in general, so I don't see any harm in having this.

xiaq:
I think hooks are at worst harmless and am open to more of them (including this one).

When we have too many if them it will be worth thinking about whether some are unnecessary, but we are not there yet.

krader1961:
I'm not opposed to hooks. On the contrary, I have some in mind that might have merit. For example, rather than hardcoding an execution tracing mechanism as is done in shells like ksh it would be better to do it via hooks that would allow the user to customize how the execution tracing is implemented.

Having thought about it some more I agree this is a generally useful feature. In particular, the current hardcoded special-case of not recording command lines that have a leading space should be implemented in terms of this mechanism. Hopefully examples like the X-AUTH token case will never be default behavior. Which means that any user depending on such behavior will have to explicitly enable it. Making it their responsibility to ensure it doesn't accidentally leak sensitive data into the command history.

I haven't looked at the implementation. Presumably it handles multiline commands and a compound statement involving pipelines calls the hook just once with the entire statement.

xofyarg:
@xiaq I've updated the PR. It should resolve all the issues you mentioned above. Please take another look, thanks.

pkg/edit/config_api.go
```diff
@@ -49,6 +49,17 @@ func initAfterReadline(appSpec *cli.AppSpec, ev *eval.Evaler, ns eval.Ns) {
 	})
 }
 
+//elvdoc:var add-cmd-filters
+//
+// List of filter to run before add a command to history.
```
xiaq, line 54: List of filters to run before adding a command to history.

pkg/edit/config_api.go
```diff
@@ -49,6 +49,17 @@ func initAfterReadline(appSpec *cli.AppSpec, ev *eval.Evaler, ns eval.Ns) {
...
+//
+// A filter is a function which takes current command as input and
+// outputs a boolean value. `$false` indicates current command should
+// not be saved into history, while `$true` means the command passed
+// this filter and should be checked by next filter. Current command
+// will be saved only if the variable is empty, or it passed all the
+// filters.
```
xiaq, line 61: These can be more rephrased to be more succinct and no less clear:

```markdown
A filter is a function that takes a command as argument and outputs a boolean value. If any of the filters outputs `$false`, the command is not saved to history, and the rest of the filters are not run.
```

pkg/edit/editor.go
```diff
@@ -47,17 +50,18 @@ func NewEditor(tty cli.TTY, ev *eval.Evaler, st store.Store) *Editor {
 	}
 
 	if fuser != nil {
+		filters := newListVar(vals.EmptyList)
+		ed.ns["add-cmd-filters"] = filters
+
 		appSpec.AfterReadline = []func(string){func(code string) {
-			if code != "" && !strings.HasPrefix(code, " ") {
+			if code != "" && !strings.HasPrefix(code, " ") &&
```
xiaq, line 57: Now that we have this general filtering mechanism, can you implement the default filter (skipping commands that start with a space) as a *default value* of the filter list?

You can define a simple builtin function as the default element: `func(s string) bool { return !strings.HasPrefix(s, " ") }`

pkg/edit/editor_test.go
```diff
@@ -27,7 +27,7 @@ func TestEditor_DoesNotAddEmptyCommandToHistory(t *testing.T) {
 	testCommands(t, f.Store /* no commands */)
 }
 
-func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
+func TestEditor_DoesCommandWithLeadingSpaceNotToHistory(t *testing.T) {
```
xiaq, line 30: `TestEditor_DoesNotSaveCommandWithLeadingSpace`

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,37 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
 	testCommands(t, f.Store /* no commands */)
 }
 
+func TestEditor_FiltersCommandBeforeAddingToHistory(t *testing.T) {
+	noHistory := "_EMPTY_"
+
+	test := func(c, input, exp string) {
```
xiaq, line 43: If you let this function take variadic arguments and pass that to `testCommands`, you don't need to special-case `noHistory`.

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,37 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
...
+	// function has no output, invalid
+	test("edit:add-cmd-filters = [[_]{}]", "echo\n", "echo")
+	test("edit:add-cmd-filters = [[_]{ put $true }]", "echo\n", "echo")
+	test("edit:add-cmd-filters = [[_]{ put $false }]", "echo\n", noHistory)
+	// filters are concatenated internally with logical "and"
+	test("edit:add-cmd-filters = [[_]{ put $false } [_]{ put $true }]", "echo\n", noHistory)
+	test("edit:add-cmd-filters = [[_]{ put $true } [_]{ put $false }]", "echo\n", noHistory)
+
+	test("edit:add-cmd-filters = [[cmd]{ ==s $cmd echo }]", "echo\n", "echo")
+	test("edit:add-cmd-filters = [[cmd]{ ==s $cmd echo }]", "echo x\n", noHistory)
```
xiaq, line 68: Prefer one of the following two styles:

* Put all test cases in a table, and iterate through the table in this test. If you do that, add a `name` field to the table, and use [`t.Run`](https://godoc.org/testing#T.Run) so that each case is run as a subtest.

* Make each of the test a top-level `TestBeforeAddCmd_Foo` function; also make the `test` closure a top-level, unexported `testBeforeAddCmd` function, so that all the `TestBeforeAddCmd_Foo` function just calls `testBeforeAddCmd`.

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,37 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
 	testCommands(t, f.Store /* no commands */)
 }
 
+func TestEditor_FiltersCommandBeforeAddingToHistory(t *testing.T) {
```
xiaq, line 40: I'd call this `TestBeforeAddCmd`.

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,37 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
...
+			testCommands(t, f.Store, exp)
+		}
+	}
+
+	// item is not a callable, invalid
+	test("edit:add-cmd-filters = [$false]", "echo\n", "echo")
+	// function has no output, invalid
+	test("edit:add-cmd-filters = [[_]{}]", "echo\n", "echo")
+	test("edit:add-cmd-filters = [[_]{ put $true }]", "echo\n", "echo")
+	test("edit:add-cmd-filters = [[_]{ put $false }]", "echo\n", noHistory)
```
xiaq, line 62: IIUC these will write error messages to stderr when run. Do one the following:

* Make `callFilters` use a mutable global, instead of `os.Stderr`, and assign it to `eval.DevNull` during testing.

* Even better, assign the global to a pipe and check the output, and test the error message too. This probably creates a bit too much work for the sake of the PR so I'm OK with the lazier approach above.

xofyarg: As the error output is an arbitrary string, it doesn't seem stable enough to match on it. Maybe we could have `diag.Complainf` use the global value internally.

How about a separated PR to improve the test for `callHooks` and `callFilters`?

pkg/edit/config_api.go
```diff
@@ -49,6 +49,15 @@ func initAfterReadline(appSpec *cli.AppSpec, ev *eval.Evaler, ns eval.Ns) {
...
 
+//elvdoc:var add-cmd-filters
+//
+// List of filters to run before adding a command to history.
+//
+// A filter is a function that takes a command as argument and outputs
+// a boolean value. If any of the filters outputs `$false`, the
+// command is not saved to history, and the rest of the filters are
+// not run.
+
```
xiaq, line 60: Please document the default value here.

pkg/edit/editor_test.go
```diff
@@ -27,7 +27,7 @@ func TestEditor_DoesNotAddEmptyCommandToHistory(t *testing.T) {
 	testCommands(t, f.Store /* no commands */)
 }
 
-func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
+func TestEditor_DoesNotSaveCommandWithLeadingSpace(t *testing.T) {
```
xiaq, line 30: This can now be made part of the test below.

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,36 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
 	testCommands(t, f.Store /* no commands */)
 }
 
+func TestEditor_TestBeforeAddCmd(t *testing.T) {
+	cases := []struct {
+		name string
+		rc   string
+		in   string
+		exp  []string
```
xiaq, line 45: Follow the naming convention of other test code, and call this field `wantHistory`.

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,36 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
...
+	}{
+		{"nonCallableItem", "edit:add-cmd-filters = [$false]", "echo\n", []string{"echo"}},
```
xiaq, line 47: Write out the name as words and make them more complete; this is intended for human consumption. For example, this name is fine, but can be spread out as `"non callable item"`

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,36 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
...
+		{"noOutputItem", "edit:add-cmd-filters = [[_]{}]", "echo\n", []string{"echo"}},
+		{"true", "edit:add-cmd-filters = [[_]{ put $true }]", "echo\n", []string{"echo"}},
```
xiaq, line 49: This name can be `"callback outputs true"`

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,36 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
...
+		{"false", "edit:add-cmd-filters = [[_]{ put $false }]", "echo\n", nil},
+		{"chain1", "edit:add-cmd-filters = [[_]{ put $false } [_]{ put $true }]", "echo\n", nil},
+		{"chain2", "edit:add-cmd-filters = [[_]{ put $true } [_]{ put $false }]", "echo\n", nil},
+		{"positive", "edit:add-cmd-filters = [[cmd]{ ==s $cmd echo }]", "echo\n", []string{"echo"}},
+		{"negative", "edit:add-cmd-filters = [[cmd]{ ==s $cmd echo }]", "echo x\n", nil},
+	}
```
xiaq, line 55: Please add a test case for the short-circuiting behavior. Such a test will need to set up a callback that returns `$false` and another that does some kind of side effect. To check the side effect, the framework of this table-driven test will be insufficient, so write another test.

pkg/edit/editor.go
```diff
@@ -47,17 +50,21 @@ func NewEditor(tty cli.TTY, ev *eval.Evaler, st store.Store) *Editor {
 	}
 
 	if fuser != nil {
+		ignoreLeadingSpace := eval.NewGoFn("ignore-leading-space",
```
xiaq, line 53: This name does not correspond to an actual symbol, so put it in brackets: `<ignore-cmd-with-leading-space>`

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,36 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
...
 
+func TestEditor_TestBeforeAddCmd(t *testing.T) {
+	cases := []struct {
+		name string
+		rc   string
+		in   string
+		exp  []string
+	}{
+		{"nonCallableItem", "edit:add-cmd-filters = [$false]", "echo\n", []string{"echo"}},
+		{"noOutputItem", "edit:add-cmd-filters = [[_]{}]", "echo\n", []string{"echo"}},
```
xiaq, line 48: Since these will pollute the terminal when run, I'd rather you comment them out now and reenable them later.

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,36 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
 	testCommands(t, f.Store /* no commands */)
 }
 
+func TestEditor_TestBeforeAddCmd(t *testing.T) {
+	cases := []struct {
+		name string
+		rc   string
+		in   string
+		exp  []string
+	}{
```
xiaq, line 46: Can you write out the tests more sparsely and using keys? That is the style used by other tests and are much easier to read:

```
{
  name: "",
  rc: "...",
  input: "...",
  wantHistory: ...
}
```

pkg/edit/editor_test.go
```diff
@@ -37,6 +37,36 @@ func TestEditor_DoesCommandWithLeadingSpaceToHistory(t *testing.T) {
 	testCommands(t, f.Store /* no commands */)
 }
 
+func TestEditor_TestBeforeAddCmd(t *testing.T) {
+	cases := []struct {
+		name string
+		rc   string
+		in   string
```
xiaq, line 44: Call this `input` instead? `in` is quite ambiguous.

pkg/edit/editor_test.go
```diff
@@ -27,14 +27,102 @@ func TestEditor_DoesNotAddEmptyCommandToHistory(t *testing.T) {
...
 
-	testCommands(t, f.Store /* no commands */)
+			feedInput(f.TTYCtrl, c.input)
+			f.Wait()
+
+			testCommands(t, f.Store, c.wantHistory...)
+		})
+	}
+
+	t.Run("short-circuiting behavior", func(t *testing.T) {
```
xiaq, line 107: Make this a separate top-level function.

pkg/edit/editor_test.go
```diff
@@ -27,14 +27,102 @@ func TestEditor_DoesNotAddEmptyCommandToHistory(t *testing.T) {
...
+		c := `
+edit:add-cmd-filters = [
+    [_]{ put $false }
+    [_]{ edit:add-cmd-filters = []; put $true }
```
xiaq, line 111: I'd rather you test the side effect in a more straightforward way, for example by setting a variable, and test the value of the variable using the `testGlobal` helper.

pkg/edit/editor_test.go
```diff
@@ -27,14 +27,102 @@ func TestEditor_DoesNotAddEmptyCommandToHistory(t *testing.T) {
...
+			feedInput(f.TTYCtrl, c.input)
+			f.Wait()
+
+			testCommands(t, f.Store, c.wantHistory...)
+		})
+	}
+
+	t.Run("short-circuiting behavior", func(t *testing.T) {
+		c := `
+edit:add-cmd-filters = [
```
xiaq, line 109: Can these fit on the same line? You can also make it slightly shorter by writing `@edit:add-cmd-filters = ...`.

pkg/edit/config_api.go
```diff
@@ -49,6 +49,16 @@ func initAfterReadline(appSpec *cli.AppSpec, ev *eval.Evaler, ns eval.Ns) {
...
 }
 
+//elvdoc:var add-cmd-filters
+//
+// List of filters to run before adding a command to history.
+//
+// A filter is a function that takes a command as argument and outputs
+// a boolean value. If any of the filters outputs `$false`, the
+// command is not saved to history, and the rest of the filters are
+// not run. Default contains a filter which ignores command starts
```
xiaq, line 59: > The default value of this list ...

pkg/edit/editor_test.go
```diff
@@ -27,14 +27,102 @@ func TestEditor_DoesNotAddEmptyCommandToHistory(t *testing.T) {
...
+    [_]{ put $false }
+    [_]{ edit:add-cmd-filters = []; put $true }
+]
+`
+		f := setup(rc(c))
```
xiaq, line 114: BTW, `rc` takes multiple commands, so you can write

```go
f := setup(rc(
  "x = ''",
  "@edit:add-cmd-filters = ..."))
```

Assuming `$x` is the variable used to test side effect.

xiaq:
Thank you. I'd name the test function differently but I'll do that later. Merging.

PR #896, opened 2020-01-13
pkg/eval: Add commands append and concat

dvdkon:
This adds two basic functions for working with lists. Elvish already allows both operations by special syntax, but it's often useful to have them as functions.

### DIFF
--- a/pkg/cli/lscolors/feature_string.go
+++ b/pkg/cli/lscolors/feature_string.go
@@ -4,6 +4,30 @@ package lscolors
 
 import "strconv"
 
+func _() {
+	// An "invalid array index" compiler error signifies that the constant values have changed.
+	// Re-run the stringer command to generate them again.
+	var x [1]struct{}
+	_ = x[featureInvalid-0]
+	_ = x[featureOrphanedSymlink-1]
+	_ = x[featureSymlink-2]
+	_ = x[featureMultiHardLink-3]
+	_ = x[featureNamedPipe-4]
+	_ = x[featureSocket-5]
+	_ = x[featureDoor-6]
+	_ = x[featureBlockDevice-7]
+	_ = x[featureCharDevice-8]
+	_ = x[featureWorldWritableStickyDirectory-9]
+	_ = x[featureWorldWritableDirectory-10]
+	_ = x[featureStickyDirectory-11]
+	_ = x[featureDirectory-12]
+	_ = x[featureCapability-13]
+	_ = x[featureSetuid-14]
+	_ = x[featureSetgid-15]
+	_ = x[featureExecutable-16]
+	_ = x[featureRegular-17]
+}
+
 const _feature_name = "featureInvalidfeatureOrphanedSymlinkfeatureSymlinkfeatureMultiHardLinkfeatureNamedPipefeatureSocketfeatureDoorfeatureBlockDevicefeatureCharDevicefeatureWorldWritableStickyDirectoryfeatureWorldWritableDirectoryfeatureStickyDirectoryfeatureDirectoryfeatureCapabilityfeatureSetuidfeatureSetgidfeatureExecutablefeatureRegular"
 
 var _feature_index = [...]uint16{0, 14, 36, 50, 70, 86, 99, 110, 128, 145, 180, 209, 231, 247, 264, 277, 290, 307, 321}
--- a/pkg/eval/builtin_fn_container.go
+++ b/pkg/eval/builtin_fn_container.go
@@ -28,9 +28,11 @@ func init() {
 		"has-key":   hasKey,
 		"has-value": hasValue,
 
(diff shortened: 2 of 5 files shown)


xiaq:
Both commands make a lot of sense, but I am not sure how well they fit into Elvish's paradigm yet:

* As a command, `append $li $elem` is not shorter than `put [$@li $elem]`, and as an expression, `(append $li $elem)` is longer than `[$@li $elem]`. The latter is not very efficient now as it involves a lot of unnecessary conversions, but that can be optimized to be a `Cons` call internally.

* Similarly, writing `concat $li1 $li2` is not shorter than `put [$@li1 $@li2]`, and `(concat $li1 $li2)` is longer than `[$@li1 $@li2]`. Also, Elvish has an internal concept of [concat](https://github.com/elves/elvish/blob/304d270d5af0a5539978f1de66a5a063080c5d3a/pkg/eval/vals/concat.go#L34) that is not the same as list concatenation.

dvdkon:
I initially implemented `append` and `concat` as functions in Elvish, because the built-in expressions are awkward when the "parameters" aren't variables. For example, `traversed = (concat $traversed (list-files $dir))` can't be expressed directly with the syntax you showed. The functions could also be used for higher-order functions, but that's probably more niche.

I picked the name `concat` because it's used in several other languages, but I'm definitely not opposed to other names.

xiaq:
Aha, this kind of use case discussion is exactly what I hoped to have.

Elvish uses multiple values instead of lists in many places. In particular, an idiomatic Elvish function that produces multiple values should just output multiple values, instead of a list. In your example `traversed = (concat $traversed (list-files $dir))`, the function `list-files` is outputting a list of filenames in a directory. If it is rewritten to output all filenames individually, there is no need for a `concat` function, and this piece of code can be written as `@traversed = $@traversed (list-files $dir)`.

zzamboni:
"an idiomatic Elvish function that produces multiple values should just output multiple values, instead of a list" - this is great advise, and I had not seen it in writing before - it explains a lot about the behavior of the Elvish builtins. Maybe this could be added to https://elv.sh/learn/effective-elvish.html#working-with-multiple-values?

krader1961:
Am I misunderstanding this assertion:

> For example, traversed = (concat $traversed (list-files $dir)) can't be expressed directly with the syntax you showed.

It works for me as expected using external commands. For example,

```
~> t = [-2 -1 0]
~> t = [$@t (e:seq 3)]
~> put $t
▶ [-2 -1 0 1 2 3]
```

As @xiaq pointed out an elvish function should behave similarly by `put`ing individual values rather than constructing, then `put`ing, a list. Obviously a function which does emit a list will result in nested lists. Which is presumably the impetus for this PR. For example,

```
~> t = [1 2]
~> fn lf { put [a b] }
~> lf
▶ [a b]
~> put [$@t (lf)]
▶ [1 2 [a b]]
```

So while I can appreciate why @dvdkon felt the need for this change, and I might have proposed something like it when I first started using elvish, I don't think it should be merged. If you have to deal with a function that emits a list rather than a sequence of values it is probably better to explicitly use an intermediate var and the `$@var` explosion syntax.

dvdkon:
I've looked at [the script](https://gitlab.com/dvdkon/dotfiles/-/blob/e879b3fe0b79675dc8285124475a8d8ab0ae7f9f/sus/.bin/sus) that was my reason for writing this PR, and while some of my use of an `append` function can be replaced by directly using Elvish syntax, there are still some uses that (as far as I know) can't.

For example, `(append $conf[sync-items] $path)`. I still think there are use-cases for `append` and `concat` functions, there's a point when special syntax gets too confusing.

krader1961:
Hmmm, the first use of `append` in that module is this: `xs2 = (append $xs2 $x)`. Which is really just `xs2 = [$@xs2 $x]`. The second use, and the example in your comment, can be written as `conf[sync-items] = (dedup [(explode conf[sync-items]) $path])`. I'll grant you that isn't quite as obvious as using your proposed `append` command but neither is it particularly hard to read or understand.

On the other hand, your `concat` command is more interesting. Consider the unit test you wrote:

```
That(`explode (concat [1 2] [3 4])`).Puts("1", "2", "3", "4"),
```

That's also trivial to do using `explode` or the `$@` syntax. **However**, if it were generalized to support merging two lists **and appending a singleton to a list** you wouldn't need `append`. It seems to me `concat` is easier to justify and more useful if generalized.

xiaq:
Revisiting this: the PR itself is reasonable, but I feel Elvish needs own idiomatic way of doing functional programming that works with the rest of the language - especially pipelines and multiple values - and it's better tackled systematically later.

So I am going to close this for now, sorry. Before Elvish gets its functional programming idioms, feel free to implement these functions in Elvish itself.

PR #903, opened 2020-01-22
Appends pathSeparator to dir name

TacoVox:
> * The old quotedPathSeparator was renamed to pathSeparator
> * pathSeparator is not put inside quotes before it is used
> * Instead of using the pathSeparator in quotes a suffix, it is appended
>   to the full directory name and afterwords wrapped into quotes iff
>   necessary

Not quite sure that is what you intended, but it seems to work for me just fine :)

Looking forward to a small code review from you.

### DIFF
--- a/pkg/edit/complete/complete_test.go
+++ b/pkg/edit/complete/complete_test.go
@@ -109,9 +109,8 @@ func TestComplete(t *testing.T) {
 		},
 	}
 
-	pathSep := parse.Quote(string(os.PathSeparator))
 	allFileNameItems := []completion.Item{
-		fc("a.exe", " "), fc("d", pathSep), fc("non-exe", " "),
+		fc("a.exe", " "), fc("d" + string(os.PathSeparator), ""), fc("non-exe", " "),
 	}
 
 	allCommandItems := []completion.Item{
@@ -244,7 +243,7 @@ func TestComplete(t *testing.T) {
 			panic(err)
 		}
 		allLocalCommandItems := []completion.Item{
-			fc("./a.exe", " "), fc("./d", "/"), fc("./d2", "/"),
+			fc("./a.exe", " "), fc("./d/", ""), fc("./d2/", ""),
 		}
 		tt.Test(t, tt.Fn("Complete", Complete), tt.Table{
 			// Filename completion treats symlink to directories as directories.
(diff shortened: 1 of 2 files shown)


TacoVox:
Will take care of the tests tmrw. 😎

xiaq:
Hi! Still going to fix the unit tests? :)

TacoVox:
Hi @xiaq,
sorry for my delay here. Quite busy at work these days. Fixed all except one. Will try to attend it as soon as possible. Haven't forgotten about this one here ;)

Thanks for your patience!

TacoVox:
Managed to quickly fix it before work! Thanks for your patience, @xiaq. 🌵

pkg/edit/complete/generators.go
```diff
@@ -123,19 +122,19 @@ func generateFileNames(seed string, onlyExecutable bool) ([]RawItem, error) {
...
 				// Symlink to directory.
-				suffix = quotedPathSeparator
+				full += pathSeparator
 			}
 		}
 
 		items = append(items, ComplexItem{
-			Stem: full, CodeSuffix: suffix,
+			Stem:         full,
+			CodeSuffix:   " ",
```
xiaq: `CodeSuffix: " "` should only apply for non-directories. For directories, don't append a space; this is to allow the user to start another completion to complete the files within the directory.

pkg/edit/complete/complete_test.go
```diff
@@ -280,10 +279,10 @@ func cb(s string) CodeBuffer { return CodeBuffer{s, len(s)} }
...
 func fc(s, suffix string) completion.Item {
-	return completion.Item{ToShow: s, ToInsert: s + suffix,
+	return completion.Item{ToShow: s, ToInsert: parse.Quote(s) + suffix,
 		ShowStyle: ui.StyleFromSGR(lscolors.GetColorist().GetStyle(s))}
 }
 
 func r(i, j int) diag.Ranging { return diag.Ranging{From: i, To: j} }
 
-func withPathSeparator(d string) string { return d + parse.Quote(string(os.PathSeparator)) }
+func withPathSeparator(d string) string { return d + string(os.PathSeparator) }
```
xiaq: `withPathSeparator(d)` is not any shorter than `d + pathSeparator`. I'd prefer that you remove this function.

pkg/edit/complete/generators.go
TacoVox: Good point!

pkg/edit/complete/complete_test.go
TacoVox: Agree ;)

PR #906, opened 2020-01-26
Make the power operator take multiple args

hanche:
This change makes the `^` operator take multiple arguments, in a right associative manner, as is appropriate for exponentiation. It can also take 0 arguments (result: 1.0) or one argument (returning the argument).

I altered the documentation part too, but cheated a bit, replacing `(float64 256.0)` by `256.0` in order to stay consistent with the rest. But arguably, everything should be changed to reflect the new behaviour.

I added a test, but couldn't test the test, as I haven't figured that part out.

### DIFF
--- a/pkg/eval/builtin_fn_num.go
+++ b/pkg/eval/builtin_fn_num.go
@@ -11,17 +11,18 @@ import (
 
 // TODO(xiaq): Document float64.
 
-//elvdoc:fn + - * /
+//elvdoc:fn + - * / ^
 //
 // ```elvish
 // + $summand...
 // - $minuend $subtrahend...
 // * $factor...
 // / $dividend $divisor...
+// ^ $number
 // ```
 //
-// Basic arithmetic operations of adding, subtraction, multiplication and division
-// respectively.
+// Basic arithmetic operations of adding, subtraction, multiplication, division
+// and exponentiation respectively.
 //
 // All of them can take multiple arguments:
 //
@@ -34,12 +35,18 @@ import (
 // ▶ 70
 // ~> / 2 5 7 # 2 / 5 / 7
 // ▶ 0.05714285714285715
+// ~> ^ 2 3
+// ▶ 8
+// ~> ^ 2 0.5
+// ▶ 1.4142135623730951
+// ~> ^ 2 2 3
+// ▶ 256
 // ```
 //
 // When given one element, they all output their sole argument (given that it is a
 // valid number). When given no argument,
 //
-// -   `+` outputs 0, and `*` outputs 1. You can think that they both have a
+// -   `+` outputs 0, and `*` and `^` outputs 1. You can think that they have a
 // "hidden" argument of 0 or 1, which does not alter their behaviors (in
 // mathematical terms, 0 and 1 are
 // [identity elements](https://en.wikipedia.org/wiki/Identity_element) of
@@ -49,6 +56,9 @@ import (
 //
 // -   `/` becomes a synonym for `cd /`, due to the implicit cd feature. (The
 // implicit cd feature will probably change to avoid this oddity).
+//
+// -    `^` is right associative, so that `^ 2 2 3` is equivalent to
+// `^ 2 (^ 2 3)`, not to `^ (^2 2) 3`.
 
 //elvdoc:fn %
 //
@@ -62,21 +72,6 @@ import (
 // ```elvish-transcript
 // ~> % 23 7
 // ▶ 2
-// ```
-
-//elvdoc:fn ^
-//
-// ```elvish
-// ^ $base $exponent
-// ```
-//
-// Output the result of raising `$base` to the power of `$exponent`. Examples:
-//
-// ```elvish-transcript
-// ~> ^ 2 10
-// ▶ 1024
-// ~> ^ 2 0.5
-// ▶ 1.4142135623730951
 // ```
 
 //elvdoc:fn &lt; &lt;= == != &gt; &gt;=
@@ -169,7 +164,7 @@ func init() {
 		"-": minus,
 		"*": times,
 		"/": slash,
-		"^": math.Pow,
+		"^": power,
 		"%": mod,
 
 		// Random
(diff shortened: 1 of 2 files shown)


xiaq:
I am not actually sure it is common to have this operation; and when the user does have the need to write something along this line, `(^ a (^ b c))` is probably much clearer in what it does than `(^ a b c)`.

In particular, I am not aware of any Lisp with a variadic `expt` function, so it seems that this is indeed not particularly useful. See [R6RS](http://www.r6rs.org/final/html/r6rs/r6rs-Z-H-14.html#node_sec_11.7.4.3), [Racket](https://docs.racket-lang.org/reference/generic-numbers.html#%28def._%28%28quote._~23~25kernel%29._expt%29%29), [Common Lisp](https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/node127.html) and [Clojure](https://github.com/clojure/math.numeric-tower/blob/08d50ecec92c9aad8d13041013d5c0f2a28bb895/src/main/clojure/clojure/math/numeric_tower.clj#L80).

hanche:
@xiaq Well, I admit it may not be a _super_ convenience, and it had not occurred to me to check all those languages that you refer to. My inspiration mostly came from Haskell, which of course does infix arithmetic, so it's not directly comparable. But there, exponentiation is automatically right associative, without needing parentheses; so `a^b^c` is the same as `a^(b^c)`, not to `(a^b)^c`. The latter is better computed as `a^(b*c)` anyhow. And, I thought it would be nice to put exponentiation on a more equal footing with addition and multiplication as the natural next step the hierarchy, if you will. I still like the patch, but I won't be terribly disappointed if you don't want it.

krader1961:
FWIW, I agree with @xiaq. That Haskell is right-associative for the expression `a ^ b ^ c` seems arbitrary to me. If I ever saw such an expression I would have to consult the language specification to be certain I understood its behavior (i.e., the order or precedence). I just checked the behavior of the `calc` command and it too is right-associative for this case.  This interpretation might be common among mathematicians, even when explicit typographic hints aren't present, but it's not clear it belongs in Elvish. Or any other programming language that isn't like (APL)[https://en.wikipedia.org/wiki/APL_(programming_language)]. It seems to me this is sufficiently obscure that explicitness (e.g., `(^ a (^ b c))` is preferable to adding a "here be dragons" feature.

P.S., This is a good example of a proposed change that would have benefited from a discussion before a code review was requested. Sometimes a code review is warranted even before there is consensus that the change is a good idea. Typically because a proof of concept implementation is needed to provide a concrete basis for evaluating the idea. That doesn't seem to be the case in this instance. Not to mention most enhancements I see proposed in this, and other, projects that start with a pull-request rather than an enhancement issue. 😺

hanche:
First, if you think the right associativity of `a ^ b ^ c` is arbitrary, it's only because you haven't thought about it. That's okay – most people never needed to know why this is the only reasonable default (if you wish to allow the expression at all). Nobody should ever compute `(a ^ b) ^ c`, as `a ^ (b * c)` is more efficient and doesn't overflow as easily.

That said, I can see more easily now why this would be confusing to some users, though I expect few would ever stumble across the problem in the first place. So I am perfectly happy to let the matter drop.

This situation is new to me. I have a button below called Close and comment. Will that withdraw the pull request, so it no longer appears on the elvish project page? Or would @xiaq have to formally reject it on his side? How does this work?

Now on to the PS from @krader1961 : I am not sure I follow. To me, a pull request is not a _request_ for a code review, it is (in this case, where I am not addressing an existing item in the issue tracker) a feature request with attached patch to implement it. That includes an _invitation_ to a code review, but only if the feature request meets with approval. Why should you waste time on reviewing code for a feature you don't want? I apologise if you wasted time on reviewing the code. (Though the patch is so small, and so trivial, I trust not much time was wasted in this case.)

Would it have been better if I just posted this as an issue? The main reason I did not do so, is because I figured it would likely be rejected, not because the feature was unwanted, but because it is not of sufficient interest to warrant the work of implementing it. Though I suppose I could make it clear that I already have the code and would send a pull request if the issue met with approval.

I am not a seasoned github user, so would gratefully accept advice on the proper way to do things. I suppose this would very between projects, though.

krader1961:
> Would it have been better if I just posted this as an issue?

Yes. :smile: Note that I did not spend anytime reviewing your code change. Because your proposed change had not been discussed. Let alone a consensus that the change should be made. I looked at the P.R. description and had a "nak" reaction. :frown:

This is why I wish people would open an issue for any change that isn't trivial (e.g., fixing a typo). Such proposals are better served by an issue where the merits of the proposed change are discussed. Also, when I'm searching for whether something has been discussed I normally only search "issues" and exclude "pull requests". Yes, Github (not to mention a generic Google search) blurs this distinction. But note that Github chooses a reasonable default, searching just issues or pull-requests, when clicking the search icon in each view. That's because software engineers treat items found in the "issue" category different from code review (i.e., look at my pull-request) category.

> if you think the right associativity of a ^ b ^ c is arbitrary, it's only because you haven't thought about it.

I took A.P. calculus my senior year of high-school back when dinosaurs roamed the earth. However, I admit my use of maths more difficult than algebra is rare these days. 😢  My point is I understand how exponentiation works. My, poorly expressed point above, is that anyone who isn't a mathematician will find the behavior of this syntax to be arbitrary and requiring a careful read of the documentation. Only a mathematician will find your change useful and immediately understandable. It seems like the change will be used and appreciated by a tiny fraction of elvish users. On the order of one in a thousand.

The fact we are even having this discussion about the merits of the change, and not the change itself, is why I made my previous comment. I'm not picking on you, @hanche, as this is something I see happen far too often in open source projects.

hanche:
Okay @krader1961 , thanks for the lessons on github culture.

And, since this did not meet with approval, I am closing this pull request. (I am not sure if, or where, it will remain as a historical document. But I guess I am about to find out.

xiaq:
Thanks to you both, @hanche and @krader1961. I like it when the community runs itself, although I have a few words to add.

Personally I don't mind the form in which I receive feedbacks or proposals, and code is a perfectly valid form, so what you did here is completely acceptable. What needs to be understood is that features are not automatically accepted, so PRs can get rejected because I have other ideas, or simply haven't really thought about the topic because I am unsure.

If you enjoy directly hacking the codebase and talking in terms of code (and don't mind having the PR rejected!), there is nothing wrong with continuing to do this. Your code on its own is good and you have documented what it does very clearly. However, if you do start to be annoyed by me always picking on design questions, that is a sign that you should discuss the feature first :)

Issue #897, opened 2020-01-16
cd auto-completion quoting issue on Windows hosts


TacoVox:
Let me start with thanking you for this wonderful project. I really enjoy it! And it works like a charm on my macOS machine.

---

However, I am forced to use MS Windows 10 at work. As I want to use `elvish` for my daily work, I encountered an issue when using the auto-completion on the press of `TAB`.

To visualize the problem, let me show you how a path builds up using the feature:
```sh
~ > cd
[TAB]
~ > cd go'\'
[TAB]
~ > cd 'go\pkg''\'
```

Needless to say, the user would expect that to look like:
```sh
~ > cd 'go\pkg\'
```

I believe to have identified some lines that are partly causing the issue:

`pkg/edit/complete/raw_item.go:41+`
```go
func (c ComplexItem) Cook(q parse.PrimaryType) completion.Item {
	quoted, _ := parse.QuoteAs(c.Stem, q)

	return completion.Item{
		ToInsert:  quoted + c.CodeSuffix, // <- a simple concatenation
		ToShow:    c.Stem + c.DisplaySuffix,
		ShowStyle: c.DisplayStyle,
	}
}
```

Due to the concatenation, we have these multiple single cuoted expressions after one another.

As I see it, one could argue that whenever there is such a case, all single-quoted expressions should  be combined into a single one. Howevever, as I just started looking at the code, I have not yet a complete overview and this may be a bad idea.

@xiaq what do you think? How could this issue be effectively avoided? I would be glad to help out.

---

Thanks for your work. Much appreciated!

TacoVox:
To follow up on my issue:

What I mean is an additional function in the `parse` package that is capable of combining `n` quoted expressions into a single one. Of course it would need to feature logic to check, if we need to quote with single quotes or double quotes.

Is there any case where such a function would cause harm? Otherwise the marked line above could look something along the lines of:
```go
func (c ComplexItem) Cook(q parse.PrimaryType) completion.Item {
	[...]
	return completion.Item{
		[...]
		ToInsert: parse.CombineQuotes(quoted, c.CodeSuffix),
		[...]
	}
}
```

xiaq:
Rather than combining quotes, a simpler approach is to make it possible to attach a code suffix before quoting.

xiaq:
[...]

Issue #583, opened 2018-02-02
Reclaim backslash as a metacharacter


xiaq:
Right now, backslashes in Elvish are literal; it was so to make it easier to type paths in Windows.

However, too many times have I pasted a shell script from the Internet and have to change the backslash at the end of the line to a backquote.

Besides the immediate motivation of line continuation, we can do other interesting things with backslashes. One is traditional escaping, like using `\[` for a literal `[`. Another is C-like escaping, like `\n` for a newline.

After those, we still have a lot of free characters. For example, I am thinking of allowing arbitrary quoting delimiters after `\q`, like:

```elvish-transcript
~> echo \q(both " and ')
both " and '
~> echo \q[both " and ']
both " and '
```

xiaq:
Backslash is now a metacharacter again, although it is now only used in line continuations.

PR #911, opened 2020-02-13
Windows pathsep

Xjs:
Relating to #897, in my daily use I have found that completing Windows paths with slashes is usually better than with escaped backslashes, even if not entirely correct. This is due to the slash being a character that needs not be escaped. Completion with a character that is only valid in quotes does not feel good in everyday usage, at least to me -- even if it works.

This PR fixes several completion uses I had when using Elvish on Windows 10 (as well as a typo in pkg/edit/builtins.go, for which I couldn't be bothered to open an extra PR). 

Let me know what you think!

Further enhancements could include the quoted string concatenation function mentioned in #897, and using it to complete backslashy paths as-is, but still use slashes as default for relative navigation, or making the behaviour configurable.

### DIFF
--- a/pkg/edit/builtins.go
+++ b/pkg/edit/builtins.go
@@ -184,7 +184,7 @@ var bufferBuiltinsData = map[string]func(*cli.CodeBuffer){
 	"move-dot-right":            makeMove(moveDotRight),
 	"move-dot-left-word":        makeMove(moveDotLeftWord),
 	"move-dot-right-word":       makeMove(moveDotRightWord),
-	"move-dot-left-small-word":  makeMove(moveDotLeftWord),
+	"move-dot-left-small-word":  makeMove(moveDotLeftSmallWord),
 	"move-dot-right-small-word": makeMove(moveDotRightSmallWord),
 	"move-dot-left-alnum-word":  makeMove(moveDotLeftAlnumWord),
 	"move-dot-right-alnum-word": makeMove(moveDotRightAlnumWord),
--- a/pkg/edit/complete/complete_test.go
+++ b/pkg/edit/complete/complete_test.go
@@ -109,7 +109,7 @@ func TestComplete(t *testing.T) {
 		},
 	}
 
-	pathSep := parse.Quote(string(os.PathSeparator))
+	pathSep := parse.Quote("/")
 	allFileNameItems := []completion.Item{
 		fc("a.exe", " "), fc("d", pathSep), fc("non-exe", " "),
 	}
@@ -286,4 +286,4 @@ func fc(s, suffix string) completion.Item {
 
 func r(i, j int) diag.Ranging { return diag.Ranging{From: i, To: j} }
 
-func withPathSeparator(d string) string { return d + parse.Quote(string(os.PathSeparator)) }
+func withPathSeparator(d string) string { return d + parse.Quote("/") }
--- a/pkg/edit/complete/generators.go
+++ b/pkg/edit/complete/generators.go
@@ -116,7 +116,7 @@ func generateFileNames(seed string, onlyExecutable bool) ([]RawItem, error) {
 		}
 		// Only accept searchable directories and executable files if
 		// executableOnly is true.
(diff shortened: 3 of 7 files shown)


xiaq:
#897 was fixed by #903.

I know that Windows supports forward slashes as a path separator on the kernel API level, but not all commands work well with it. Keeping the use of backslashes is a better choice.

xiaq:
Thanks for noting the typo, I fixed it in d623dafd35958be23d0a4f92118f6c657058b535.

I see that you made a change to how files are determined to be executable but I'm not sure why you made it - care to elaborate?

Xjs:
Thanks for fixing the typo.

As for the slashes/backslashes thing, I agree that using backslashes is technically the better choice, but I find mandatorily having to use quoted strings for file paths is a usability regression. I would be happy if using slashes or backslashes were a configurable option, even at compile time (which it de facto is, I can of course patch it myself and will probably continue doing so for the time being). If you cannot be swayed at all, could we maybe discuss #583 again? I don't have a really good solution though, just the complaint that I feel the path separator is so frequently used it should not be a metacharacter.

(Sorry for being so fussy about that. I really like Elvish, and I'm glad that I found the project, and I'm really thankful for all the time and work you put in this so far -- and I'm really enthusiastic about using it daily and contributing. But obviously I'm very demanding regarding the UX of my daily used tools. That's why I would be really happy if we could reach a consensus here :))

As for the executable change, I noticed that tab-completion for external commands didn't work on Windows, and found out that the check for executable files checks the FileMode in the infos returned from ReadDir. However, on Windows file systems (NTFS here), files generally don't have an executable bit and so the check for all executable bits will always fail. This was a very quick'n'dirty fix since I couldn't immediately find what heuristic could be used to determine if a file is executable on Windows.

I think theoretically, every file could be executable, but the syscall will complain if the file doesn't contain an executable header or a script... but it could be that certain file name extensions would be a good heuristic as well. I would need to research more to be 100% certain. The fix I made condenses all the executable bit checks in one place, though, which would it easier to iterate on these kinds of heuristics. (NB: Checking for mode & 0111 on UNIX, i. e. whether every executable bit is set, is too broad a heuristic as well. A file I own could have mode 700 and still be executable for me. So there's room for improvement here as well I'd say.)

SitiSchu:
Regarding the slash conversion, a system like bash on msys2 would be nice where the shell uses forward slashes but converts them to backslashes for commands, this would make the experience a lot more consistent across OSs

SitiSchu:
Detecting executables on windows should probably be done using the `PATHEXT` variable, it's used to determine which fine extensions are executable: https://superuser.com/questions/1027078/what-is-the-default-value-of-the-pathext-environment-variable-for-windows

krader1961:
Regarding forward versus back slash I present for your amusement:

```
krader@WINDOWS10 C:\Users\krade\.ssh>ssh-keygen
Generating public/private rsa key pair.
Enter file in which to save the key (C:\Users\krade/.ssh/id_rsa):
```

The `ssh-keygen` command I ran in a Windows 10 command shell mixes the two types of path separators. And that is the official Microsoft port of the OpenSSH command. Even Microsoft can't get this right.

> Regarding the slash conversion, a system like bash on msys2 would be nice where the shell uses forward slashes but converts them to backslashes for commands, this would make the experience a lot more consistent across OSs

Presumably that is needed so that something like `e:a_command /a/path` won't cause `a_command` to construe `/a/path` as a switch rather than a path had it been written `e:a_command '\a\path'`.  The question is how to do the conversion only on arguments that are paths rather than switches. In other words if the path `/x` is valid and the user runs `e:a_command /x` how does Elvish know `/x` is not a switch and the forward slashes should be converted to back slashes?

SitiSchu:
At least the built-in commands by Microsoft all seem to support using - for switches so you don't have to use a slash.

xiaq:
Revisiting this: My preferred way to go now is to make `\` a valid bareword character again, see #979.

PR #913, opened 2020-02-13
Initial crude attempt at adding a built-in variable which returns the runtime OS

zzamboni:
**NOT FOR MERGE YET** - this is meant for discussion.

This originated from some discussion in Telegram about how it would be useful sometimes to easily identify the OS in which Elvish is running (e.g. to have functions which execute different commands depending on it).

Some quick exploration revealted that  this information is already available in the `runtime` module. This PR exposes this value as a new built-in variable `$os`.

Feedback of any kind welcome.

### DIFF
--- a/pkg/eval/builtin_ns.go
+++ b/pkg/eval/builtin_ns.go
@@ -3,6 +3,7 @@ package eval
 import (
 	"strconv"
 	"syscall"
+        "runtime"
 
 	"github.com/elves/elvish/pkg/eval/vars"
 	"github.com/xiaq/persistent/vector"
@@ -87,6 +88,7 @@ var builtinNs = Ns{
 	"nil":   vars.NewReadOnly(nil),
 	"true":  vars.NewReadOnly(true),
 	"false": vars.NewReadOnly(false),
+        "os":    vars.NewReadOnly(runtime.GOOS),
 	"paths": &EnvList{envName: "PATH"},
 	"pwd":   PwdVariable{},
 	"args":  vars.NewReadOnly(vector.Empty),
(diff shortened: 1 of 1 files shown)


krader1961:
`gofmt`? Other than the style nitpick I'd like to see this, or something like it, merged. One of the first things I do in my fish/bash/ksh startup scripts is export an `OS_NAME` var; e.g., `set -gx OS_NAME (uname -s | string lower)`. That way I can use the same shell "dot" files on all of my platforms and easily condition initializations based on the platform OS. For example,

```
if test "$OS_NAME" = openbsd
    set -g MANPATH /usr/share/man /usr/local/share/doc/posix/
end
```

krader1961:
Also, if we're going to do something like this it might be useful to also expose GOARCH at the same time. While the platform architecture is (almost? probably?) irrelevant for an elvish program I would argue that it should also be exposed for consistency if for no other reason.

zzamboni:
(fixed formatting)

I agree with you, there are probably a few other internal parameters that could be easily exposed. Maybe a new namespace, something like `sys` would make sense, so that we could have `$sys:os-name`, `$sys:arch`, etc.

xiaq:
Yes, go with a new `sys` module. I have wanted to introduce it for a while but never got around to doing it.

zzamboni:
@xiaq as I'm not sure I have the go skills nor the time to work on this, I have created a new issue #916 so that this does not get forgotten.

krader1961:
Note that I've created PR #927 to implement the `sys:` namespace and implement the two vars discussed here.

xiaq:
Closing since @krader1961 opened #927.

PR #922, opened 2020-02-23
Mathematical Improvements

enty8080:
Added more mathematic operations.


krader1961:
I'm not a fan of blindly adding every math function provided by libc; or, in this case, the standard Go runtime. A lot of those functions aren't useful for a shell like elvish and will never be used. Some of them can't even be used without a shim function; e.g., `math.Float32frombits()`.

```
~/projects/3rd-party/elvish> math:float32frombits 234
Exception: wrong type of 1'th argument: wrong type: need !!uint32, got string
[tty], line 1: math:float32frombits 234
```

Also, all of the functions which are ultimately included require at least some basic unit tests.

@enty8080 I suggest changing this PR to focus on a small number of math functions that pretty much everyone would agree could be useful and that accept a single `float64` argument; e.g., `trunc`, `sin`, `cos`, `tan`, `log` and `log2`. Add those to the dispatch table and also include appropriate tests in the *pkg/eval/module_math_test.go* file.

Then tackle more complex functions that might be useful. Such as `math.Min()`. Although that probably requires some discussion since the Go version only takes two values and it might be better if elvish had a wrapper to support an arbitrary, greater than one, number of values so you can write `max = (math:max 3  7 5)` rather than `max = (math:max 3 (math:max 7 5))`.

krader1961:
FWIW: As part of adding documentation of the `math:` module I am also adding support for functions such `log` (the natural logarithm function).

krader1961:
@enty8080 See https://github.com/elves/elvish/pull/925/commits/73bc68975281eb025146c9c163f06b644a10a633 for an example of how to add new math functions that have appropriate unit tests. That you did not document the functions you added is understandable since the `math:` namespace you augmented did not have any documentation. However, the lack of unit tests for the math commands you introduced is not acceptable. If you had written unit tests for each math command you added you would have noticed that many of them are not usable.

xiaq:
Rejecting in favor of #937.

PR #925, opened 2020-02-27
Document math module

krader1961:
This doesn't document `math:round` because I intend to introduce `math:round-to-even` and document both commands at the same time. But I felt like this PR was already complex enough that more changes would simply slow down the review.

I also make no claims that the documentation is optimal. This is just a first step which simply documents the availability of the functions in the `math:` namespace.


krader1961:
Note, this makes PR #922 obsolete and thus should be rejected.

krader1961:
@enty8080 Take a look at this PR for an example of how to add new math functions. Each change (other than the first because it reorganizes the code) is small and includes unit tests and documentation.

pkg/eval/vals/equal.go
```diff
@@ -22,6 +23,12 @@ func Equal(x, y interface{}) bool {
 	case bool:
 		return x == y
 	case float64:
+		if math.IsNaN(x) && math.IsNaN(y.(float64)) {
+			// `NaN == NaN` is always false since NaN is never equal to any
+			// other value; not even NaN. But for unit tests we want to ensure
+			// that if the test is expected to produce NaN it does so.
+			return true
```
krader1961: Ugh! I just realized this changes the behavior of `eq (float64 nan) (float64 nan)` but not `== (float64 nan) (float64 nan)`. The `eq` documentation says it checks if the values `are structurally equivalent` which isn't rigorously defined. Ideally this redefinition of NaN equality would only apply to the unit test framework and not affect `eq`. Nonetheless, it's not clear that the current behavior is a problem. In fact, it might be considered a feature; albeit one that needs to be documented.

krader1961:
In light of review feedback on PR #928 I see that more changes are needed to this PR. I'm going to close this PR, rebase on master, fold #930 into the branch, make the needed improvements, and open a new PR. Sorry, @xiaq, if you're in the middle of making comments about this PR.

PR #927, opened 2020-02-28
Add platform: module to expose platform data

### LINKED ISSUE #916: New `sys` namespace
As proposed in #913, it would be nice to have a new `sys` module that contains system-specific variables (and maybe functions?), such as:

- `$sys:os-name`
- `$sys:arch`
- etc.

krader1961:
Resolves #916

### DIFF
--- a/pkg/eval/platform/platform.go
+++ b/pkg/eval/platform/platform.go
@@ -1,5 +1,5 @@
-// Package `os` exposes variables and functions that deal with the specific
-// platform being run on; such as the OS name and CPU architecture.
+// Package `platform` exposes variables and functions that deal with the
+// specific platform being run on; such as the OS name and CPU architecture.
 package platform
 
 import (
@@ -11,14 +11,14 @@ import (
 
 //elvdoc:var arch
 //
-// The architecture of the platform; e.g., amd64, arm, ppc.
+// The architecture of the platform; e.g. amd64, arm, ppc.
 // This corresponds to Go's
 // [`GOARCH`](https://pkg.go.dev/runtime?tab=doc#pkg-constants) constant.
 // This is read-only.
 
 //elvdoc:var os
 //
-// The name of the operating system; e.g., darwin (macOS), linux, etc.
+// The name of the operating system; e.g. darwin (macOS), linux, etc.
 // This corresponds to Go's
 // [`GOOS`](https://pkg.go.dev/runtime?tab=doc#pkg-constants) constant.
 // This is read-only.
--- a/pkg/eval/platform/platform.go
+++ b/pkg/eval/platform/platform.go
@@ -1,0 +1,29 @@
+// Package `os` exposes variables and functions that deal with the specific
+// platform being run on; such as the OS name and CPU architecture.
+package platform
+
+import (
+	"runtime"
+
+	"github.com/elves/elvish/pkg/eval"
+	"github.com/elves/elvish/pkg/eval/vars"
+)
+
+//elvdoc:var arch
+//
+// The architecture of the platform; e.g., amd64, arm, ppc.
+// This corresponds to Go's
+// [`GOARCH`](https://pkg.go.dev/runtime?tab=doc#pkg-constants) constant.
+// This is read-only.
+
+//elvdoc:var os
+//
+// The name of the operating system; e.g., darwin (macOS), linux, etc.
+// This corresponds to Go's
+// [`GOOS`](https://pkg.go.dev/runtime?tab=doc#pkg-constants) constant.
+// This is read-only.
+
+var Ns = eval.Ns{
+	"arch": vars.NewReadOnly(runtime.GOARCH),
+	"os": vars.NewReadOnly(runtime.GOOS),
+}
--- a/pkg/eval/platform/platform_test.go
+++ b/pkg/eval/platform/platform_test.go
@@ -9,7 +9,7 @@ import (
 
 var That = eval.That
 
-func TestOs(t *testing.T) {
+func TestPlatform(t *testing.T) {
 	setup := func(ev *eval.Evaler) { ev.Builtin.AddNs("platform", Ns) }
 	eval.TestWithSetup(t, setup,
 		That(`put $platform:arch`).Puts(runtime.GOARCH),
--- a/pkg/eval/platform/platform_test.go
+++ b/pkg/eval/platform/platform_test.go
@@ -1,0 +1,18 @@
+package platform
+
+import (
+	"runtime"
+	"testing"
+
+	"github.com/elves/elvish/pkg/eval"
+)
+
+var That = eval.That
+
+func TestOs(t *testing.T) {
+	setup := func(ev *eval.Evaler) { ev.Builtin.AddNs("platform", Ns) }
+	eval.TestWithSetup(t, setup,
+		That(`put $platform:arch`).Puts(runtime.GOARCH),
+		That(`put $platform:os`).Puts(runtime.GOOS),
+	)
+}
--- a/website/ref/index.toml
+++ b/website/ref/index.toml
@@ -11,6 +11,10 @@ name = "edit"
 title = "The Editor Module (edit:)"
 
 [[articles]]
+name = "platform"
+title = "Access to the platform’s identifying data (platform:)"
+
+[[articles]]
 name = "re"
 title = "The Regular Expression Module (re:)"
 
--- a/website/ref/index.toml
+++ b/website/ref/index.toml
@@ -12,7 +12,7 @@ title = "The Editor Module (edit:)"
 
 [[articles]]
 name = "platform"
-title = "Access to the platform’s identifying data (platform:)"
+title = "Access to the platform's identifying data (platform:)"
 
 [[articles]]
 name = "re"
--- a/website/ref/platform.md
+++ b/website/ref/platform.md
@@ -1,0 +1,7 @@
+<!-- toc -->
+
+# Introduction
+
+The `platform:` module provides access to the platform’s identifying data.
+
+@elvdoc -ns platform: -dir ../pkg/eval/platform
--- a/website/ref/platform.md
+++ b/website/ref/platform.md
@@ -2,6 +2,6 @@
 
 # Introduction
 
-The `platform:` module provides access to the platform’s identifying data.
+The `platform:` module provides access to the platform's identifying data.
 
 @elvdoc -ns platform: -dir ../pkg/eval/platform
--- a/pkg/program/shell/runtime.go
+++ b/pkg/program/shell/runtime.go
@@ -15,6 +15,7 @@ import (
 	"github.com/elves/elvish/pkg/eval/re"
 	"github.com/elves/elvish/pkg/eval/store"
 	"github.com/elves/elvish/pkg/eval/str"
+	"github.com/elves/elvish/pkg/eval/platform"
 	daemonp "github.com/elves/elvish/pkg/program/daemon"
 )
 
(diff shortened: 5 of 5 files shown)


pkg/eval/sys/sys.go
```diff
@@ -0,0 +1,25 @@
...
+
+//elvdoc:var platform
+//
+// The name of the platform; e.g., darwin (macOS), linux. This is read-only.
+
+var Ns = eval.NewNs()
+
+func init() {
+	Ns.Add("arch", vars.NewReadOnly(runtime.GOARCH))
+	Ns.Add("platform", vars.NewReadOnly(runtime.GOOS))
```
xiaq: I imagine you called it after Python's `sys.platform`, but I'd rather this be called `os` to align with Go.

pkg/eval/sys/sys_test.go
```diff
@@ -0,0 +1,17 @@
+package sys
+
+import (
+	"runtime"
+	"testing"
+
+	"github.com/elves/elvish/pkg/eval"
+)
+
+func TestMath(t *testing.T) {
```
xiaq: TestSys

pkg/eval/sys/sys.go
```diff
@@ -0,0 +1,25 @@
...
+package sys
+
+import (
+	"runtime"
+
+	"github.com/elves/elvish/pkg/eval"
+	"github.com/elves/elvish/pkg/eval/vars"
+)
+
+//elvdoc:var arch
```
xiaq: It's worth documenting which constants in the Go standard library these correspond to by linking to https://pkg.go.dev/runtime?tab=doc#pkg-constants for both constants.

pkg/eval/sys/sys.go
```diff
@@ -0,0 +1,25 @@
...
+//
+// The architecture of the platform; e.g., amd64, arm, ppc. This is read-only.
+
+//elvdoc:var platform
+//
+// The name of the platform; e.g., darwin (macOS), linux. This is read-only.
+
+var Ns = eval.NewNs()
```
xiaq: The following is not succinct:

```go
var Ns = eval.Ns{
  "arch": ...,
  ...
}
```

website/ref/sys.md
```diff
@@ -0,0 +1,10 @@
+<!-- toc -->
+
+# Introduction
+
+The `sys:` module provides platform specific functions and constants.
```
xiaq: I know you put in the part about functions for consistency with other docs, but this module doesn't have any functions yet, so it's better to leave them out.

pkg/eval/sys/sys.go
```diff
@@ -0,0 +1,25 @@
+// Package sys exposes variables and functions that deal with the specific
+// system being run on. Such as the OS name and CPU architecture.
```
xiaq: Join the two sentences: ", such as ..."

pkg/eval/sys/sys_test.go
```diff
@@ -0,0 +1,17 @@
...
+	setup := func(ev *eval.Evaler) { ev.Builtin.AddNs("sys", Ns) }
+	That := eval.That
```
xiaq: This is really a nitpick, but I'd declare `That` as a global variable to be consistent with other test files:

pkg/eval/sys/sys.go
```diff
@@ -0,0 +1,25 @@
+// Package sys exposes variables and functions that deal with the specific
```
xiaq: I am considering putting these variables related to interpreter internals in this module too:

https://github.com/elves/elvish/blob/7868b68ff34c7b0884b6f6e6520ee24b38f32b93/pkg/eval/eval.go#L148-L162

Which won't fit this description something along the description [Python's sys module](https://docs.python.org/3/library/sys.html) would be more accurate:

> This module provides access to some variables used or maintained by the interpreter and to functions that interact strongly with the interpreter.

pkg/eval/sys/sys.go
krader1961: I did but on second thought agree with using `os` for consistency given how tightly Elvish is bound to Go.

krader1961: I also assume you'll want the namespace changed from `sys` to `os`.

pkg/eval/sys/sys_test.go
krader1961: Stupid copy/paste. Bad, bad copy/paste dog :smile:

pkg/eval/sys/sys.go
krader1961: I assume you meant "more succinct". Fair enough. Done. Although, as a grey beard this is a situation where I think the use of `init()` is preferable so that when functions are added in the future only the `eval.NewNS()` is changed to include an invocation of the `AddGoFns()` method. I hope you're not suggesting the chaining two or more mutating methods in this situation is good style. If the overall sequence was very short I could live with it; and, in fact, have written such code.   The problem is that style tends to devolve into a "bike shedding" discussion which is never good.

website/ref/sys.md
krader1961: It seemed to me that functions like `os.Geteuid()` were likely to be exposed in the near future but I see your point. There is no reason to make people think this elvish namespace provides functions until it actually does os.

pkg/eval/sys/sys.go
krader1961: Done.

pkg/eval/sys/sys_test.go
krader1961: Fair enough. I copied that pattern from *./pkg/eval/re/re_test.go*. I'll create another PR to fix that inconsistency. :smile:

krader1961: FWIW, It appears that *./pkg/eval/re/re_test.go* is the only other unit test which employs this pattern. All of the other unit test files which use `That()` are in the `eval` package and thus do not need an alias.

pkg/eval/sys/sys.go
krader1961: And the straightforward rename exposes a problem: 

```
pkg/program/shell/runtime.go:18:2: os redeclared as imported package name
        previous declaration at pkg/program/shell/runtime.go:7:2
```

Which means that any import of the elvish os module must rename it. Which means we now need to establish a convention for this situation. 😦 I have tentatively settled on importing `elvish_os "github.com/elves/elvish/pkg/eval/os"`.

pkg/eval/sys/sys_test.go
krader1961: The `That` function reference can't be made a global in the unit tests outside the `eval` package, AFAICT.

pkg/eval/sys/sys.go
krader1961: My preference is for anything that is explicitly tied to the OS or architecture, and which cannot be papered over by an OS agnostic API, be in the `os` namespace rather than in the builtin namespace.

krader1961:
@xiaq Please take another look.

pkg/eval/sys/sys.go
xiaq: My preference is calling this module `sys`; it is a bit awkward to have a variable called `$os:os`, no?

xiaq:
I posted a comment, but actually I think we need to think harder on how to name this module and what its scope should be.

I was considering modelling the module after Python's `sys` (and commented that I preferred it), or Go's `runtime`, but after thinking about it more, I think both make the mistake of mixing at least two kinds of things:

* Things important to the runtime environment itself: Python has `sys.path`; Go has `runtime.GOROOT`.
* Things related to the low-level environment: Python has `sys.platform`, Go has `runtime.GOOS`.

I think it's wise to separate both things, and the variables in this PR definitely belong to the *second* bullet. Your proposed name of `os` comes close to describing it, but I'm not enthusiastic about the name. Not because of the clash with Go's `os`, but because `$os:os` is too awkward.

Maybe `platform` is a good choice? Variable names `$platform:os` and `$platform:arch` make a lot of sense. It is slightly too long, though.

(I am inspired by Python's [platform](https://docs.python.org/3.9/library/platform.html) module, although the Python module makes the mistake of including information about the interpreter.)

It's probably also worth researching how other languages call such modules.

pkg/eval/sys/sys.go
krader1961: Note that my first incarnation did call this the `sys:` namespace :smile: Not sure what you mean by your `$os:os` example. There is no such var at this time. There is `$os:platform` and `$os:arch`. I deliberately used `platform` because `os` has too much baggage. For example, why is the value `darwin` rather than `macos` on my Macs? Or `linux` rather than `ubuntu`? Etcetera. I think `platform` better conveys the meaning of this var.

krader1961:
> I think we need to think harder on how to name this module and what its scope should be.

Okay, but since it seems like we're still a ways away from a elvish 1.0 release it seems like doing something reasonable, if not ideal, is better than not doing anything and forcing people to use the `uname` command. As people use it and provide feedback we can refactor as appropriate.

> Maybe platform is a good choice? Variable names $platform:os and $platform:arch make a lot of sense. It is slightly too long, though.

I'm not bothered by the length of those two example var names. However, I was thinking that builtins like `chmod` and `umask` which cannot be implemented in an OS agnostic fashion would be placed in the `os:` namespace. That, in fact, is what Python [does](https://docs.python.org/3.9/library/os.html). Perhaps we should start with a `platform:` namespace that exposes basic facts (e.g., the CPU architecture) then add an `os:` namespace for OS specific capabilities like `umask`.

xiaq:
@krader1961 I agree with what you said, so let's rename the module in this PR to platform and we are good to go.

krader1961:
@xiaq, Please take another look.

pkg/eval/platform/platform.go
```diff
@@ -0,0 +1,29 @@
+// Package `os` exposes variables and functions that deal with the specific
```
xiaq: Update this comment.

pkg/eval/platform/platform.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"runtime"
+
+	"github.com/elves/elvish/pkg/eval"
+	"github.com/elves/elvish/pkg/eval/vars"
+)
+
+//elvdoc:var arch
+//
+// The architecture of the platform; e.g., amd64, arm, ppc.
```
xiaq: Remove the comma after `e.g.`.

pkg/eval/platform/platform.go
```diff
@@ -0,0 +1,29 @@
...
+// This corresponds to Go's
+// [`GOARCH`](https://pkg.go.dev/runtime?tab=doc#pkg-constants) constant.
+// This is read-only.
+
+//elvdoc:var os
+//
+// The name of the operating system; e.g., darwin (macOS), linux, etc.
```
xiaq: Same here, remove the comma after `e.g.`.

pkg/eval/platform/platform_test.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"runtime"
+	"testing"
+
+	"github.com/elves/elvish/pkg/eval"
+)
+
+var That = eval.That
+
+func TestOs(t *testing.T) {
```
xiaq: TestPlatform

website/ref/index.toml
```diff
@@ -10,6 +10,10 @@ title = "The Builtin Module"
 name = "edit"
 title = "The Editor Module (edit:)"
 
+[[articles]]
+name = "platform"
+title = "Access to the platform’s identifying data (platform:)"
```
xiaq: We do not use the typographically correct apostrophe `’` elsewhere in the website, so stick to `'`.

website/ref/platform.md
```diff
@@ -0,0 +1,7 @@
+<!-- toc -->
+
+# Introduction
+
+The `platform:` module provides access to the platform’s identifying data.
```
xiaq: Same about the apostrophe.

pkg/eval/platform/platform.go
krader1961: Blame my elementary school teachers :-) They taught me that both e.g. and i.e. should be followed by a comma. Googling "should e.g. have a comma" turns up a lot of conflicting advice with most sources saying it's okay either way. But, happy to oblige as this is a bike shedding issue :-)

xiaq: Interesting, I never knew that.

PR #928, opened 2020-02-28
Document the str: module and add unit tests

krader1961:


pkg/eval/str/str.go
```diff
@@ -8,6 +8,316 @@ import (
 	"github.com/elves/elvish/pkg/eval"
 )
 
+//elvdoc:fn compare
+//
+// ```elvish
+// str:compare $a $b
+// ```
+//
+// Compare returns an integer comparing two strings lexicographically. The
```
xiaq: * Go's convention of using the symbol name (`compare`) here unfortunately doesn't work well for Elvish; In Go all exported symbols are capitalized and can be used as the first word of an English sentence. In Elvish all symbols are (for now at least) and they are not capitalized.

  Existing docs start the sentence with the verb, so just start the sentence with "Returns" here (but see bullets below on this particular one).

  Same elsewhere.

* Elvish does not have return values, say "outputs" instead of "returns".

  Same elsewhere.

* I know you copied Go's doc here, but the original wording can use some improvement. An "integer comparing two strings" is not quite intuitive. I'd rather this say:

  > Compares two strings and outputs an integer representing the lexicographical ordering of the two strings. The result...

pkg/eval/str/str.go
```diff
@@ -8,6 +8,316 @@ import (
...
+// result will be 0 if a==b, -1 if a < b, and +1 if a > b.
+//
+// ```elvish-transcript
+// ~> str:compare a a
+// > 0
```
xiaq: Please stick to the default value output marker ▶.

pkg/eval/str/str.go
```diff
@@ -8,6 +8,316 @@ import (
...
+// > 1
+// ```
+
+//elvdoc:fn contains
+//
+// ```elvish
+// str:contains $str $substr
+// ```
+//
+// Contains returns `$true` if `$substr` is within `$str` else `$false`.
```
xiaq: This is kind of hard to parse. I know you are trying to be clear what this returns, but I like how Go's wording "reports whether" makes it clear that this is a predicate, so I'd say stick to the original wording. Same elsewhere.

pkg/eval/str/str.go
```diff
@@ -8,6 +8,316 @@ import (
...
+// > $false
+// ```
+
+//elvdoc:fn has-prefix
+//
+// ```elvish
+// str:has-prefix $str $prefix
+// ```
+//
+// Has-prefix tests whether the string `$str` begins with `$prefix`
```
xiaq: Missing period.

pkg/eval/str/str.go
```diff
@@ -8,6 +8,316 @@ import (
...
+// > $false
+// ```
+
+//elvdoc:fn has-suffix
+//
+// ```elvish
+// str:has-suffix $str $suffix
+// ```
+//
+// Has-suffix tests whether the string `$str` begins with `$suffix`
```
xiaq: Missing period.

pkg/eval/str/str_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"testing"
+
+	"github.com/elves/elvish/pkg/eval"
+)
+
+func TestStr(t *testing.T) {
+	setup := func(ev *eval.Evaler) { ev.Builtin.AddNs("str", Ns) }
+	That := eval.That
```
xiaq: Make it a global variable.

pkg/eval/str/str.go
krader1961: Done

pkg/eval/str/str.go
krader1961: Done.

pkg/eval/str/str.go
krader1961: Done, but in an elvish fashion.

pkg/eval/str/str.go
krader1961: Fixed

pkg/eval/str/str.go
krader1961: Fixed

pkg/eval/str/str_test.go
krader1961: Done

krader1961:
@xiaq Please take another look.

pkg/eval/str/str.go
```diff
@@ -8,6 +8,313 @@ import (
 	"github.com/elves/elvish/pkg/eval"
 )
 
+//elvdoc:fn compare
+//
+// ```elvish
+// str:compare $a $b
+// ```
+//
+// Compare two strings and output an integer that will be 0 if a == b,
```
xiaq: Please use the verb form with "s", as if there is an implicit subject "this function". I have documented this in https://github.com/elves/elvish/blob/master/CONTRIBUTING.md#documenting-builtin-functions-and-variables.

pkg/eval/str/str.go
```diff
@@ -8,6 +8,313 @@ import (
...
+// ▶ 1
+// ```
+
+//elvdoc:fn contains
+//
+// ```elvish
+// str:contains $str $substr
+// ```
+//
+// If `$substr` is within `$str` output `$true` else `$false`.
```
xiaq: This still feels a bit too verbose to document a predicate. I find the following much easier to read:

Outputs whether `$substr` is within `$str`.

Also, "within" is a bit vague; I'd change that to:

Outputs whether `$str` contains `$substr` as a substring.

This also matches the order of the arguments.

Same for other predicates below.

pkg/eval/str/str.go
```diff
@@ -8,6 +8,313 @@ import (
...
+
+//elvdoc:fn index
+//
+// ```elvish
+// str:index $str $substr
+// ```
+//
+// Output the index of the first instance of `$substr` in `$str`, or -1
+// if `$substr` is not present in `$str`.
+
```
xiaq: Missing //. The elvdoc for a function or variable must be a continuous comment block.

Issue #208, opened 2016-09-11
Allow inspecting cause of exceptions


xiaq:
Errors are now opaque values. One should be able to inspect them by indexing into relevant fields. For instance, the following may be used to retrieve the exit code of an external command:

```
try {
  ls
} except e {
  println "exited with "$e[exit-code]
}
```

hanche:
*Related* – perhaps it should have been a separate issue, but here goes:

I'd like to be able to re-throw an exception after capturing it, either via the `?(…)` construct or the `except` block of a `try` statement.

Suggestion:
- If `$e` is a caught exception, `fail $e` should re-throw it.
- however, `fail $ok` should do nothing at all, both in order to save the user a conditional and in order to avoid dealing with the dreaded “no error” error.

xiaq:
A comprehensive design is not there yet, but at least it should also be possible to access the message inside an exception raised by `fail` (see #986).

xiaq:
Exposing the traceback is a bit hairy, so for the sake of this issue I'll focus on just the cause. Exposing the traceback is tracked in #1042.

xiaq:
Design notes:

* Add predicates in the `exc:` for determining the type of the cause, e.g. `exc:is-external-cmd-exit` (may want to rename this; it's too long). Internal structured errors such as [`errs.OutOfRange`](https://github.com/elves/elvish/blob/8af76083827a4d968d5f35ace2e89dc4dcc573e6/pkg/eval/errs/errs.go#L9) can have similar predicates.

* Add fields for the data contained in the exception, e.g. `$e[exit-code]` for the numerical exit code (only present if `exc:is-external-cmd-exit $e` is true).

There is an open question regarding how to expose the many methods of Go's [`WaitStatus`](https://godoc.org/syscall#WaitStatus). There are several options:

* Expose them as fields of the exception too.

* Expose them as functions.

* Expose them as methods; Elvish doesn't have the concept of methods today, so this will be blocked on Elvish having methods in the first place.

hanche:
I think `exc:is-exit-code` will work. Only external commands produce exit codes anyway; there is no need to spell it out. Also `exc:is-signal`. For the latter, it would be good to get both the signal number and the name.

xiaq:
After researching a bit more on how to name this: "exit" and "exit code" actually have narrower meanings than I thought (at least on the POSIX C API level). A process *exits* only when it terminates on its own. If it was killed by a signal, it never exited, and there is no exit code, only a signal number.

[...]

PR #932, opened 2020-03-05
Document the exception data type and `cause` attr

krader1961:
This also explicitly documents `float64` as a data type, although that
is a byproduct of this change. I created this change because someone
asked on the project's Gitter channel how to extract useful info from an
exception object.

Related #208

### DIFF
--- a/website/ref/builtin.md
+++ b/website/ref/builtin.md
@@ -96,20 +96,34 @@ but when the current directory is empty, `count` will wait for inputs. Hence it
 is required to put the input in a list: `count [*]` unambiguously supplies input
 in the argument, even if there is no file.
 
-## Numerical Commands
+## Commands That Operate On Numbers
 
-Commands that operate on numbers are quite flexible about the format of the
-numbers. Integers can be specified as decimals (e.g. `233`) or hexadecimals
-(e.g. `0xE9`) and floating-point numbers can be specified using the scientific
-notation (e.g. `2.33e2`). These are different strings, but equal when considered
-as commands.
+Commands that operate on numbers are quite flexible about the
+format of those numbers. See the discussion of the [number data
+type](./language.html#number).
 
-Elvish has no special syntax or data type for numbers. Instead, they are just
-strings. For this reason, builtin commands for strings and numbers are
-completely separate. For instance, the numerical equality command is `==`, while
-the string equality command is `==s`. Another example is the `+` builtin, which
-only operates on numbers and does not function as a string concatenation
-commands.
+Because numbers are normally specified as strings, rather than as
+an explicit `float64` data type, some builtin commands have variants
+intended to operate on strings or numbers exclusively. For instance, the
+numerical equality command is `==`, while the string equality command is
+`==s`. Another example is the `+` builtin, which only operates on numbers
+and does not function as a string concatenation command. Consider these
+examples:
+
+```elvish-transcript
+~/projects/3rd-party/elvish> + x 1
+Exception: wrong type of 1'th argument: cannot parse as number: x
+[tty], line 1: + x 1
+~> + inf 1
+▶ (float64 +Inf)
+~> + -inf 1
+▶ (float64 -Inf)
+~> + -infinity 1
+▶ (float64 -Inf)
+~> + -infinityx 1
+Exception: wrong type of 1'th argument: cannot parse as number: -infinityx
+[tty], line 1: + -infinityx 1
+```
 
 ## Predicates
 
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -46,12 +46,14 @@ statements **executes** to produce side effects; an expression **evaluates** to
 some values. (The traditional terms for the two levels are "commands" and
 "words", but those terms are quite ambiguous.)
 
-# String
+# Data types
+
+## String
 
 The most common data structure in shells is the string. String literals can be
 quoted or unquoted (barewords).
 
-## Quoted
+### Quoted
 
 There are two types of quoted strings in Elvish, single-quoted strings and
 double-quoted strings.
@@ -72,7 +74,7 @@ effect, simply concatenate strings: instead of `"my name is $name"`, write
 `"my name is "$name`. Under the hood this is a
 [compound expression](#compound-expression-and-braced-lists).
 
-## Barewords
+### Barewords
 
 If a string only consists of bareword characters, it can be written without any
 quote; this is called a **bareword**. Examples are `a.txt`, `long-bareword`, and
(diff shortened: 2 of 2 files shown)


krader1961:
I noticed the "Numerical Commands" section of the builtins reference document was confusing and could be improved by, in part, referencing the new numeric data type section I added earlier. So I've force-pushed an update that does so.

website/ref/language.md
```diff
@@ -106,16 +106,75 @@ error.
...
 
-Elvish does have a dedicated `float64` number type, but it does not have a
-literal syntax; instead it needs to be constructed using a function call, as in
-`put (float64 42)`; the argument `42` is a string. For convenience, all Elvish
-functions that expect number arguments also accept strings that can be parsed as
-a number, so explicit number constructors are not needed often.
+# Numeric data type
 
+Elvish has a numeric data type, `float64`, created using this syntax:
+`(float64 numeric-literal)`. For convenience, all Elvish functions that
```
xiaq: It's inaccurate to call the argument `numeric-literal`. The concept of "literal" is syntactical: this is implying that the argument must be something *literal*, as opposed to, say, a variable.

I prefer using a variable as a placeholder here, so just say `(float64 $number)`.

website/ref/language.md
```diff
@@ -106,16 +106,75 @@ error.
...
+expect a numeric argument will accept a string that can be parsed as a
+number. They will automatically convert the string to a `float64`. Which
+means explicit `float64` constructors are almost never needed. Not even
+for special values like infinity (`Inf`) or not-a-number (`NaN`). Note
+that the letter case of these magic values doesn't matter so `inf` and
+`Inf` are equivalent.
 
-# List and Map
+Numeric literals can be specified using
```
xiaq: See above; these are not literals. If you used the plaeceholder `$number` above, just say:

The $number argument can be a string in one of the following formats:

website/ref/language.md
```diff
@@ -106,16 +106,75 @@ error.
...
+
+* decimal notation (e.g., `10`),
+* hexadecimal notation (e.g., `0xA`),
+* octal notation (e.g., `0o12`),
+* binary notation (e.g., `0b1010`),
+* floating point notation (e.g., `10.0`), and
+* scientific notation (e.g., `1.0e1`).
+
+All of those numeric literals are equivalent. However, note that all of
```
xiaq: You have already documented above that strings that can be parsed as numbers are automatically converted to numbers by builtins. Better not to repeat that.

In fact, since this is a convention of the standard builtins, not a property of the number type itself, it is better to

1. Consolidate all the explanations in builtin.md;
2. At the end of this section here, add a link to the builtin.md section as a "tip".

website/ref/language.md
```diff
@@ -106,16 +106,75 @@ error.
...
+
+To convert a `float64` data type to a string do `(to-string $float64_var)`
+or format it: `(printf '%.3f' $float64_var)`.
+
+# Exception data type
+
+Elvish has an exception data type, but it does not have a literal
+syntax for that type (despite the literal representation shown below by
+`put`). Instead it needs to be constructed using a function call, as in
+`(fail 'was unexpected')`. If an external command returns a non-zero exit
```
xiaq: This example should be `?(fail ...)`.

website/ref/language.md
```diff
@@ -46,7 +46,7 @@ statements **executes** to produce side effects; an expression **evaluates** to
 some values. (The traditional terms for the two levels are "commands" and
 "words", but those terms are quite ambiguous.)
 
-# String
+# String data type
```
xiaq: I like it that you added "data type" to the section titles, but it would be even clearer if you make a new section called "data type" and move all the existing sections under it.

website/ref/language.md
krader1961: I decided to just remove this paragraph. Eventually we'll want to document the behavior of `==` and `!=`. That is, whether or not they apply an appropriate epsilon, but that should be done in a separate change after we've decided whether we want the current behavior or for the commands to be friendlier for people who are naive about the pitfalls of comparing floating point quantities.

krader1961:
@xiaq Please take another look. You'll probably want to squash the last two, of the three, commits. Alternatively, I can do so and force push so you only need to press the Github merge button.

website/ref/language.md
krader1961: No, because it shows how to throw/raise an exception; not its representation when shown using `put`. However, I agree the capture parens aren't appropriate in this context so I'll remove them.

website/ref/language.md
```diff
@@ -102,27 +104,79 @@ metacharacters; use quoted strings instead. For instance, to echo a star, write
...
 The three syntaxes above all evaluate to strings, and they are interchangeable.
 For instance, `xyz`, `'xyz'` and `"xyz"` are different syntaxes for the same
-string, and they are always equivalent.
+string, and they are always equivalent with the exception of
+**escape sequences** as documented above.
+
+## Number
+
+Elvish has a numeric data type, `float64`, created using this syntax:
+`(float64 $number)`. For convenience, all Elvish functions that
```
xiaq: > For convenience, all Elvish functions...

Please move the entirety of this discussion to builtin.md and link to it near the end of this section.

website/ref/language.md
```diff
@@ -102,27 +104,79 @@ metacharacters; use quoted strings instead. For instance, to echo a star, write
...
+expect a numeric argument will accept a string that can be parsed as a
+number. They will automatically convert the string to a `float64`. Which
+means explicit `float64` constructors are almost never needed. Not even
```
xiaq: > Not even for special values...

"Not even" sounds unnecessarily surprised; just say "this conversion also includes special values...".

website/ref/language.md
```diff
@@ -102,27 +104,79 @@ metacharacters; use quoted strings instead. For instance, to echo a star, write
...
+for special values like infinity (`Inf`) or not-a-number (`NaN`). Note
+that the letter case of these magic values doesn't matter so `inf` and
```
xiaq: > Note that the letter case of these magic values...

"Magic value" sounds necessarily strong; just say "special value".

website/ref/language.md
```diff
@@ -102,27 +104,79 @@ metacharacters; use quoted strings instead. For instance, to echo a star, write
...
+The `$number` argument can be a string in the following formats:
+
+* decimal notation (e.g., `10`),
+* hexadecimal notation (e.g., `0xA`),
+* octal notation (e.g., `0o12`),
+* binary notation (e.g., `0b1010`),
+* floating point notation (e.g., `10.0`), and
+* scientific notation (e.g., `1.0e1`).
+
+Note that the examples in the above list all express the same value.
```
xiaq: Better say this before the bullet, instead of after the bullet. Say

The `$number` argument can be a string in any of the following formats (the examples all express the same value):

website/ref/language.md
```diff
@@ -102,27 +104,79 @@ metacharacters; use quoted strings instead. For instance, to echo a star, write
...
+
+To convert a `float64` variable to a string do `(to-string $number)`
+which is equivalent to `(printf '%.16f' $number)`.  Alternatively,
```
xiaq: > which is equivalent to...

This part of the explanation is problematic in several ways:

* It is not accurate; the actual printing algorithm is much more complex than that. 
* The `printf` command is not an Elvish builtin, so avoid referring to it in the language doc.
* Users of Elvish should *not* depend on how the number is formatted, so it's better to leave this unspecified.

One property that users *can* depend on is that the string representation can be passed back to `float64` to get the same value. You can document that property instead here.

website/ref/language.md
```diff
@@ -102,27 +104,79 @@ metacharacters; use quoted strings instead. For instance, to echo a star, write
...
-Elvish does have a dedicated `float64` number type, but it does not have a
-literal syntax; instead it needs to be constructed using a function call, as in
-`put (float64 42)`; the argument `42` is a string. For convenience, all Elvish
-functions that expect number arguments also accept strings that can be parsed as
-a number, so explicit number constructors are not needed often.
+## Exception
 
+Elvish has an exception data type, but it does not have a literal
+syntax for that type (despite the literal representation shown
+below by `put`). Instead it needs to be constructed using the
```
xiaq: > Instead it needs to be constructed using the fail command: ...

This is technically correct. However, since this is the *language* doc, the text should focus on the important language construct related to exceptions. In this case, the important language construct is [exception capture](https://elv.sh/ref/language.html#exception-capture); `fail` is *not* a fundamental part of the language, it just happens to be a builtin that always throws an exception.

Also, there is already some doc on exceptions in [another section](https://elv.sh/ref/language.html#exception-and-flow-commands); better not to repeat things.

I suggest moving the exception section to a different PR; the documentation improvements for the number type is much more solid and can be merged soon.

krader1961: Okay, but I think it is important for all data types, including exceptions, to be documented in this major section. What I'll do is turn this minor section into a stub that refers the reader to https://elv.sh/ref/language.html#exception-and-flow-commands for a broader discussion of how exceptions work.

website/ref/builtin.md
```diff
@@ -96,20 +96,50 @@ but when the current directory is empty, `count` will wait for inputs. Hence it
...
+
+Because numbers are normally specified as strings, rather than as
+an explicit `float64` data type, some builtin commands have variants
+intended to operate on strings or numbers exclusively. For instance, the
+numerical equality command is `==`, while the string equality command is
+`==s`. Another example is the `+` builtin, which only operates on numbers
+and does not function as a string concatenation command. Consider these
+examples:
+
+An explicit `float64` data type is created using this syntax: `(float64
```
xiaq: The paragraph and the next paragraph belong to the number type section in language.md; it is the discussion of how builtin commands work that belong to this doc.

Sorry for the back and forth; I hope it's clear now.

krader1961: Gah! Sorry, as you can see from the `... examples:` at line 111 I intended to do that since the examples are now two paragraphs later. Not sure what happened. Please take another look.

xiaq:
I'd reorganize the number data type section, but it's easiest to merge this PR first.

Issue #576, opened 2018-01-22
Pour go's strings package into a "str:" module


xiaq:
In the same spirit of pouring Go's `regexp` package into the `re:` module, let's do the same to the excellent `strings` package.

occivink:
Would `joins` and `splits` be moved to that module?

xiaq:
@occivink Yes, and they will be renamed `str:join` and `str:split`.

krader1961:
Which, if any, elvish commands should automatically coerce their values to strings if they expect that argument to be a string? I'm asking this in the context of PR #928 and PR #934. What does it mean to pass a `(float64 ...)` as an argument to a command that requires a string. Should elvish automatically convert the value or should we require the user to do a `(to-string $value)`? Given the recent introduction of a `float64` data type is it reasonable to require the user to use `to-string` when passing the value of a variable to a command that requires a string?

krader1961:
@xiaq Should builtins like `has-prefix` be marked as deprecated and the reader told to use `str:has-prefix`? Or should the documentation be left as is but with a note that the command is simply an alias for the equivalent `str:` command and the `str:` form is preferred?

krader1961:
Also, if `has-prefix` is simply an alias for `str:has-prefix` shouldn't the former defer to the latter rather than reinvent the wheel? Which reinvention leading to possible incompatible behavior?

zzamboni:
@xiaq has described somewhere (but I can't find it at the moment) his idea for deprecating features over a two-release cycle, I think this is the way this could be handled.

krader1961:
@zzamboni Understood, vis-a-vis the deprecation model. In fact, that's the model I normally use for this situation. So I've created a PR to formally deprecate the `has-prefix` and `has-suffix` builtins and start the clock ticking. Which doesn't necessarily mean they will actually be removed prior to the 0.16 release but it does put people on notice that they need to start using the commands in the `str:` rather than the `builtin:` name space. Note that `str:has-prefix` and `str:has-suffix` was introduced by commit 9d342125f way back on 2018-01-26. Since the `str:` name space seems unlikely to go away it's time to start simplifying the `builtin:` name space by removing such redundancies.

krader1961:
@xiaq, It seems like this can be closed. While we haven't poured the entirety of Go's `string` package into the `str` module we've got all the most useful functions. Other functions can be handled on a case-by-case basis as people find a need a particular function.

xiaq:
Sounds good, closing.

[...]

PR #934, opened 2020-03-06
Implement str:join

krader1961:
This presumes PR #928 has been merged; thus, only the last commit is relevant.

### DIFF
--- a/pkg/eval/str/str.go
+++ b/pkg/eval/str/str.go
@@ -74,6 +74,102 @@ import (
 // > 7
 // ```
 
+//elvdoc:fn equal-fold
+//
+// ```elvish
+// str:equal-fold $str1 $str2
+// ```
+//
+// Equal-fold reports whether `$str1` and `$str2`, interpreted as UTF-8
+// strings, are equal under Unicode case-folding.
+//
+// ```elvish-transcript
+// ~> str:equal-fold ABC abc
+// > $true
+// ~> str:equal-fold abc ab
+// > $false
+// ```
+
+//elvdoc:fn has-prefix
+//
+// ```elvish
+// str:has-prefix $str $prefix
+// ```
+//
+// Has-prefix tests whether the string `$str` begins with `$prefix`
+//
+// ```elvish-transcript
+// ~> str:has-prefix abc ab
+// > $true
+// ~> str:has-prefix abc bc
+// > $false
+// ```
+
+//elvdoc:fn has-suffix
+//
+// ```elvish
+// str:has-suffix $str $suffix
+// ```
+//
+// Has-suffix tests whether the string `$str` begins with `$suffix`
+//
+// ```elvish-transcript
+// ~> str:has-suffix abc ab
+// > $false
+// ~> str:has-suffix abc bc
+// > $true
+// ```
+
+//elvdoc:fn index
+//
+// ```elvish
+// str:index $str $substr
+// ```
+//
+// Index returns the index of the first instance of `$substr` in `$str`, or -1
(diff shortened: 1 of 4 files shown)


zzamboni:
It seems that this should somehow be coalesced with the existing `joins` function, which is implemented here: https://github.com/elves/elvish/blob/master/pkg/eval/builtin_fn_str.go#L286

I'm not sure how I feel about Go library functions that duplicate functionality of existing builtins. Maybe one should just be an alias for the other? @xiaq any thoughts about this?

krader1961:
Note that `joins` is considerably more limited and annoying, IMHO:

```
~/projects/3rd-party/elvish> joins : x (float64 111.222333444555666777888999) y [a '' b] '' c [] d
Exception: want 1 or 2 arguments, got 9
[tty], line 1: joins : x (float64 111.222333444555666777888999) y [a '' b] '' c [] d
~/projects/3rd-party/elvish> put x (float64 111.222333444555666777888999) y [a '' b] '' c [] d | joins :
Exception: join wants string input, got !!float64
~/projects/3rd-party/elvish> e = ?(joins : (float64 2))
~/projects/3rd-party/elvish> joins : [x $e]
Exception: join wants string input, got exception

```

My version handles float64 as well as any other type that implements the elvish `Stringer` interface:

```
~/projects/3rd-party/elvish> str:join : x (float64 111.222333444555666777888999) y [a '' b] '' c [] d
▶ x:111.22233344455567:y:a::b::c:d
~/projects/3rd-party/elvish> e = ?(joins : (float64 2))
~/projects/3rd-party/elvish> str:join : x (float64 111.222333444555666777888999) y [a '' b] '' c [] d $e
▶ 'x:111.22233344455567:y:a::b::c:d:?(fail ''!!float64 cannot be iterated'')'
```

It doesn't, however, read the args from a pipe at the moment --- I'll add that capability:

```
~/projects/3rd-party/elvish> put x y | str:join :
▶ ''
```

krader1961:
I've modified the `str:join` implementation to read from stdin if only a separator string is provided as an argument so you can now do `put a b | str:join :` or `str:join : a b`.

krader1961:
BTW, @xiaq, I recognize that my choice of signature for this function is controversial. I won't be hurt if you argue it should use `type Inputs` and behave like  the `joins` command with regard to how arguments are specified. I actually wrote a version of this function which works that way. Ultimately I think the idiom this establishes is preferable. It can even be applied to `count` in a backward compatible fashion by special casing one argument, if it isn't a list, as a string whose char count should be returned.

krader1961:
@xiaq You should ignore this PR until PR #928 is merged. At which point I'll rebase and open a new PR. I'm going to leave this open for the moment in case anyone wants to comment about the API divergence from established builtins like `joins`.

@zzamboni Note that the `str:` namespace has other functions that mirror existing builtins. For example, `str:has-prefix` mirrors `has-prefix`. Yet they are not explicitly documented as analogs. They are also implemented identically implicitly rather than explicitly. That is, they both directly delegate to the Go `strings.HasPrefix()` function but neither is defined in terms of the other which is a source of potential confusion and bugs.

krader1961:
@zzamboni See also #576 where @xiaq documents the intent to replace builtins such as `joins` with a `str:join`. How best to do so while minimizing transition pain is an open question.

zzamboni:
@krader1961 fair enough - thanks for the pointer! I agree it would be best to have the functions within the corresponding namespace eventually.

krader1961:
A week has gone by and there haven't been any comments. So close this until PR #928 is merged at which time I'll open a new PR.

PR #936, opened 2020-03-08
Print panic message after dumping the stack

### LINKED ISSUE #935: Reorder output of `rescue()` function
A user recently posted a picture of an elvish panic on Gitter. Unfortunately, the picture didn't include the panic reason. Presumably because it scrolled off the top of the terminal. It probably makes more sense for the `println(r)` in `rescue()` to occur after the `print(sys.DumpStack())`. That way, the most important piece of information is more likely to remain visible and be included in problem reports.

Yes, the user should have copy/pasted the panic text rather than taking a picture of the screen. :smile: However, MS Windows users almost always do the latter rather than the former. :cry:

### LINKED ISSUE #953: Redirecting to a closed pipe end should fail without running the command
When attempting to redirect to a closed pipe end, the process has its I/O channel closed instead:

```
> p=(pipe)
> e:ls -l /proc/self/fd/ <$p >$p
total 0
lr-x------ 1 tetsujin tetsujin 64 Mar 30 15:30 0 -> 'pipe:[52855684]'
lrwx------ 1 tetsujin tetsujin 64 Mar 30 15:30 1 -> /dev/pts/1
l-wx------ 1 tetsujin tetsujin 64 Mar 30 15:30 2 -> 'pipe:[52855684]'
lr-x------ 1 tetsujin tetsujin 64 Mar 30 15:30 3 -> /proc/8081/fd
> prclose $p; pwclose $p
> ls -l /proc/self/fd 0<$p 2>$p
total 0
lr-x------ 1 tetsujin tetsujin 64 Mar 30 15:32 0 -> /proc/8089/fd
lrwx------ 1 tetsujin tetsujin 64 Mar 30 15:32 1 -> /dev/pts/1

```
When elvish performs the redirection to/from the pipe, it already knows the file descriptor for that end of the pipe is no longer valid: at that point it should bail out with an exception rather than continue and launch the child process without the requested redirections. For instance, if you attempt to redirect to a file that doesn't exist or can't be created:
```

> ls -l /proc/self/fd </gleendarpf
Exception: failed to open file /gleendarpf: open /gleendarpf: no such file or directory
> ls -l /proc/self/fd >/sprogzermpt
Exception: failed to open file /sprogzermpt: open /sprogzermpt: permission denied
```

krader1961:
Resolves #935

### DIFF
--- a/pkg/program/shell/shell.go
+++ b/pkg/program/shell/shell.go
@@ -61,9 +61,10 @@ func rescue() {
 	r := recover()
 	if r != nil {
 		println()
-		println(r)
 		print(sys.DumpStack())
-		println("\nexecing recovery shell /bin/sh")
+		println()
+		println(r)
+		println("\nExecing recovery shell /bin/sh")
 		syscall.Exec("/bin/sh", []string{"/bin/sh"}, os.Environ())
 	}
 }


SitiSchu:
This actually resolves #935 (I suppose you accidentally typoed)

krader1961:
Gah! Yes. Thx for catching that.

Issue #727, opened 2018-07-23
Pour go's math functions into a `math:` module?


kwshi:
In the spirit of #576, should we also consider doing the same thing with some math functions?

This would benefit some people like me, who would like to open up terminals to do quick calculations without having to also type in an `ipython` each time I want to do so.

kwshi:
It is not at all essential that this be part of the main elvish repository.  Regardless of whether we end up wanting to have `math` be part of the standard elvish library, it is always possible to write an external `math` module.  But it's worth considering, anyway.

kwshi:
Aha.  Sounds like something I'd be interested in doing...

kwshi:
The main motivation behind me suggesting that we include all these libraries is that, in a way, I've kind of come to anticipate Elvish becoming some sort of a Python replacement--it's incredibly concise/terse while still being expressive, and in some ways, it's even out-flexing Python (in particular, closure and functional programming support, maybe?).

Here's a number of things I once would do in Python that I would now much more gladly do in Elvish (that I wouldn't as gladly have done in Bash, because of Bash's language eccentricities):
- Batch-processing data from a website's JSON API
- Write shell scripts to power a Linux status bar, to monitor battery life, or screen brightness
- Do clever batch file renames or directory restructuring

Here are some things I'd like to do in Elvish instead of Python, but can't quite *yet* because Elvish is still young:
- Do quick calculator calculations (scientific functions, i.e. math/numerical libraries)
- Scrape data from websites (alright, this one's a bit aggressive)

Anyway, if Elvish is to become almost a sort of Python replacement, then I would also envision it assembling a "standard library" whose scale is comparable to that of Python's.  But, of course, that may not be everyone's vision of Elvish.  But those are my thoughts, anyway.

krader1961:
Note that a `math:` namespace was introduced by commit 2d8a045cf two months ago. I have created PR #925 and PR #930 to expose most of the Go math functions likely to be used by an Elvish program. More changes of that nature, such as adding the hyperbolic trigonometric functions, are trivial to implement and can be done if there is sufficient demand.

```
~/projects/3rd-party/elvish> put $math:pi
▶ (float64 3.141592653589793)
~/projects/3rd-party/elvish> math:sin $math:pi
▶ (float64 1.2246467991473515e-16)
~/projects/3rd-party/elvish> math:cos $math:pi
▶ (float64 -1)
~/projects/3rd-party/elvish> math:tan $math:pi
[...]

PR #937, opened 2020-03-10
Document and add math functions

krader1961:
In addition to documenting the `math:` module introduced by commit 2d8a045c
two months ago this also:

a) restructures the code to match that of the `re:` and `str:` modules, and

b) adds many more useful constants and functions.

I don't think this completely addresses issue #727 so I'm marking this
as related to, rather than fixing, that issue.

Related #727


pkg/edit/complete/complete_test.go
```diff
@@ -110,7 +110,7 @@ func TestComplete(t *testing.T) {
 	}
 
 	allFileNameItems := []completion.Item{
-		fc("a.exe", " "), fc("d" + string(os.PathSeparator), ""), fc("non-exe", " "),
+		fc("a.exe", " "), fc("d"+string(os.PathSeparator), ""), fc("non-exe", " "),
```
krader1961: This change is the result of `go fmt`. Since it was the only line in the existing code that was reformatted I decided to just include it even though it is not otherwise relevant to this PR.

krader1961:
Ping? @xiaq, If you're not in the middle of commenting on this P.R. I'd like to rebase it on the master branch to resolve conflicts. Not to mention ensuring it builds on all of my VMs including my recently created Windows 10 plus Msys2 environment.

pkg/eval/vals/equal.go
```diff
@@ -22,7 +25,17 @@ func Equal(x, y interface{}) bool {
 	case bool:
 		return x == y
 	case float64:
-		return x == y
+		if math.IsNaN(x) && math.IsNaN(y.(float64)) {
```
xiaq, line 28: I don't like the idea of making floating point equality smart; best to remove this change from the PR.

pkg/program/shell/runtime.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/elves/elvish/pkg/daemon"
 	"github.com/elves/elvish/pkg/eval"
 	daemonmod "github.com/elves/elvish/pkg/eval/daemon"
+	elvish_math "github.com/elves/elvish/pkg/eval/math"
```
xiaq, line 15: Do not use underscores in identifiers. Follow the pattern above and call the import alias `mathmod`.

website/ref/math.md
```diff
@@ -0,0 +1,16 @@
+<!-- toc -->
+
+# Introduction
+
+The `math:` module provides mathematical functions and
+constants. Everyplace a `$float64` is referenced below that data type
```
xiaq, line 6: This explanation is no longer needed. Just add a reference to It now suffices to just refer to https://elv.sh/ref/builtin.html#commands-that-operate-on-numbers at the end of the next paragraph (row 14).

pkg/eval/vals/equal.go
krader1961: Fair enough. This was mostly a Go learning experience and a "what if...". But as the subsequent comment explains this was also an expedient way to test that NaN is output in unit tests. I'll revert this and give more thought to the unit test case.

pkg/program/shell/runtime.go
krader1961: Okay

krader1961: Wouldn't it be more consistent to use snake-case to mimic the var naming convention; i.e., `mathMod`?

website/ref/math.md
krader1961: Done

krader1961:
I addressed the above issues, rebased on master, and replaced `$float64` mentions in the command documentation with `$number`. Note that I've replaced the problematic code in `vals.Equal()` with equivalent code in `testutils.matchOut()`.

pkg/eval/vals/equal.go
```diff
@@ -22,7 +25,11 @@ func Equal(x, y interface{}) bool {
 	case bool:
 		return x == y
 	case float64:
-		return x == y
+		if math.Signbit(x) == math.Signbit(y.(float64)) &&
```
xiaq, line 28: Let's just stick to Go's standard `==` semantics here.

krader1961: We can't because your code does not stick to Go's standard `==` semantics. The code I added is compatible with Go's standard `==` semantics for infinities while still supporting the +/- epsilon of your code. Without this change if I replace your

```
return math.Abs(x-y.(float64)) <= float64EqualityThreshold
```

with

```
return x == y.(float64)
```

then my unit test passes but this failure occurs:

```
ok      github.com/elves/elvish/pkg/eval/daemon 0.300s
--- FAIL: TestMath (0.09s)
    --- FAIL: TestMath/math:sin_3.141592653589793 (0.00s)
        testutils.go:143: got value out [1.2246467991473515e-16], want [0]
    --- FAIL: TestMath/math:tan_1 (0.00s)
        testutils.go:143: got value out [1.557407724654902], want [1.5574077246549023]
    --- FAIL: TestMath/math:tan_3.141592653589793 (0.00s)
        testutils.go:143: got value out [-1.2246467991473515e-16], want [0]
```

pkg/eval/math/math.go
```diff
@@ -0,0 +1,288 @@
...
+//
+// ```elvish-transcript
+// ~> math:trunc -1.1
+// ▶ (float64 -1)
+// ~> math:trunc 2.5
+// ▶ (float64 2)
+// ```
+
+// Ns is the namespace for the math: module.
+var Ns = eval.NewNs().AddGoFns("math:", fns)
```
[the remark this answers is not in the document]
krader1961: I like, and use, method chaining in a lot of Python code I deal with. I am not a fan of lengthy method chaining. It tends to obscure what is happening. Which makes it harder to understand the code. I appreciate that at the moment the alternative formulation you propose is easy to understand. The problem is what happens when additional constants, methods, etc., are chained together. Which will almost certainly occur. You've objected to the use of `init()` functions before but I don't understand why since, AFAICT, they don't introduce any performance or correctness problems when used in this manner. The mixing of `eval.Ns{...}` declarations with `eval.NewNs()`

Nonetheless, to facilitate getting this merged I've made the suggested change.

P.S., There are just four uses of `NewNs()`. Compared to twenty-four uses of `Ns{}`. Should the former be converted to the latter form?

pkg/eval/vals/equal.go
krader1961: /facepalm :smile: I'm the one who added the epsilon test. Okay, I'll revert that portion of the change.

pkg/eval/math/math.go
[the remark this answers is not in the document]
xiaq: **Re use of method chaining**:

The problem with `init` is that it does not detect cyclic dependencies in initialization. Compare the following two pieces of code:

Using `init`:

```go
// In file a.go
var a int
func init() { a = b + 1 }

// In file b.go
var b int
func init() { b = a + 1}
```

Not using `init`:

```go
// In file a.go
var a int = b + 1

// In file b.go
var b int = a + 1 
```

Go will correctly point out that the second version has an initialization loop, but it will happily compile the first version. Now, what should the values of `a` and `b` be? :)

It is often a matter of trade-off, of course. Sometimes using `init` is much cleaner than the alternative that avoids it. But in this case, my call is that method chaining is the lesser evil here.

**Re use of NewNs**: Yes, they can be all replaced by `Ns{}`.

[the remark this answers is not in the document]
krader1961: I understand your point about an indeterminate initialization sequence. That is inherent in something like Go's module level `init()` mechanism. I was not aware that the compiler would error out if that happens in an example such as you showed. I'll explore that feature. Nonetheless, I still maintain that using the `init()` mechanism is clearer in this situation for non-trivial initialization and, when used in a stylized fashion with an understanding of the possible indeterminacy, is easier to understand.

krader1961:
Sorry about the squashed force push. But I wanted to resolve the conflict with the recent changes to the master branch. Despite that this update only addresses the two most recent issues:

1) Restoring the behavior of `== (float64 $x) (float64 $y)` to use only Go's built-in equality without applying an epsilon. That logic has been moved into `matchOut()` in the *testutils.go* module.

2) Removing the use of `init()` in *math.go*.

pkg/eval/math/math_test.go
```diff
@@ -0,0 +1,104 @@
...
+		That(`math:round-to-even (float64 NaN)`).Puts(math.NaN()),
+
+		That(`math:trunc 2.1`).Puts(2.0),
+		That(`math:trunc -2.1`).Puts(-2.0),
+		That(`math:trunc 2.5`).Puts(2.0),
+		That(`math:trunc -2.5`).Puts(-2.0),
+		That(`math:trunc (float64 Inf)`).Puts(math.Inf(1)),
+		That(`math:trunc (float64 NaN)`).Puts(math.NaN()),
+
+		// That(`math:log 2.71828182845904523536028747135266249775724709369995`).Puts(1.0),
```
xiaq, line 70: Do not commit commented code; just remove it.

pkg/eval/math/math_test.go
```diff
@@ -0,0 +1,104 @@
...
+		That(fmt.Sprintf(`math:log %g`, math.E)).Puts(1.0),
```
xiaq, line 71: Use `math:log $math:e`.

Same for the test cases that use `math.Pi`.

pkg/eval/vals/equal.go
```diff
@@ -4,6 +4,8 @@ import (
 	"reflect"
 )
 
+const float64EqualityThreshold = 1e-9
```
xiaq, line 7: Remove

website/ref/math.md
```diff
@@ -0,0 +1,11 @@
+<!-- toc -->
+
+# Introduction
+
+The `math:` module provides mathematical functions and constants.
+
+Function usages are given in the same format as in the reference doc for the
+[builtin module](builtin.html). In particular, the discussion of [Commands
```
xiaq, line 8: > In particular, ...

* Write complete sentences to make the meaning unambiguous.

* Only use title case when explicitly referring to a section (such as "see section Number Type"). In this case you are referring to the *content* of a section.

In this case, write:

`In particular, all the commands in this module conform to the pattern of [commands that operate on numbers](...).`

pkg/eval/testutils.go
```diff
@@ -223,6 +226,29 @@ func matchOut(want, got []interface{}) bool {
 		return false
 	}
 	for i := range got {
+		// Equality of some data types needs to be special-cased in unit tests.
```
xiaq, line 229: I like how you have moved the approximate comparison into `matchOut`, but I'd prefer to make it even more explicit by doing the following:

1. Add a new type, `type Approximately struct{ f float64 }`.

2. When `want[i]` has that type, compare with `got[i]` with an epsilon.

3. Change the unit tests that need to compare float64 approximately from:

    ```go
    That("some code").Puts(1.0)
    ```

    to

    ```go
    That("some code").Puts(Approximately{1.0})
    ```

    Most test cases will *not* need approximate comparison. You can simply run the tests, see which ones are failing, and change those to use `Approximately`.

This way, test cases that need precise will continue to do so, and test cases that need approximate comparison are very explicit.

pkg/eval/math/math.go
```diff
@@ -0,0 +1,286 @@
...
+	"github.com/elves/elvish/pkg/eval/vars"
+)
+
+//elvdoc:var e
+//
+// ```elvish
+// $math:e
+// ```
+//
+// The value of E: 2.718281.... This variable is read-only.
```
xiaq, line 18: Write *e* with a lower case and italic to be consistent with how it is usually written.

Linking to Wikipedia would also be nice.

pkg/eval/math/math.go
```diff
@@ -0,0 +1,286 @@
...
+
+//elvdoc:var pi
+//
+// ```elvish
+// $math:pi
+// ```
+//
+// The value of Pi: 3.141592.... This variable is read-only.
```
xiaq, line 26: I'd write it as *π*, using the Greek letter in italic, and link to Wikipedia.

pkg/eval/math/math.go
```diff
@@ -0,0 +1,286 @@
...
+
+//elvdoc:fn abs
+//
+// ```elvish
+// math:abs $number
+// ```
+//
+// Compute the absolute value of a number. Example:
```
xiaq, line 34: Use third-person singular form of the verb, "computes".

In case you haven't noticed, there is now a [style guide](https://github.com/elves/elvish/blob/master/CONTRIBUTING.md#documenting-builtin-functions-and-variables) for the "elvdoc" comments.

Same for functions below.

pkg/eval/math/math.go
```diff
@@ -0,0 +1,286 @@
...
+// ▶ (float64 5.3)
+// ```
+
+//elvdoc:fn ceil
+//
+// ```elvish
+// math:ceil $number
+// ```
+//
+// Compute the floor value of a number. Example:
```
xiaq, line 49: This should say "ceiling".

Also, I like how https://godoc.org/math#Ceil actually explains the definition of ceiling and lists the special cases.

pkg/eval/math/math.go
```diff
@@ -0,0 +1,286 @@
...
+// ▶ (float64 -1)
+// ```
+
+//elvdoc:fn floor
+//
+// ```elvish
+// math:floor $number
+// ```
+//
+// Compute the floor value of a number. Example:
```
xiaq, line 79: See comment on `ceil` above.

pkg/eval/math/math.go
```diff
@@ -0,0 +1,286 @@
...
+// ▶ (float64 1)
+// ~> math:floor -2.3
+// ▶ (float64 -3)
+// ```
+
+//elvdoc:fn is-inf
+//
+// ```elvish
+// math:is-inf $number $sign
+// ```
```
xiaq, line 92: This function could have a more [Elven](https://elv.sh/ref/name.html#related-words) API by taking the `$sign` argument as an option instead, and defaulting to 0. This way it can be used as follows:

```
math:is-inf $n # tests if $n is +inf or -inf
math:is-inf $n &sign=-1 # tests if $n is -inf
math:is-inf $n &sign=1 # tests if $n is +inf
```

I also like how https://godoc.org/math#IsInf explains the use of `$sign` in separate sentences - it's more readable that way.

pkg/eval/math/math.go
```diff
@@ -0,0 +1,286 @@
...
+// ▶ $true
+// ```
+
+//elvdoc:fn is-nan
+//
+// ```elvish
+// math:is-nan $number $sign
+// ```
+//
+// If the number is not-a-number (NaN) output `$true` else `$false`.
```
xiaq, line 116: Say

> Tests whether the number is NaN.

pkg/eval/math/math.go
```diff
@@ -0,0 +1,286 @@
...
+// ▶ (float64 -0.0000000035897930298416118)
+// ```
+
+//elvdoc:fn trunc
+//
+// ```elvish
+// math:trunc $number
+// ```
+//
+// Output the integer value of the float64.
```
xiaq, line 255: Say

> Outputs the integer value of $number.

pkg/eval/math/math_test.go
krader1961: Yes, that was an oversight from a prior experiment for improving the test. And the result of that experiment was also flawed since the test should be using the elvish `$math:e` constant, rather than interpolating the Go `math:E` constant as you pointed out in a different comment. :smile:

pkg/eval/math/math_test.go
krader1961: Done

pkg/eval/math/math.go
krader1961: Yes, stupid copy/paste mistake. I will add a "Read the Go documentation..." link but that begs the question of when such links should be present.

pkg/eval/math/math.go
krader1961: Done

pkg/eval/vals/equal.go
krader1961: Even if I implement `Approximate()`, as you suggested in a different comment, we still need this constant or something like it. So I don't understand this comment.

pkg/eval/testutils.go
krader1961: > Most test cases will not need approximate comparison. You can simply run the tests, see which ones are failing, and change those to use Approximately.

Yes, most test cases do not need an approximate comparison. But that is simply because the existing tests have predictable bit patterns. For example, the constant `1` has a predicable float64 bit pattern. If every time a test involving float64 values fails due to a small epsilon, and the code is changed to use `Approximately()`, what is the point?

krader1961: I'm going to go ahead and implement your suggestion but note that a test such as

```
That(`math:cos 1`).Puts(fmt.Sprintf("%g", math.Cos(1.0))),
```

still fails because the default invocation of `vals.Equal()` doesn't handle the inconsequential difference between the two values.

pkg/eval/math/math.go
krader1961: Done

pkg/eval/math/math.go
krader1961: Done

pkg/eval/math/math.go
krader1961: Okay, but how to square the circle? That is, what about the commands like `math:round` which say `Output the nearest integer` rather than `Outputs...`?

pkg/eval/math/math.go
krader1961: SGTM

pkg/eval/math/math.go
krader1961: Ah, never mind, I see you made another comment about s/Output/Outputs/.

pkg/eval/vals/equal.go
xiaq: You already the same constant in testutils.go, with a different value :)

pkg/eval/testutils.go
xiaq: > But that is simply because the existing tests have predictable bit patterns. 

It's because Elvish needs to guarantee that certain floating point number operations *have* accurate results. For example, 1.0 + 2.0 needs to be exactly 3.0 (unlike 0.1 + 0.2, which is approximately 3.0). Since Elvish does not offer an integer type yet this has to be part of what Elvish promises.

> If every time a test involving float64 values fails due to a small epsilon, and the code is changed to use Approximately(), what is the point?

I suggested that you add `Approximately` to failing tests as a quick way to fix the tests, but that's *not* the criteria for whether a test should have precise or approximate in the first place. See comment above.

> ... still fails because the default invocation of vals.Equal() doesn't handle the inconsequential difference between the two values.

The difference *is* consequential because the two values *are* different. They have different JSON presentations for example, and the tests should be precise in which form it expects.

In general, I feel you are missing an important distinction between what *an average user would expect* and *what Elvish should promise*. It's completely OK for an Elvish user to think "oh floating point numbers are messy, better to not expect anything accurate", but the same attitude is counter-productive for an Elvish developer. You need to think about all the edges cases where the subtle difference can translate into a bigger one and be as rigorous as possible.

xiaq:
Please resolve the merge conflict and rebase to get accurate CI results.

pkg/eval/vals/equal.go
krader1961: Gah! Sorry, I thought I had removed that already.

krader1961:
Rebase done, all tests are passing, and I don't see any obvious documentation problems after running `cd website; make gen` and viewing the results in my web browser.

xiaq:
Going to merge this. @krader1961, thank you for your patience!

Issue #898, opened 2020-01-17
Warn user when using deprecated builtin features


xiaq:
(Edit: see https://github.com/elves/elvish/issues/898#issuecomment-602282707 for the current thinking on how to implement.)

Implement a method on `*Frame` that shows a deprecation message:

```go
func (*Frame) Deprecate(reason string)
```

However, during the entire lifecycle of an `Evaler`, the same deprecation message should only be shown for the same source location once. For example, if you have a function that uses a deprecated feature, the same message should only be shown the first time the function is called.

Until this feature is implemented, deprecated feature will have a `TODO(#898): Formally deprecate feature X` attached to them in the Go code.

krader1961:
I don't see how this proposal proactively tells users that a given feature is deprecated. They have to actually use the feature to see the deprecation message. Presumably you still intend to publish the deprecation, independent of this feature implementation, in a *CHANGELOG* file or similar mechanism.

What happens if the feature is used in a non-interactive context? If it occurs in a non-interactive context how does the deprecation message interact with the consumer of the output of the deprecated feature?

xiaq:
@krader1961 you are right, deprecating a feature also requires documenting it properly. I changed the title of this issue to reflect it.

Let me elaborate a bit on why I closed #939 citing this issue; I consider this functionality to be a *prerequisite* to deprecating any feature. If we document that something is deprecated, people *can* proactively change their code to avoid deprecated features, but (1) it's easy for people to miss part of the release notes (2) humans are error-prone and may not find all uses. Instead, it's much better for a machine (in this case, the Elvish interpreter) to discover uses of deprecated features and warn the user.

Re. where the warning goes in a non-interactive context: the warning goes to stderr.

xiaq:
On the technical side, a better way to implement this mechanism is during compilation instead of evaluation. This has a few advantages:

* The deprecation warning is shown even if the code that uses the deprecated function or variable is not executed

* Deprecation warnings are clustered - all deprecation warnings for the same source are shown at the same time, making it easier to fix in batch.

* There is no need to remember that a deprecation message has already been shown for a certain source location, since each source is compiled only once.

krader1961:
> Re. where the warning goes in a non-interactive context: the warning goes to stderr.

[...]

PR #939, opened 2020-03-11
Deprecate builtin:has-prefix and has-suffix

krader1961:
Related #576

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,12 @@
+# elvish 0.14
+
+## Deprecations
+
+The features being deprecated will be removed after the 0.15 release.
+
+- `builtin:has-prefix` is replaced by `str:has-prefix`.
+- `builtin:has-suffix` is replaced by `str:has-suffix`.
+
+## Notable non-backward compatible changes
+
+## Notable fixes and improvements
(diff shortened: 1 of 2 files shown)


krader1961:
Note that I've introduced a *CHANGELOG.md* because I really appreciate having such a file when looking at the history of other projects. It doesn't simply mirror the git commit history. It documents only the notable changes and organizes that information in a way that is hopefully useful to a human.

xiaq:
Elvish already has changelogs; it is the release notes in website/blog. Up until now I only write release notes shortly before releasing a new version, but nothing prevents us from creating the release notes for the next version early. As long as it is not put in website/blog/index.toml it won't be visible on the website.

This is actually what Go does; you can find notes for changes introduced after the last release (Go 1.14) at the unpublished release notes for Go 1.15, at https://tip.golang.org/doc/go1.15.

While documenting that a feature is deprecated is fine, I prefer introducing a formal deprecation warning mechanism (#898) before starting to deprecate anything.

krader1961:
Okay, I can live with an alternative to a *CHANGELOG* file. But I disagree that the proposal in #898 is a substitute for publishing a deprecation warning for feature "x". Let's move that discussion to that issue.

Issue #561, opened 2018-01-10
Inconsistencies when using a period in a namespace name


zzamboni:
It seems that periods are in general allowed in namespace names, except when assigning variables. Assume for example:

```
[~/.elvish]─> epm:install github.com/zzamboni/theme.elv
```

Then the following works:

```
[~/.elvish]─> use github.com/zzamboni/theme.elv/chain
[~/.elvish]─> chain:setup
[~/.elvish]─> echo $chain:prompt_pwd_dir_length
1
[~/.elvish]─> chain:prompt_pwd_dir_length = 2
[~/.elvish]─> echo $chain:prompt_pwd_dir_length
2
```

If `theme.elv` is left in the namespace, then function calls work, but no variable references or assignments:

```
[~/.elvish]─> use github.com/zzamboni/theme.elv:chain
[~/.elvish]─> theme.elv:chain:setup
[~/.elvish]─> echo $theme.elv:chain:prompt_pwd_dir_length
Compilation error: variable $theme not found
[tty], line 1: echo $theme.elv:chain:prompt_pwd_dir_length
[~/.elvish]─> theme.elv:chain:prompt_pwd_dir_length = 2
Parse error: bad assignment LHS
, line 1: theme.elv:chain:prompt_pwd_dir_length = 2
```

Is there an established rule here? I think if periods are allowed in the namespace at all, then they should work consistently, otherwise they should be disallowed.

xiaq:
Right, We need to check that namespaces introduced with `use` are valid variable names.

Moreover, there is a similar situation with `fn`:

```
fn a.b { }
a.b # works
put $a.b~ # doesn't
```

PR #940, opened 2020-03-17
Document namespace/function naming rules

krader1961:
This only documents the rules for namespace and function names.
Specifically, that a namespace or function name must be a valid variable
name.  It does not enforce that restriction.

Related #561

### DIFF
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -1618,8 +1618,8 @@ also comes with the following modules:
 
 You can define your own modules with Elvishscript by putting them in
 files under `~/.elvish/lib` and giving them a `.elv` extension. The file
-name must be a valid [variable name](#variable). So `a.elv` is valid but
-`a.b.elv` is not valid as a module name.
+name must be a valid [variable name](#variable). For example, `a.elv`
+is valid but `a.b.elv` is not valid as a module name.
 
 For instance, to define a module named `a`, store it in
 `~/.elvish/lib/a.elv`:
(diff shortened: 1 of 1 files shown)


website/ref/language.md
```diff
@@ -1399,7 +1399,11 @@ Syntax:
 fn <name> <lambda>
 ```
 
-Define a function with a given name. The function behaves in the same way to the
+Define a function with a given name. The function name must be a valid
+[variable name](#variable). So `a` is valid but `a.b` is not valid as a
```
xiaq: I find it clearer to only use "so" to introduce general conclusions, rather than examples. Stick to "for example" for examples.

krader1961: Works for me. Done.

website/ref/language.md
```diff
@@ -1612,9 +1616,13 @@ following modules:
 
 ## User-Defined Modules
 
-You can define your own modules with Elvishscript by putting them under
-`~/.elvish/lib` and giving them a `.elv` extension. For instance, to define a
-module named `a`, store it in `~/.elvish/lib/a.elv`:
+You can define your own modules with Elvishscript by putting them in
+files under `~/.elvish/lib` and giving them a `.elv` extension. The file
+name must be a valid [variable name](#variable). So `a.elv` is valid but
```
xiaq: Same wording feedback here.

website/ref/language.md
```diff
@@ -1662,8 +1670,10 @@ use `use x/y/z xyz`:
 f from x/y/z
 ```
 
-This is especially useful when you need to import several modules that are in
-different directories but have the same file name.
+This is especially useful when you need to import several modules that
+are in different directories but have the same file name. The alias name
+must be a valid [variable name](#variable). So `c_prime` is valid but
```
xiaq: Same wording feedback here.

PR #942, opened 2020-03-19
Validate function names

krader1961:
Related #561

### DIFF
--- a/pkg/eval/builtin_special.go
+++ b/pkg/eval/builtin_special.go
@@ -169,6 +169,9 @@ func compileFn(cp *compiler, fn *parse.Form) effectOpBody {
 	args := cp.walkArgs(fn)
 	nameNode := args.next()
 	varName := mustString(cp, nameNode, "must be a literal string") + FnSuffix
+	if !parse.IsValidVariableName(varName) {
+		cp.errorpf(nameNode.Range().From, nameNode.Range().To, "must be a valid variable name")
+	}
 	bodyNode := args.nextMustLambda()
 	args.mustEnd()
 
@@ -215,6 +218,7 @@ func (op fnWrap) invoke(fm *Frame) error {
 // UseForm = 'use' StringPrimary
 func compileUse(cp *compiler, fn *parse.Form) effectOpBody {
 	var name, spec string
+	var nameArg parse.Node
 
 	switch len(fn.Args) {
 	case 0:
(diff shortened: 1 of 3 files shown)


krader1961:
I'm thinking it makes more sense to add a `ValidVariableName` function, call it with the function name string, and letting it validate each rune. Updated commit forthcoming....

krader1961:
I have another commit that builds on this one to validate `use` namespace strings. I'll create a PR with that commit after this one is merged.

xiaq:
I saw you commented about exposing a ValidVariableName function but for some reason that comment is gone?

I am going to make the same comment. Code outside the parse package has little need for a function that tests if a rune is allowed in variable name; it is almost always going to test a string.

krader1961:
Okay, I'll introduce such a function. I had opted to expose the heretofore private function because I thought it might be useful to tell the user which specific rune was invalid. But then I decided not to do so since the user should be expected to know the var naming rules and telling them that `.`, for example, was the invalid char adds complexity for very little benefit.

krader1961:
@xiaq, Please review. I snuck a trivial `go fmt` cleanup into this commit that fixes a style problem from my previous commit you merged earlier today. I verified that the tests fail without the corresponding validation and that this works as expected when using `fn` or `use` interactively with invalid names.

pkg/eval/builtin_special.go
```diff
@@ -215,6 +218,7 @@ func (op fnWrap) invoke(fm *Frame) error {
 // UseForm = 'use' StringPrimary
 func compileUse(cp *compiler, fn *parse.Form) effectOpBody {
 	var name, spec string
+	var name_arg *parse.Compound
```
xiaq: Go variables never use snake_case. Also, the code below does not need to know that the name argument is kept in a `parse.Compound` node, so declare it with the more general `parse.Node`:

```go
var nameArg parse.Node
```

pkg/eval/platform/platform.go
```diff
@@ -25,5 +25,5 @@ import (
 
 var Ns = eval.Ns{
 	"arch": vars.NewReadOnly(runtime.GOARCH),
-	"os": vars.NewReadOnly(runtime.GOOS),
+	"os":   vars.NewReadOnly(runtime.GOOS),
```
xiaq: I have fixed the format in the latest commit, please rebase.

In case you haven't seen it, please use `goimports` to format code and configure your editor to run this automatically upon save (https://github.com/elves/elvish/blob/master/CONTRIBUTING.md#formatting-source-files).

krader1961: I installed `goimports` several months ago but haven't developed the "muscle memory" to run it when needed. I'm not a huge fan of running such tools in my editor context. I prefer git hooks or a `make style` target.

pkg/eval/builtin_special.go
krader1961: That results in these build errors:

```
pkg/eval/builtin_special.go:233:11: cannot use fn.Args[0] (type *parse.Compound) as type *parse.Node in assignment:
        *parse.Node is pointer to interface, not interface
pkg/eval/builtin_special.go:240:11: cannot use fn.Args[1] (type *parse.Compound) as type *parse.Node in assignment:
        *parse.Node is pointer to interface, not interface
pkg/eval/builtin_special.go:249:21: nameArg.Range undefined (type *parse.Node is pointer to interface, not interface)
```

xiaq: You wrote *parse.Node I assume? Use parse.Node.

krader1961: Doh! Thx.

krader1961:
I think we're good to go. I also added two more test cases to verify `use` aliasing.

pkg/parse/parse.go
```diff
@@ -718,6 +718,13 @@ func allowedInVariableName(r rune) bool {
 		r == '-' || r == '_' || r == ':' || r == '~'
 }
 
+func IsValidVariableName(s string) bool {
```
xiaq: Missed this earlier - since this is an exported function, this should have a godoc comment.

krader1961: Done

PR #949, opened 2020-03-25
Introduce a unix module and implement $unix:umask

### LINKED ISSUE #681: Add umask command
The title speaks for itself, I think. Especially on shared systems, it can sometimes be useful to be able to change the process's `umask` value.

krader1961:
Resolves #681


pkg/eval/unix/umask.go
```diff
@@ -0,0 +1,94 @@
...
+		i, err := strconv.ParseInt(v, 8, 0)
+		if err != nil {
+			i, err = strconv.ParseInt(v, 0, 0)
+			if err != nil {
+				return errors.New("umask value not a valid number")
+			}
+		}
+		umask = int(i)
+	case float64:
+		int_part, frac_part := math.Modf(v)
```
xiaq: Do not use underscores in variable names.

pkg/eval/unix/umask.go
```diff
@@ -0,0 +1,94 @@
...
+//
+// This variable only has meaning on UNIX like operating systems and does not
+// exist on other platforms. Do not use this in elvish scripts meant to work
+// on non-UNIX compatible systems.
+
+// UmaskVariable is a variable whose value always reflects the current file
+// creation permission mask. Setting it changes the current file creation
+// permission mask for the process (not an individual thread).
+type UmaskVariable struct {
+	ev *eval.Evaler
```
xiaq: Remove?

pkg/eval/unix/umask.go
```diff
@@ -0,0 +1,94 @@
...
+	"strconv"
+	"syscall"
+
+	"github.com/elves/elvish/pkg/eval"
+	"github.com/elves/elvish/pkg/eval/vars"
+)
+
+//elvdoc:var umask
+//
+// The file system creation mask. When used in a string context the output is
```
xiaq: * POSIX calls this "file mode creation mask", let's stick to that.

* Elvish does not have "string context". Just say "its value is a string in octal representation".

pkg/eval/unix/umask.go
```diff
@@ -0,0 +1,94 @@
...
+// its Elvish octal representation; e.g. 0o027. This makes it possible to use
+// it in any context that expects a `$number`.
+//
+// When assigning a new value a string is implicitly treated as an octal
+// number. If that fails the usual rules for interpreting
+// [numbers](./language.html#number) are used. This means you can do both
```
xiaq: > This means you can do both ...

Instead of saying one can *do* both, say that they are equivalent:

> The following are equivalent: ...

pkg/eval/unix/umask.go
```diff
@@ -0,0 +1,94 @@
...
+// `unix:umask = 027` and `unix:umask = 0o27`. You can also assign to it a
+// `float64` data type that has no fractional component. The legal values are
```
xiaq: > The legal values are ...

The word "legal" is not well-defined. Say what will happen if it is not within the range:

> The assigned value must be within the range [0 ... 0o777], otherwise the assignment will throw an exception.

pkg/eval/unix/umask.go
```diff
@@ -0,0 +1,94 @@
...
+// in the range [0 .. 0o777].
+//
+// You can do a temporary assignment to affect a single command; e.g.
+// `umask=077 touch a_file`. After the command completes the old umask will be
+// restored. **Warning**: Since the umask applies to the entire process, not
+// individual threads, changing it temporarily in this manner is dangerous if
+// you are doing anything in parallel. Such as via the
+// [`peach`](ref/builtin.html#peach) command.
+//
+// This variable only has meaning on UNIX like operating systems and does not
```
xiaq: Put this warning in unix.md; everything in the `unix` module is unavailable to non-UNIX operating systems.

pkg/eval/unix/umask.go
```diff
@@ -0,0 +1,94 @@
+// +build !windows,!plan9
+
+package unix
+
+import (
+	"errors"
+	"fmt"
+	"math"
+	"strconv"
+	"syscall"
```
xiaq: Do not use the `syscall` package; it's deprecated. Use golang.org/x/sys/unix instead.

pkg/eval/unix/unix.go
```diff
@@ -0,0 +1,11 @@
...
+// unique to UNIX like operating systems. On other operating systems it will
+// be an empty namespace; or have a subset of the features normally available
+// on a UNIX like operating system.
+package unix
+
+import (
+	"github.com/elves/elvish/pkg/eval"
+)
+
+var Ns = eval.Ns{}
```
xiaq: I think it's cleaner to declare the namespace literally here, and have another `unix_non_unix.go` that declares an empty namespace.

pkg/eval/unix/unix.go
```diff
@@ -0,0 +1,11 @@
+// Package `unix` exposes variables and functions that deal with features
+// unique to UNIX like operating systems. On other operating systems it will
```
xiaq: * UNIX like -> UNIX-like

* Let's make up our mind and just say non-UNIX operating systems will see an empty namespace.

pkg/program/shell/runtime.go
```diff
@@ -72,10 +73,13 @@ func InitRuntime(binpath, sockpath, dbpath string) (*eval.Evaler, string) {
 
 	ev := eval.NewEvaler()
 	ev.SetLibDir(filepath.Join(dataDir, "lib"))
+
 	ev.InstallModule("math", eval.MathNs)
+	ev.InstallModule("platform", platform.Ns)
 	ev.InstallModule("re", re.Ns)
 	ev.InstallModule("str", str.Ns)
-	ev.InstallModule("platform", platform.Ns)
+	ev.InstallModule("unix", unix.Ns)
```
xiaq: Don't install this namespace at all for non-UNIX operating systems.

You can do this by declaring a constant in the `unix` package, whose value is true on UNIX but false on others.

website/ref/unix.md
```diff
@@ -0,0 +1,10 @@
+<!-- toc -->
+
+# Introduction
+
+The `unix:` module provides access to features that only make sense on UNIX
+compatible operating systems such as Linux, FreeBSD, macOS, etc. Do not use
```
xiaq: Instead of saying "do not use", just say that this module does not exist on non-UNIX OSes (assuming you acted on my other comment of not installing the module at all for non-UNIX OSes).

pkg/eval/unix/umask.go
```diff
@@ -0,0 +1,94 @@
...
+// permission mask for the process (not an individual thread).
+type UmaskVariable struct {
+	ev *eval.Evaler
+}
+
+var _ vars.Var = UmaskVariable{}
+
+// Get returns the current file creation umask as a string.
+func (UmaskVariable) Get() interface{} {
+	// Note: The seemingly redundant syscall is because the syscall.Umask() API
```
xiaq: Interesting, I never knew this weird restriction of UNIX syscall API.

The workaround is fine per se, but now you have a race condition: if another thread calls `Umask` between the two `Umask` call here, this code will revert that change.

You can use a sync.RWMutex to guard the calls made in this file, although that still won't fix calls made in other places. Fortunately nowhere else in Elvish calls `Umask` (except for the daemon, but then the daemon does not run the Elvish interpreter).

pkg/eval/unix/umask.go
krader1961: Breaking that habit, instilled by3+ decades of doing that in nearly all the C code I've ever authored or touched, is going to take a few applications of the old clue-by-four :smile:

pkg/eval/unix/umask.go
krader1961: Good point.

krader1961: In fact, I already have an equivalent warning in *unix.md*.

pkg/eval/unix/unix.go
krader1961: Okay, that works for me. But it does mean the `// +build` directive will need to be updated when new platforms are supported by Go. And we'll want to proactively include the `js` platform name.

pkg/program/shell/runtime.go
krader1961: Okay, but it seemed cleaner to me to unconditionally install an empty namespace rather than predicate this statement on whether we're actually building on a UNIX-like system. In fact, if I do this why would we need a `unix_non_unix.go` that declares an empty namespace?

pkg/eval/unix/umask.go
krader1961: Yes, this is racy but that is inherent in the umask API. Even with a mutex you'll have a problem since that won't stop any code that uses the `umask()` syscall from racing with this method. Note that the fish shell (written in C++ and uses threads) does this without a mutex and it hasn't been a problem. A mutex here would provide a false sense of security against a race.

krader1961:
@xiaq Do you prefer people push PR updates as deltas (i.e., standalone commits) or forced pushes after squashing the update into the previous commit? I'm asking because I noticed you merged a couple of my earlier PRs without squashing the subsequent change(s) I made in response to your review feedback. Which is not how other projects typically handle pull-request deltas made as a result of review feedback.

pkg/eval/unix/umask.go
krader1961: Removed, although I'll note that I copied this from the *pkg/eval/pwd.go* module. I now see that it's needed in that module to support the `pwd.ev.Chdir(path)` callback. Looking at the *pkg/eval/pwd.go* module I don't understand the point of this statement:

```
var _ vars.Var = PwdVariable{}
```

Removing that statement (and the associated vars import) doesn't seem to cause any harm. The program builds and all unit tests pass.

krader1961:
@xiaq, I've addressed all of your points with one exception. You made two contradictory recommendations with regarding to the `unix:` namespace on non-UNIX platforms. I have tentatively decided it makes more sense to make the namespace visible on non-UNIX platforms. Adding a `$unix:usable` variable makes it easy, and clear, to determine if that namespace is usable. The original approach of creating an empty namespace required doing `== 0 (count (keys $unix:))` to determine if the namespace was usable. The `$unix:usable` var means any Elvish program can safely do `use unix` then `if (not $unix:usable) { echo error message; exit 123 }`.

krader1961:
My introduction of the `$unix:usable` bool is also meant, obviously, to set a precedent. This makes it easy for an Elvish program to determine if it is running on a non-UNIX platform without resorting to considerably more complicated mechanisms such as comparing the `$platform:os` var against a list of acceptable values. The intent is to make it possible to introduce other optional namespaces such as `windows:` (or `win:`) that Elvish programs can easily, and clearly, test if they are usable.

pkg/eval/unix/non_unix.go
```diff
@@ -0,0 +1,22 @@
+// +build  windows plan9 js
+
+// Package `unix` exposes variables and functions that deal with features
```
xiaq: Do not use backticks in godoc; they are not interpreted as markdown.

pkg/eval/unix/umask.go
```diff
@@ -0,0 +1,85 @@
...
+// individual threads, changing it temporarily in this manner is dangerous if
+// you are doing anything in parallel. Such as via the
+// [`peach`](ref/builtin.html#peach) command.
+
+// UmaskVariable is a variable whose value always reflects the current file
+// creation permission mask. Setting it changes the current file creation
+// permission mask for the process (not an individual thread).
+type UmaskVariable struct{}
+
+var _ vars.Var = UmaskVariable{}
```
xiaq: The statement acts as an assertion that the UmaskVariable type satisfies the vars.Var interface.

pkg/eval/unix/umask.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Get returns the current file creation umask as a string.
+func (UmaskVariable) Get() interface{} {
+	// Note: The seemingly redundant syscall is because the unix.Umask() API
+	// doesn't allow querying the current value without changing it. So ensure
+	// we reinstate the curent value.
+	umask := unix.Umask(0)
```
xiaq: Re adding a mutex: since Elvish does not use Umask elsewhere concurrently, guarding UmaskVariable with a mutex will prevent possible race conditions. Elvish does not link with arbitrary Go code, so as long as we control the Elvish source code we have a guarantee.

The fact that Fish has a race condition does not make the race condition less likely :)

pkg/eval/unix/umask_test.go
```diff
@@ -0,0 +1,40 @@
+// +build !windows,!plan9,!js
+
+// Note that this unit test assumes a UNIX environment with a POSIX compatible
```
xiaq: Do not put comments before the `package` statement; it is only used for package godoc. Put it after the `package` statement.

pkg/program/shell/runtime.go
xiaq: You need a `unix_non_unix.go` to make the code compile :)

My preference is to have a **Go** `unix` package with an empty `Ns` variable on non-Unix OS, and to not install the **Elvish** `unix` module on non-Unix OS.

xiaq:
Re. making it easy for Elvish code to decide whether the `unix` module is available: you can introduce a `$platform:is-unix` variable instead.

Re. preference of how you push code: I don't mind, since when merging a PR in GitHub you can optionally squash all the commits. When I see people pushing commits that say "address PR feedback" I'd use "sqush and merge"; when I see people force-push I'd use "rebase and merge".

pkg/eval/unix/umask.go
krader1961: Thanks. Googling "go assert type satisfies interface" returns many documents that address this question. Here are a couple for those thinking about contributing pull-requests like this one in the future:

https://splice.com/blog/golang-verify-type-implements-interface-compile-time/
https://medium.com/@matryer/golang-tip-compile-time-checks-to-ensure-your-type-satisfies-an-interface-c167afed3aae

krader1961:
> Re. making it easy for Elvish code to decide whether the unix module is available: you can introduce a $platform:is-unix variable instead.

SGTM. I chose the current solution because it localizes the var for determining the usability of a module to that module. But given the existence of the `platform` module it is also reasonable to place equivalent vars in that module. I'll create an independent PR to do that and also introduce a `$platform:is-windows` variable at the same time since those are the cases of greatest interest at this time.

pkg/eval/unix/non_unix.go
krader1961: Is there some other convention for indicating that a string is a code literal in such comments?

pkg/eval/unix/umask.go
krader1961: I still think a mutex for this is overkill but it's trivial to do and won't hurt performance unless someone is intentionally trying to trigger the race -- in which case the mutex is needed -- so I'll go ahead and add one :-)

pkg/eval/unix/umask_test.go
krader1961: Okay.

pkg/eval/unix/non_unix.go
krader1961: In any case I've removed the backticks.

krader1961:
@xiaq, Please take another look. I've rebased on the recent changes to the master branch but otherwise the changes only addressed your most recent feedback.

pkg/eval/unix/umask.go
krader1961: I added a mutex: `var umaskMutex sync.Mutex`. Is something wrong with how I used it?

pkg/program/shell/runtime.go
krader1961: Done

pkg/eval/unix/umask.go
```diff
@@ -0,0 +1,92 @@
...
+	unix.Umask(umask)
+	return fmt.Sprintf("0o%03o", umask)
+}
+
+// Set changes the current file creation umask. It can be called with a string
+// (the usual case) or a float64.
+func (UmaskVariable) Set(v interface{}) error {
+	var umask int
+
+	umaskMutex.Lock()
```
xiaq, line 63: This Lock call can be moved below, just above the `unix.Umask` call.

pkg/eval/unix/unix.go
```diff
@@ -0,0 +1,16 @@
...
+
+// Package unix exposes variables and functions that deal with features
+// unique to UNIX-like operating systems. On non-UNIX operating systems it
+// will be an empty namespace.
+package unix
+
+import (
+	"github.com/elves/elvish/pkg/eval"
+)
+
```
xiaq, line 11: Exported symbols should have godocs.

pkg/eval/unix/umask.go
xiaq: Ah, I missed that, sorry.

It's correct, but a small nitpick: use a `RWMutex` to allow concurrent reads. Not that it's likely to have a lot of concurrent reads, more of a stylistic issue.

pkg/eval/unix/umask.go
```diff
@@ -0,0 +1,92 @@
...
+// you are doing anything in parallel. Such as via the
+// [`peach`](ref/builtin.html#peach) command.
+
+// UmaskVariable is a variable whose value always reflects the current file
+// creation permission mask. Setting it changes the current file creation
+// permission mask for the process (not an individual thread).
+type UmaskVariable struct{}
+
+var _ vars.Var = UmaskVariable{}
+var umaskMutex sync.Mutex
```
xiaq, line 44: It's worth documenting that this mutex operates on the assumption that no other part of the Elvish codebase would be calling `unix.Umask` concurrently.

pkg/eval/unix/umask.go
krader1961: done

pkg/eval/unix/unix.go
krader1961: done

pkg/eval/unix/umask.go
krader1961: done

krader1961:
> It's correct, but a small nitpick: use a RWMutex to allow concurrent reads.

That would be incorrect in this instance since the `Get()` method mutates the state. Allowing concurrent readers means that one of them might see, and return, the transitory zero value.

xiaq:
> That would be incorrect in this instance since the Get() method mutates the state. Allowing concurrent readers means that one of them might see, and return, the transitory zero value.

Oops, you are right!

Please format the files with `goimports`. You can format all Go files with this Elvish command:

```
goimports -w **.go
```

If you don't like the option of running `goimports` when the editor saves files, I'd suggest that you set up some Git commit hook to run this.

krader1961:
I ran goimports. Please squash merge.

P.S., I've gone ahead and added a write hook to my Vim config. The problem with that approach is that if I edit a Go source file using some other program then goimports won't automatically be run. How would you feel about adding a `make style` build target that runs goimports and prettier if they are available?

krader1961:
Also, I just installed `prettier` and ran it via `prettier --tab-width 4 --prose-wrap always --write **.md`. It modified six files :smile:

xiaq:
@krader1961 I introduced a `make style` target, enjoy :)

A few more nitpicks on the PR, but I'll merge it anyway and address myself:

* The godoc comment for the package should say clearly that it introduces an **Elvish** namespace;
* The godoc comment block for an exported symbol should always start with the exported symbol itself.

Also, I am not sure what the test case in `unix_test.go` is supposed to do: currently it verifies that the `unix:` module has at most 1 symbol; that number will simply need to be bumped when it gets more symbols, which seems repetitive and not very useful to me.

Do you have a specific failure mode in mind that this test case is supposed to catch? I'll remove it if it's not clear why it is there.

krader1961:
> Also, I am not sure what the test case in unix_test.go is supposed to do: currently it verifies that the unix: module has at most 1 symbol; that number will simply need to be bumped when it gets more symbols, which seems repetitive and not very useful to me.

That's leftover from when I intended to make it possible to `use unix` even on non-unix systems. It simply ensures that the namespace isn't empty -- which it would have been on non-unix systems. Nuke it.

PR #950, opened 2020-03-30
Implement $platform:is-unix and is-windows

krader1961:
This will facilitate elvish scripts determining whether it is safe to
`use unix` or `use windows`. The implementation of the `unix` module is
out for review in a separate commit. The `windows` module is likely to
be implemented in the near future.

### DIFF
--- a/pkg/eval/platform/platform.go
+++ b/pkg/eval/platform/platform.go
@@ -23,7 +23,21 @@ import (
 // [`GOOS`](https://pkg.go.dev/runtime?tab=doc#pkg-constants) constant.
 // This is read-only.
 
+//elvdoc:var is-unix
+//
+// Whether or not the platform is UNIX-like. This includes Linux, macOS
+// (Darwin), FreeBSD, NetBSD, and OpenBSD. This can be used to decide, for
+// example, if the `unix` module is usable.
+// This is read-only.
+
+//elvdoc:var is-windows
+//
+// Whether or not the platform is Microsoft Windows.
+// This is read-only.
+
 var Ns = eval.Ns{
-	"arch": vars.NewReadOnly(runtime.GOARCH),
-	"os":   vars.NewReadOnly(runtime.GOOS),
+	"arch":       vars.NewReadOnly(runtime.GOARCH),
+	"os":         vars.NewReadOnly(runtime.GOOS),
+	"is-unix":    vars.NewReadOnly(isUnix),
+	"is-windows": vars.NewReadOnly(isWindows),
 }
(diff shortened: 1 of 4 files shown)


pkg/eval/platform/is-unix.go
```diff
@@ -0,0 +1,6 @@
+// +build aix darwin dragonfly freebsd illumos linux netbsd openbsd solaris
```
xiaq, line 1: For consistency with code elsewhere, stick to `!windows,!plan9,!js`.

pkg/eval/platform/is-unix.go
```diff
@@ -0,0 +1,6 @@
...
+
+package platform
+
+var isUnix bool = true
```
xiaq, line 5: Leave out the type, and use a `var` block:

```go
var (
    isUnix = true
    isWindows = false
)

File: pkg/eval/platform/is-windows.go [hunk 3]
```diff
@@ -0,0 +1,6 @@
+// +build windows
+
+package platform
+
+var isUnix bool = false
```
xiaq, line 5: Ditto

pkg/eval/platform/platform.go
```diff
@@ -23,7 +23,21 @@ import (
 // [`GOOS`](https://pkg.go.dev/runtime?tab=doc#pkg-constants) constant.
 // This is read-only.
 
+//elvdoc:var is-unix
+//
+// Whether or not the platform is UNIX-like (i.e., not a specific flavor such
```
xiaq, line 28: The part in the parentheses is a bit confusing. Instead, enumerate all the OSes that are now considered Unix.

Since Elvish only supports Linux, macOS and {Free,Net,Open}BSD, listing those would be sufficient.

pkg/eval/platform/is-unix.go
krader1961: Okay

pkg/eval/platform/is-unix.go
krader1961: Done

pkg/eval/platform/platform.go
krader1961: Okay, but I don't see any reason why other UNIX-like platforms such as AIX, DragonFly and Illumos (the Solaris fork) shouldn't also work. However, I can appreciate why you might not want to give that impression since you probably haven't tested elvish on any of those platforms.

krader1961:
> .... Naming them unix.go and windows.go would be much better....

Done

krader1961:
> in Go it's customary to use underscores as part of filenames

It seems to me that is not so much a Go custom as a general UNIX custom. But that doesn't invalidate your comment :smile:

Issue #383, opened 2017-07-05
Consider following XDG base directory spec


cesarandreu:
The docs show the default config location is `~/.elvish/rc.elv`. Please consider following the [XDG base directory spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). The default location would then be `~/.config/elvish/rc.elv`.

Moving configs into a folder helps reduce clutter. In the last few years many tools have slowly been adding support: [XDG Base Directory support](https://wiki.archlinux.org/index.php/XDG_Base_Directory_support). For comparison, [fish shell](https://fishshell.com) uses `~/.config/fish/config.fish` as its startup file, and lazy-loaded functions go in `~/.config/fish/functions/`.

On macOS most tools generally continue using `~/.config` as the default location. However, should you choose to follow Apple's guidelines ([macOS Library Directory Details](https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html#//apple_ref/doc/uid/TP40010672-CH10-SW1)), you would use `~/Library/Application Support/elvish/` to place any configs, data, and files. Two other commonly used folders are Caches (`~/Library/Caches/elvish/`) and Logs (`~/Library/Logs/elvish/`).

zchee:
FYI: I wrote xdg basedir package for Go.
https://github.com/zchee/go-xdgbasedir

fehnomenal:
I want to take a stab on this.

@xiaq are you OK with the following design?

- `~/.elvish/` exists: Do not change behavior because this is most likely a legacy (pre-xdg) install. Output a message on launch to advertise moving to `$E:XDG_CONFIG_HOME/elvish`.
- Add a flag `-config` that defaults to `$E:XDG_CONFIG_HOME/elvish` or `~/.config/elvish`.
`dataDir` from https://github.com/elves/elvish/blob/702edb3d6523348cb0a8dd44d835432bf11412ab/runtime/runtime.go#L53 gets assigned that flag value.

I would use @zchee's implementation with the `Native` mode to honor Apple's guidelines, see https://github.com/zchee/go-xdgbasedir/blob/7a0b8922e178cedb342090b90ef662084d88eba8/xdgbasedir.go#L16-L27

xiaq:
I'm OK with following XDG. However if we are going to move directory, better make it the last directory move we ever have to do. The current `.elvish` directory has 3 things:

[...]

PR #955, opened 2020-04-02
Use XDG runtime directory on Unix when present

vially:
This pull-request uses `$XDG_RUNTIME_DIR` on Unix to store the runtime files. If `$XDG_RUNTIME_DIR` is not set it falls back to the old location (e.g.: `/tmp/elvish-$uid`).

This is a small first step towards [XDG](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) compliance on Unix (#383).

### DIFF
--- a/pkg/program/shell/run_dir_unix.go
+++ b/pkg/program/shell/run_dir_unix.go
@@ -15,12 +15,13 @@ var (
 	errBadPermission = errors.New("bad permission")
 )
 
-// getSecureRunDir stats elvish-$uid under the default temp dir, creating it if
-// it doesn't yet exist, and return the directory name if it has the correct
-// owner and permission.
+// getSecureRunDir stats elvish under the directory defined by the
+// XDG_RUNTIME_DIR environment variable, creating it if it doesn't yet exist,
+// and return the directory name if it has the correct owner and permission.
+// If XDG_RUNTIME_DIR is not set it falls back to elvish-$uid under the default
+// temp dir.
 func getSecureRunDir() (string, error) {
-	uid := os.Getuid()
-	runDir := filepath.Join(os.TempDir(), fmt.Sprintf("elvish-%d", uid))
+	runDir := getRunDirPath()
 	err := os.MkdirAll(runDir, 0700)
 	if err != nil {
 		return "", fmt.Errorf("mkdir: %v", err)
@@ -31,7 +32,19 @@ func getSecureRunDir() (string, error) {
 		return "", err
 	}
 
-	return runDir, checkExclusiveAccess(info, uid)
+	return runDir, checkExclusiveAccess(info, os.Getuid())
+}
+
+// getRunDirPath returns a path that's used by Elvish to store runtime files.
+// This path will be either XDG_RUNTIME_DIR/elvish when the XDG_RUNTIME_DIR
+// environment variable is set, or TMPDIR/elvish-$uid otherwise.
+func getRunDirPath() string {
+	if os.Getenv("XDG_RUNTIME_DIR") != "" {
+		return filepath.Join(os.Getenv("XDG_RUNTIME_DIR"), "elvish")
+	}
+
+	uid := os.Getuid()
+	return filepath.Join(os.TempDir(), fmt.Sprintf("elvish-%d", uid))
 }
 
 func checkExclusiveAccess(info os.FileInfo, uid int) error {
--- a/pkg/program/shell/run_dir_unix.go
+++ b/pkg/program/shell/run_dir_unix.go
@@ -21,7 +21,14 @@ var (
 // If XDG_RUNTIME_DIR is not set it falls back to elvish-$uid under the default
 // temp dir.
 func getSecureRunDir() (string, error) {
-	runDir := getRunDirPath()
+	runDirs := getRunDirPaths()
+	for _, runDir := range runDirs {
+		if runDirExistsWithExclusiveAccess(runDir) {
+			return runDir, nil
+		}
+	}
+
+	runDir := runDirs[0]
 	err := os.MkdirAll(runDir, 0700)
 	if err != nil {
 		return "", fmt.Errorf("mkdir: %v", err)
@@ -35,16 +42,30 @@ func getSecureRunDir() (string, error) {
 	return runDir, checkExclusiveAccess(info, os.Getuid())
 }
 
-// getRunDirPath returns a path that's used by Elvish to store runtime files.
-// This path will be either XDG_RUNTIME_DIR/elvish when the XDG_RUNTIME_DIR
-// environment variable is set, or TMPDIR/elvish-$uid otherwise.
-func getRunDirPath() string {
+// getRunDirPaths returns an array of paths that's used by Elvish to determine
+// where to store the runtime files. The paths are sorted in descending order
+// of preference and will always contain at least one path that points to
+// TMPDIR/elvish-$uid as the last item in the array.
+// When XDG_RUNTIME_DIR is set, the XDG_RUNTIME_DIR/elvish path will be the
+// first entry in the arrray.
+func getRunDirPaths() []string {
+	tmpDirPath := filepath.Join(os.TempDir(), fmt.Sprintf("elvish-%d", os.Getuid()))
 	if os.Getenv("XDG_RUNTIME_DIR") != "" {
-		return filepath.Join(os.Getenv("XDG_RUNTIME_DIR"), "elvish")
+		xdgDirPath := filepath.Join(os.Getenv("XDG_RUNTIME_DIR"), "elvish")
+		return []string{xdgDirPath, tmpDirPath}
 	}
 
-	uid := os.Getuid()
-	return filepath.Join(os.TempDir(), fmt.Sprintf("elvish-%d", uid))
+	return []string{tmpDirPath}
+}
+
+func runDirExistsWithExclusiveAccess(runDir string) bool {
+	info, err := os.Stat(runDir)
+	if err != nil {
+		return false
+	}
+
+	err = checkExclusiveAccess(info, os.Getuid())
+	return err == nil
 }
 
 func checkExclusiveAccess(info os.FileInfo, uid int) error {
--- a/pkg/program/shell/run_dir_unix_test.go
+++ b/pkg/program/shell/run_dir_unix_test.go
@@ -1,0 +1,112 @@
+// +build !windows,!plan9
+
+package shell
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"syscall"
+	"testing"
+
+	"github.com/elves/elvish/pkg/util"
+)
+
+func TestGetSecureRunDir(t *testing.T) {
+	xdgRuntimeDir, xdgCleanup := util.TestDir()
+	defer xdgCleanup()
+
+	tmpDir, tmpCleanup := util.TestDir()
+	defer tmpCleanup()
+
+	uid := os.Getuid()
+
+	tests := []struct {
+		name          string
+		xdgRuntimeDir string
+		tmpdir        string
+		want          string
+	}{
+		{
+			name:          "prefer XDG_RUNTIME_DIR over TMPDIR",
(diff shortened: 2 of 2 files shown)


krader1961:
Doesn't build (hint: unused imports). Of more significant I don't understand why you only check that env var on Linux. The XDG spec is applicable to every UNIX-like platform including macOS.

vially:
> I don't understand why you only check that env var on Linux. The XDG spec is applicable to every UNIX-like platform including macOS.

I was under the impression that while some parts of the XDG spec might make sense on some other non-Linux systems (e.g.: `XDG_CONFIG_HOME`), the `XDG_RUNTIME_DIR` is only used on Linux. I may be wrong though as I don't have much experience with macOS and other unixes.

In any case I've changed the pull-request to apply to all unixes, not just Linux (as defined by the `// +build !windows,!plan9` build flag that was already present).

vially:
Any idea why the FreeBSD tests are failing? I don't currently have access to a FreeBSD machine to debug it, so any hints about a possible cause are appreciated.

krader1961:
@vially, Click on the [details] link. You'll see that it's not failing due to anything in your PR. It looks like @xiaq recently setup this test environment on the Cirrus CI and it exposed some compatibility problems with pre-existing elvish code. The tests all pass for me on my FreeBSD 11 VM with your PR applied. The Cirrus CI is using FreeBSD 12 so its possible some work is needed to deal with recent updates to FreeBSD.

vially:
@krader1961 Great, thanks for looking into it!

pkg/program/shell/run_dir_unix.go
```diff
@@ -15,10 +15,32 @@ var (
 	errBadPermission = errors.New("bad permission")
 )
 
-// getSecureRunDir stats elvish-$uid under the default temp dir, creating it if
+// getSecureRunDir stats elvish under the XDG runtime dir, creating it if
```
xiaq, line 18: Make the comment as self-contained as possible: One cannot understand what "XDG runtime dir" exactly means without looking up the standard. Just say "the `XDG_RUNTIME_DIR` environment variable" and it's clear what it means.

See also comment below; that will also affect how to update this comment.

pkg/program/shell/run_dir_unix.go
```diff
@@ -15,10 +15,32 @@ var (
...
+// it doesn't yet exist, and return the directory name.
+// If XDG runtime dir is not set it falls back to elvish-$uid under the
+// default temp dir.
+func getSecureRunDir() (string, error) {
```
xiaq, line 22: The `checkExclusiveAccess` call should also apply to $XDG_RUNTIME_DIR/elvish too.

In fact you can share the `Mkdir` call too. The only part that differs from $XDG_RUNTIME_DIR/elvish from $tmpdir/elvish is the path. There is no need to split the logic into multiple functions, either.

xiaq:
Please also rebase against HEAD to get accurate CI results.

xiaq:
This file is currently untested so I won't require you to add a new one, but it would be nice if you do.

pkg/program/shell/run_dir_unix.go
vially: I pushed a new version that's using the same function so it shares most of the code with the fallback temp directory code path.

The original version didn't apply the `checkExclusiveAccess` call because the [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables) provides some of these guarantees:
> The directory MUST be owned by the user, and he MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700.

But they're only mandated for the `XDG_RUNTIME_DIR` directory so I guess it makes sense to ensure them for `$XDG_RUNTIME_DIR/elvish` too.

pkg/program/shell/run_dir_unix.go
vially: I've reworded the comment in the latest push.

vially:
I've rebased against HEAD but apparently there's still a CI failure (although it's a different one now).

> This file is currently untested so I won't require you to add a new one, but it would be nice if you do.

I've added a basic test but it is a bit hacky because it needs to set the `XDG_RUNTIME_DIR` environment which is global and will affect other tests.

I think the current form of this function's implementation makes it a bit hard to test mainly due to its reliance on environment variables (`XDG_RUNTIME_DIR` and `TMPDIR`) and because it makes changes to the file-system which will need to be cleaned up in the test.

vially:
I've pushed a new commit that splits the logic for calculating the path and the logic that stats the directory into two different functions in order to be able to test them separately. The tests still don't look very pretty due to the environment variables, but they should provide slightly more coverage.

Let me know if you have any suggests for how to improve them.

pkg/program/shell/run_dir_unix.go
```diff
@@ -31,7 +32,19 @@ func getSecureRunDir() (string, error) {
 		return "", err
 	}
 
-	return runDir, checkExclusiveAccess(info, uid)
+	return runDir, checkExclusiveAccess(info, os.Getuid())
+}
+
+// getRunDirPath returns a path that's used by elvish to store runtime files.
```
xiaq, line 38: Nitpick: Unless referring to the command name or file path, spell elvish with a capital E.

pkg/program/shell/run_dir_unix_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+				if runDir != wantRunDir {
+					t.Errorf("Got run dir %v, want %v", runDir, wantRunDir)
+				}
+			})
+		})
+	}
+}
+
+func withEnvVarOverrides(envOverrides map[string]string, fn func()) {
```
xiaq, line 79: This function is better written with the following signature:

```go
func withEnvVarOverrides(overrides map[string]string) func()
```

And used like this:

```go
restore := withTempOverrides(...)
defer restore()
```

In fact, this is exactly the style used by `util.TestDir` you are using in one of the tests. It suits the style of Go much better.

Also, there is already a testing helper, [`util.TempEnv`](https://github.com/elves/elvish/blob/7325c89be4131a8fc22e4d405f474f31ef56f11b/pkg/util/temp_env.go#L5), that does something very similar, except that it only supports one environment variable, and treats non-existing variables as if they are empty.

I'd move this function to the `util` package, call it `WithTempEnvs`, and implement `WithTempEnv` as just a thin wrapper around it. But that means you also need to have unit tests against `WithTempEnvs` too, which will make the PR bigger and slower to merge, so I'm perfectly happy if you act on my suggestion for the signature, and keep this helper in this test file for now.

pkg/program/shell/run_dir_unix_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+// RunDirPathTest is a test case to be used in getRunDirPath.
+type RunDirPathTest struct {
+	Name          string
+	XDGRuntimeDir string
+	Tmpdir        string
+	Want          string
+}
+
+func TestGetSecureRunDir_UseXDGWhenPresent(t *testing.T) {
```
xiaq, line 22: You do not need to test `getRunDirPath`. You can test `getSecureRunDir` if you do the following:

* Create **two** temporary directories and use them as overrides for **both** of `XDG_RUNTIME_DIR` and `TMPDIR`.

* Since the expected run dir now depends on the path of the temporary directories, move the test table inside the function, after the creation of the temporary directories.

There are 4 cases to test:

* XDG_RUNTIME_DIR set, TMPDIR set
* XDG_RUNTIME_DIR set, TMPDIR unset
* XDG_RUNTIME_DIR unset, TMPDIR set
* XDG_RUNTIME_DIR unset, TMPDIR unset

However, the last case will use the system temporary directory `/tmp`, which will likely already have an existing Elvish run dir, so you can skip that case.

The unit test should also check if the run dir has the correct permission.

pkg/program/shell/run_dir_unix_test.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/elves/elvish/pkg/util"
+)
+
+// RunDirPathTest is a test case to be used in getRunDirPath.
```
xiaq, line 14: It is customary to use an anonymous struct (if possible) with unexported fields in test data tables.

pkg/program/shell/run_dir_unix.go
xiaq: Do not trust the operating system to implement the XDG spec correctly.

In fact, do not trust any spec to be implemented correctly at all, unless you can check all the implementations.

Most of the time, Elvish only depends on functionalities of the *kernel*, so it's fairly easy to verify. But in this case, Elvish is depending on the *userland*. There are only a few major BSD implementations out there, but thousands of Linux distributions. The latter is impossible to check exhaustively.

xiaq:
I pushed some more fixes to the CI config - sorry for the confusing error message. Please rebase again to get accurate CI results (and test coverage!)

vially:
@xiaq I've rebased and pushed a new version that addresses your feedback. Let me know if I've missed something.

xiaq:
The code looks great, thanks!

Hmm, just realized that this change will cause problems for users upgrading from a previous version if they have $XDG_RUNTIME_DIR set. The user may have a (old-version) daemon listening on $TMPDIR/elvish-$uid/sock; but since the new Elvish version assumes that the rundir is $XDG_RUNTIME_DIR, it will spin up a new server there. This will fail because boltdb does not allow multiple processes accessing the same database.

Instead I propose a slightly more complex algorithm: `getSecureRunDir` should first check if both of the potential run directories already exist, and use an existing one, if its permission meets the condition. This means that the new Elvish instance will continue to talk to the daemon using the old run dir. When the user reboots the machine the next time (assuming that the run directory is ephemeral), Elvish will start using the new run directory.

pkg/program/shell/run_dir_unix_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"syscall"
+	"testing"
+
+	"github.com/elves/elvish/pkg/util"
+)
+
+func TestGetSecureRunDir_UseXDGWhenPresent(t *testing.T) {
```
xiaq, line 15: Update the name of this test function to just `TestGetSecureRunDir`.

xiaq:
@vially interested in continuing to work on this? I may be doing some refactors in this package, and if you are planning to work on it soon I can hold off those changes.

It's completely fine if you don't have the capacity to work on this now; in that case I can carry on the work here.

vially:
@xiaq Sorry for the delay on this. I got caught up with some other things and didn't find the time to get back to this until now.

I just pushed a new version which implements the new algorithm that you've suggested (good catch on that by the way).

Unfortunately now the table tests no longer cover all the possible cases so I added a few more tests to try to make up for it but it's still far from being thorough. At this point I'm not sure whether it makes sense to try to cover *everything* with table driven tests for the `getSecureRunDir` or whether it would be better to split it up in smaller functions that could more easily be tested individually. Let me know your thoughts on this.

In any case, feel free to make any changes to this pull-request as needed or just ignore it completely. I don't think it's worth postponing other work in this package for this. I would still like to see it merged at some point but if we can't find a good solution now we can always come back to it later.

xiaq:
@vially Thank you!

You can definitely still cover all the possible overall code paths of `getSecureRunDir` if you make the test just a bit more complicated.

There are also a couple of nitpicks I still have, but I'll merge this PR now and address those myself, as I am already happy with the behavior of the code and it's perhaps better to save a few more review round-trips.

xiaq:
@vially FYI, I simplified the test in 3a976ef1c2e4f8c16e8c45e2b2bbafcafe985595; you can find the current version at https://github.com/elves/elvish/blob/3a976ef1c2e4f8c16e8c45e2b2bbafcafe985595/pkg/program/shell/run_dir_unix_test.go.

I abandoned the table-based approach; instead, I used a pair of helper functions (setup and test) to abstract their commonalities and kept each test case as its own function.

vially:
:+1: Nice! It does look more readable this way.

PR #957, opened 2020-04-05
Fix the `unix` namespace documentation

krader1961:
I noticed that `make style` changed the line wrapping. Which in turn
caused me to notice I had failed to correct the description about the
ability to executed `use unix` on non-UNIX platforms.

### DIFF
--- a/website/ref/unix.md
+++ b/website/ref/unix.md
@@ -2,9 +2,11 @@
 
 # Introduction
 
-The `unix:` module provides access to features that only make sense
-on UNIX compatible operating systems such as Linux, FreeBSD, macOS,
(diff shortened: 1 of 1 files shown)


xiaq:
Two nitpicks that I'll fix myself:

* Prefer the more vague "UNIX-like" to "UNIX-compatible". The latter can be taken to mean [SUS compliance](https://en.wikipedia.org/wiki/Single_UNIX_Specification) or POSIX compliance, both of which have very concrete meanings. Elvish's `unix:` namespace is based on the commonality of the UNIX-like OS supported by Elvish, which is not a formal standard, so it's better to keep it vague.

* Using "such as" and "etc" in the same sentence is redundant. Prefer to use just "such as".

PR #960, opened 2020-04-11
Fix two typos in my previous change

krader1961:


pkg/eval/math/math.go
```diff
@@ -121,7 +121,7 @@ import (
 //elvdoc:fn is-nan
 //
 // ```elvish
-// math:is-nan $number $sign
+// math:is-nan $number
```
xiaq, line 124: Document the `&sign` option.

krader1961: The point of this change is to remove the explicit `$sign` argument, as well as the implicit `&sign` option, from the documentation since neither is relevant, or valid, for arguments passed to `math:is-nan`. The original line was a copy/paste mistake from the `math:is-inf` implementation.

krader1961: Are you suggesting that `math:is-nan` explicitly support a `&sign` option? Even though that option is meaningless with respect to NaN values?

xiaq: Ah, I mistook this for `is-inf`.

PR #961, opened 2020-04-11
Fix formatting of the `styled` documentation

krader1961:


pkg/eval/builtin_fn_styled.go
```diff
@@ -48,35 +48,35 @@ var errStyledSegmentArgType = errors.New("argument to styled-segment must be a s
...
+//    prefix.
 //
 //     -   Any color from the xterm 256-color palette, as `colorX` (such as
-// `color12`)
+//    `color12`).
 //
 //     -   A 24-bit RGB color, as `#RRGGBB`, such as `#778899`.
 //
-// -   A color name prefixed by `bg-` to set the background color
+//     -   A color name prefixed by `bg-` to set the background color.
```
xiaq, line 73: This is not a child of the "color name" bullet, but on the same level with it, so do not indent it.

krader1961: Yeah, I had already fixed that in my local repo but failed to force push the correction. Done.

PR #963, opened 2020-04-13
Implement hyperbolic trigonometric math functions

krader1961:
Related #727

### DIFF
--- a/pkg/eval/math/math.go
+++ b/pkg/eval/math/math.go
@@ -76,6 +76,19 @@ import (
 // ▶ (float64 -1)
 // ```
 
+//elvdoc:fn cosh
+//
+// ```elvish
+// math:cosh $number
+// ```
+//
+// Computes the hyperbolic cosine of `$number`. Example:
+//
+// ```elvish-transcript
+// ~> math:cosh 0
+// ▶ (float64 1)
+// ```
+
 //elvdoc:fn floor
 //
 // ```elvish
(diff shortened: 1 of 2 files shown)


pkg/eval/math/math.go
```diff
@@ -76,6 +76,22 @@ import (
 // ▶ (float64 -1)
 // ```
 
+//elvdoc:fn cosh
+//
+// ```elvish
+// math:cosh $number
+// ```
+//
+// Computes the hyperbolic cosine of `$number` in units of radians (not degrees).
```
xiaq, line 85: My knowledge of hyperbolic trigonometric functions is a bit rusty, but IIRC they are unrelated to geometric angles, so the note about the unit does not apply.

pkg/eval/math/math.go
```diff
@@ -76,6 +76,22 @@ import (
...
+// Examples:
+//
+// ```elvish-transcript
+// ~> math:cosh 0
+// ▶ (float64 1)
+// ~> math:cosh $math:pi
```
xiaq, line 91: This is not a particularly good example since pi is not special in hyperbolic trigonometric functions.

pkg/eval/math/math.go
krader1961: Yes, you're right. I haven't thought about (or used) these functions since high school.

pkg/eval/math/math.go
krader1961: Good point, that was just copy/paste without much thought about what would constitute good examples for these functions. I'll follow the Go math documentation here.

Issue #318, opened 2017-02-02
glob / wildcard matching issue


muesli:
```
  $ touch a                                                                                                                                                                           
  $ touch b                                                                                                                                                                           
  $ touch c                                                                                                                                                                           

  $ ls -l                                                                                                                                                                             
  total 0
  -rw-r--r-- 1 muesli muesli 0 Feb  2 11:40 a
  -rw-r--r-- 1 muesli muesli 0 Feb  2 11:40 b
  -rw-r--r-- 1 muesli muesli 0 Feb  2 11:40 c

  $ ls bar*                                                                                                                                                                           
  a  b  c
```

As you can see the non-matching 'bar*' glob essentially translates to an empty string and "ls" is executed. This is potentially dangerous.

xiaq:
We can have wildcards that expand to nothing throw an error by default, and add a `nothing-ok` modifier (cf #298) to suppress this failing behavior.

PS. @muesli to be precise, your `bar*` did not evaluate to an empty string (that would be `ls ""`) but rather nothing (giving `ls` no argument at all).

xiaq:
Spec for new (just implemented) behavior: https://github.com/elves/elvish/wiki/Language-Specification#wildcards

PR #967, opened 2020-04-15
Implement math:min and math:max

### LINKED ISSUE #727: Pour go's math functions into a `math:` module?

krader1961:
Resolves #727

### DIFF
--- a/pkg/eval/math/math.go
+++ b/pkg/eval/math/math.go
@@ -285,6 +285,50 @@ import (
 // ▶ (float64 NaN)
 // ```
 
+//elvdoc:fn max
+//
+// ```elvish
+// math:max $number...
+// ```
+//
+// Outputs the maximum number in the arguments. If there are no arguments
+// an exception is thrown. If any number is NaN then NaN is output.
+//
+// Examples:
+//
+// ```elvish-transcript
+// ~> math:max
+// ▶ ?(WTF)
+// ~> math:max 3
+// ▶ (float 3)
+// ~> math:max 3 5 2
+// ▶ (float 5)
+// ~> range 100 | math:max (all)
+// ▶ (float 99)
+// ```
+
+//elvdoc:fn min
+//
+// ```elvish
+// math:min $number...
+// ```
+//
+// Outputs the minimum number in the arguments. If there are no arguments
+// an exception is thrown. If any number is NaN then NaN is output.
+//
+// Examples:
+//
+// ```elvish-transcript
+// ~> math:min
+// ▶ ?(WTF)
+// ~> math:min 3
+// ▶ (float 3)
+// ~> math:min 3 5 2
+// ▶ (float 2)
+// ~> range 100 | math:min (all)
+// ▶ (float 0)
+// ```
+
 //elvdoc:fn pow
 //
 // ```elvish
@@ -462,6 +506,8 @@ var fns = map[string]interface{}{
 	"log":           math.Log,
 	"log10":         math.Log10,
 	"log2":          math.Log2,
+	"max":           max,
+	"min":           min,
 	"pow":           math.Pow,
 	"pow10":         math.Pow10,
 	"round":         math.Round,
@@ -480,4 +526,18 @@ func (opts *isInfOpts) SetDefaultOptions() { opts.Sign = 0 }
 
 func isInf(opts isInfOpts, arg float64) bool {
 	return math.IsInf(arg, opts.Sign)
+}
+
+func max(fm *eval.Frame, num float64, nums ...float64) float64 {
+	for i := 0; i < len(nums); i++ {
+		num = math.Max(num, nums[i])
+	}
+	return num
+}
+
+func min(fm *eval.Frame, num float64, nums ...float64) float64 {
+	for i := 0; i < len(nums); i++ {
+		num = math.Min(num, nums[i])
+	}
+	return num
 }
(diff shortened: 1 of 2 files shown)


xiaq:
I prefer these functions to take variadic argument, in analogy with arithmetic functions like `+`.

I prefer these functions to throw an error when given no arguments. Elvish may get a Scheme-like numerical tower in future, and in such a system max and min should require at least one argument ([Racket's max](https://docs.racket-lang.org/reference/generic-numbers.html#%28def._%28%28quote._~23~25kernel%29._max%29%29) for example), which is is a consequence of two things:

* Functions that work for both inexact and exact numbers should only return inexact numbers when some of the arguments are inexact.

* +Inf and -Inf do not exist in the set of exact numbers.

krader1961:
> I prefer these functions to take variadic argument, ...

That is my preference as well. However, it is not compatible with the `type Inputs` idiom that makes it easy to handle CLI args or input from the value stream. Consider this example:

```
> { put 1; put 2 } | +
▶ (float64 0)
```

You have to do this instead:

```
> { put 1; put 2 } | + (all)
▶ (float64 3)
```

This behavioral difference between variadic and "input list" commands has bothered me for quite a while. I can't find the issue but the current "input list" behavior was predicated on protecting against empty wildcard expansion. For example, `a_cmd *` which would consume its byte/value input if the glob expanded to nothing. The argument for the current "input list" behavior was that example should be written as `a_cmd [*]` to guard against the glob/wildcard expression not matching any files by resulting in an empty list that would keep `a_cmd` from reading from its stdin.

There is issue #318 but that isn't the one I recall. There was a different issue that argued for `a_cmd [*]` rather than `a_cmd *` because the glob might expand to nothing, thus causing `a_cmd` to read from its stdin.

That argument seems no longer relevant since a wildcard (i.e., glob) will throw an exception if it does not include the `[nomatch-ok]` modifier. In other words, the user has to explicitly say they understand the risk of using a wildcard that expands to nothing.

Also, consider functions like `str:join` and `joins`. Why does `joins : a b` throw a "arity mismatch" while `{ put a; put b } | joins :` works as expected? Why doesn't `joins` or `str:join` also take a variadic argument?

krader1961:
The question about the behavior of an empty sequence should be moved to issue #727. Or, perhaps, a new issue since the #727 has been resolved, more or less, regardless of this change.

xiaq:
See #727 and #991.

xiaq:
While #991 is open, my preference remains making `math:min` and `math:max` variadic functions.

krader1961:
> While #991 is open, my preference remains making math:min and math:max variadic functions.

That has me scratching my head; i.e., perplexes me. How often would you do `math:min 3 2 1 0`? Using the variadic form makes it much harder to use these functions in a value pipeline. Which is likely to be the common case. Even when selecting the min from a list I would expect to write `math:min $list` rather than `math:min $@list`.

krader1961:
Not to mention expecting this to work:

```
{ put 1; put -1 } | math:min
```

Rather than having to do this:

```
{ put 1; put -1 } | math:min (all)
```

xiaq:
Taking the smaller of two numbers is common; `math:min $a $b` feels more intuitive.

There isn't a good reason other than familiarity for `min` to work either way; most languages have `min` taking either two or variadic arguments; [C++](http://www.cplusplus.com/reference/algorithm/min/), [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/min), [R6RS](http://www.r6rs.org/final/html/r6rs/r6rs-Z-H-14.html#node_sec_11.7.4.3).

krader1961:
Okay, variadic form it is. Please take another look.

pkg/eval/math/math.go
```diff
@@ -285,6 +285,50 @@ import (
...
+// ```
+//
+// Outputs the maximum number in the arguments. If there are no arguments
+// an exception is thrown. If any number is NaN then NaN is output.
+//
+// Examples:
+//
+// ```elvish-transcript
+// ~> math:max
+// ▶ ?(WTF)
```
xiaq, line 301: Replace this with an actual transcript of the exception.

pkg/eval/math/math.go
```diff
@@ -285,6 +285,50 @@ import (
...
+// ```
+//
+// Outputs the minimum number in the arguments. If there are no arguments
+// an exception is thrown. If any number is NaN then NaN is output.
+//
+// Examples:
+//
+// ```elvish-transcript
+// ~> math:min
+// ▶ ?(WTF)
```
xiaq, line 323: Replace this with an actual transcript of the exception.

pkg/eval/math/math.go
```diff
@@ -481,3 +527,17 @@ func (opts *isInfOpts) SetDefaultOptions() { opts.Sign = 0 }
 func isInf(opts isInfOpts, arg float64) bool {
 	return math.IsInf(arg, opts.Sign)
 }
+
+func max(fm *eval.Frame, num float64, nums ...float64) float64 {
```
xiaq, line 531: Remove the `Frame` argument.

pkg/eval/math/math.go
krader1961: Argh! Yep, I meant that as a placeholder until I confirmed the exact spelling of the exception. Fixed.

pkg/eval/math/math.go
krader1961: Yep, no longer needed.

PR #968, opened 2020-04-15
Simplify count implementation

### LINKED ISSUE #966: Error consistency: `count` returns a unique error for an invalid number of arguments
I started implementing the `math:max` command by copying the builtin `count` implementation. Which resulted in me noticing that is the only builtin that returns `errors.New("want 0 or 1 argument")` if it gets more than one argument. Similar commands return `eval.ErrArgs` or `arity mismatch`. For example,

```
> one a b
Exception: arity mismatch: arguments here must be 0 to 1 values, but is 2 values
[tty], line 1: one a b
```

It's not clear why `count` has a unique error. This looks like a historical artifact that should be normalized to be consistent with other commands. It seems to me `func count` should be using `eval.Inputs` rather than an explicit variadic arg. Yes? No?

krader1961:
Update the count command to use the same `Inputs` based implementation
as every other command that iterates over an optional iterable argument.

Resolves #966

### DIFF
--- a/pkg/eval/builtin_fn_container.go
+++ b/pkg/eval/builtin_fn_container.go
@@ -526,31 +526,11 @@ func hasKey(container, key interface{}) bool {
 	return vals.HasKey(container, key)
 }
 
-func count(fm *Frame, args ...interface{}) (int, error) {
+func count(fm *Frame, inputs Inputs) (int, error) {
 	var n int
-	switch len(args) {
-	case 0:
-		// Count inputs.
-		fm.IterateInputs(func(interface{}) {
-			n++
-		})
-	case 1:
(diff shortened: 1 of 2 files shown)


xiaq:
The current implementation is O(1) when the argument is supported by `vals.Len`.

krader1961:
Good point. Any objection to adding a comment explaining why this command deviates from the usual `Inputs` pattern?

krader1961:
Not to mention fixing the error message? And adding some tests to verify that a non-iterable value is handled correctly?

xiaq:
A comment is definitely useful.

xiaq:
Let's discuss on #966.

PR #972, opened 2020-04-18
Implement str:join

krader1961:


krader1961:
Gah! I just noticed that I didn't include documentation for this function when I refactored it from an earlier commit that implemented both `str:join` and `str:split`. Wait a moment before reviewing this change.

xiaq:
I prefer the current behavior of the `joins` builtin; commands that clearly work with strings should not convert values to strings, but rather, require them to be strings.

I just pushed 3e9d5e9e60362c0e954039aed19a15c31b67b713 which mostly copied the the implementation of the builtin versions, and deprecates the builtin versions.

krader1961:
> I prefer the current behavior of the joins builtin; commands that clearly work with strings should not convert values to strings, but rather, require them to be strings.

@xiaq, While that is reasonable for objects like maps and lists it means the user has to explicitly coerce `float64` objects. That is suboptimal if we want to minimize the friction of using binary numbers rather than their string form. I strongly feel that if you don't want to automatically stringify all data types passed to `str:join` and `str:split` these functions should implicitly stringify the `float64` type.

xiaq:
I am cautious about "weak typing". You may (correctly) argue that Elvish already has some weak typing when it comes to numbers and strings, but there is a very simple rule: any numeric commands that only operates on numbers also accept strings that parse to numbers. The rule is simple and universally useful.

If the choice to accept numbers in `str:join` is part of a simple, universally useful rule, then that's OK; but I don't see a rule emerging here, so the decision to allow numbers seems arbitrary. I usually take discipline over convenience.

FWIW, other dynamic languages don't have a consensus:

* `join` requires strings: Python, Racket
* `join` accepts numbers: JavaScript, Ruby

Stylistically Elvish is closer to the first group, so following their decision is a safef design choice.

PR #973, opened 2020-04-18
Implement str:split

krader1961:


xiaq:
I just pushed 3e9d5e9e60362c0e954039aed19a15c31b67b713 which mostly copied the the implementation of the builtin versions, and deprecates the builtin versions.

PR #977, opened 2020-04-20
Document `base` command

krader1961:


pkg/eval/builtin_fn_str.go
```diff
@@ -97,7 +97,29 @@ var ErrInputOfEawkMustBeString = errors.New("input of eawk must be string")
...
 
-// TODO(xiaq): Document "base".
+//elvdoc:fn base
+//
+// ```elvish
+// base $base $number...
+// ```
+//
+// Outputs a string for each `$number` written in `$base`. The `$base` must be
+// between 2 and 36, inclusive. Examples,
```
xiaq, line 107: Use a comma, `Examples:`

krader1961: Fixed, and rebased on current master.

PR #980, opened 2020-04-22
Change and document how abbreviation expansion works

### LINKED ISSUE #947: Document $edit:abbr
The CLI abbreviation expansion feature isn't documented. You can create abbreviations via statements such as `edit:abbr['l '] = 'less '` or `edit:abbr[gcp] = 'git cherry-pick -x'`. Abbreviations are sometimes preferable to a function of the same name for several reasons:

1) You might want to modify the expanded text before executing it.

1) You want the expanded text to appear in the command history.

1) You want a visual reminder of what will be executed by typing a cryptic shortcut like `gcm`.

1) You only want the shortcut to be available at an interactive prompt, not as a function that can be used in elvish scripts, to avoid potential problems.

The existing behavior isn't optimal, IMHO, but let's start by documenting it so people are aware it exists and know how to use it.

krader1961:
Document interactive abbreviation expansions.

Change the implementation in two ways:

1) Skip an abbreviation if it begins with an alphanum and the previous
   char in the insert buffer is also an alphanum or there is no
   preceding char.

2) Only perform an expansion when a space, pipe, newline, or punctuation
   char is typed rather than when the last char of the abbreviation is
   typed. This makes the behavior more consistent with editors like Vim
   and Emacs and friendlier to use.

Resolves #947

### DIFF
--- a/pkg/cli/app_spec.go
+++ b/pkg/cli/app_spec.go
@@ -20,6 +20,8 @@ type AppSpec struct {
 	Abbreviations  func(f func(abbr, full string))
 	QuotePaste     func() bool
 
+	SmallWordAbbreviations func(f func(abbr, full string))
+
 	CodeAreaState CodeAreaState
 	State         State
 }
--- a/pkg/cli/codearea.go
+++ b/pkg/cli/codearea.go
@@ -38,7 +38,8 @@ type CodeAreaSpec struct {
 	// A function that calls the callback with string pairs for abbreviations
 	// and their expansions. If this function is not given, the Widget does not
 	// expand any abbreviations.
-	Abbreviations func(f func(abbr, full string))
+	Abbreviations          func(f func(abbr, full string))
+	SmallWordAbbreviations func(f func(abbr, full string))
 	// A function that returns whether pasted texts (from bracketed pastes)
 	// should be quoted. If this function is not given, the Widget defaults to
 	// not quoting pasted texts.
@@ -125,6 +126,9 @@ func NewCodeArea(spec CodeAreaSpec) CodeArea {
 	if spec.Abbreviations == nil {
 		spec.Abbreviations = func(func(a, f string)) {}
 	}
+	if spec.SmallWordAbbreviations == nil {
+		spec.SmallWordAbbreviations = func(func(a, f string)) {}
+	}
 	if spec.QuotePaste == nil {
 		spec.QuotePaste = func() bool { return false }
 	}
@@ -196,6 +200,71 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
 	return true
 }
 
+// Try to expand a simple abbreviation that ignores word boundaries.
+func (w *codeArea) expandSimpleAbbr() {
+	var abbr, full string
+	// Find the longest matching abbreviation.
+	w.Abbreviations(func(a, f string) {
+		if strings.HasSuffix(w.inserts, a) && len(a) > len(abbr) {
+			abbr, full = a, f
+		}
+	})
+	if len(abbr) > 0 {
+		c := &w.State.Buffer
+		*c = CodeBuffer{
+			Content: c.Content[:c.Dot-len(abbr)] + full + c.Content[c.Dot:],
+			Dot:     c.Dot - len(abbr) + len(full),
+		}
+		w.resetInserts()
+	}
+}
+
+// Try to expand a small word abbreviation.
+func (w *codeArea) expandSmallWordAbbr(triggerRune rune, triggerLen int) {
+	// We only do small word expansions at the end of the line and only if
+	// there are characters in the insert buffer other than the trigger rune.
+	c := &w.State.Buffer
+	if len(w.inserts) <= triggerLen || c.Dot < len(c.Content) {
+		return
+	}
+
+	inserts := w.inserts[:len(w.inserts)-triggerLen] // ignore abbr trigger rune
+	var abbr, full string
+	// Find the longest matching abbreviation.
+	w.SmallWordAbbreviations(func(a, f string) {
+		if len(a) >= len(w.inserts) || len(a) < len(abbr) {
+			return // check the next abbreviation
+		}
+
+		// Verify the abbreviation is in the insert buffer.
+		if !strings.HasSuffix(inserts, a) {
+			return
+		}
+		// Verify the trigger rune creates a word boundary.
+		r2, _ := utf8.DecodeLastRuneInString(a)
+		if CategorizeSmallWord(triggerRune) == CategorizeSmallWord(r2) {
+			return
+		}
+		// Verify the rune preceding the abbreviation, if any, creates a word boundary.
+		if len(c.Content) > len(a)+triggerLen {
+			r1, _ := utf8.DecodeLastRuneInString(c.Content[:len(c.Content)-len(a)-triggerLen])
+			r2, _ = utf8.DecodeRuneInString(a)
+			if CategorizeSmallWord(r1) == CategorizeSmallWord(r2) {
+				return
+			}
+		}
+
+		abbr, full = a, f
+	})
+	if len(abbr) > 0 {
+		*c = CodeBuffer{
+			Content: c.Content[:c.Dot-len(abbr)-triggerLen] + full + string(triggerRune),
+			Dot:     c.Dot - len(abbr) + len(full),
+		}
+		w.resetInserts()
+	}
+}
+
 func (w *codeArea) handleKeyEvent(key ui.Key) bool {
 	isFuncKey := key.Mod != 0 || key.Rune < 0
 	if w.pasting {
@@ -207,9 +276,11 @@ func (w *codeArea) handleKeyEvent(key ui.Key) bool {
 		}
 		return true
 	}
+
 	if w.OverlayHandler.Handle(term.KeyEvent(key)) {
 		return true
 	}
+
 	// We only implement essential keybindings here. Other keybindings can be
 	// added via handler overlays.
 	switch key {
@@ -245,21 +316,25 @@ func (w *codeArea) handleKeyEvent(key ui.Key) bool {
 		w.State.Buffer.InsertAtDot(s)
 		w.inserts += s
 		w.lastCodeBuffer = w.State.Buffer
-		var abbr, full string
-		// Try to expand an abbreviation, preferring the longest one
-		w.Abbreviations(func(a, f string) {
-			if strings.HasSuffix(w.inserts, a) && len(a) > len(abbr) {
-				abbr, full = a, f
-			}
-		})
-		if len(abbr) > 0 {
-			c := &w.State.Buffer
-			*c = CodeBuffer{
-				Content: c.Content[:c.Dot-len(abbr)] + full + c.Content[c.Dot:],
-				Dot:     c.Dot - len(abbr) + len(full),
-			}
-			w.resetInserts()
-		}
+		w.expandSimpleAbbr()
+		w.expandSmallWordAbbr(key.Rune, len(s))
 		return true
+	}
+}
+
+// Determine if the rune is an alphanumeric character.
+func IsAlnum(r rune) bool {
+	return unicode.IsLetter(r) || unicode.IsNumber(r)
+}
+
+// Determine if the rune is whitespace, alphanum, or something else.
+func CategorizeSmallWord(r rune) int {
+	switch {
+	case unicode.IsSpace(r):
+		return 0
+	case IsAlnum(r):
+		return 1
+	default:
+		return 2
 	}
 }
--- a/pkg/cli/codearea_test.go
+++ b/pkg/cli/codearea_test.go
@@ -294,6 +294,67 @@ var codeAreaHandleTests = []HandleTest{
 		WantNewState: CodeAreaState{Buffer: CodeBuffer{Content: "dn", Dot: 2}},
 	},
 	{
+		Name: "abbreviation expansion",
+		Given: NewCodeArea(CodeAreaSpec{
+			Abbreviations: func(f func(abbr, full string)) {
+				f("||", " | less")
+			},
+		}),
+		Events:       []term.Event{term.K('x'), term.K('|'), term.K('|')},
+		WantNewState: CodeAreaState{Buffer: CodeBuffer{Content: "x | less", Dot: 8}},
+	},
+	{
+		Name: "abbreviation expansion",
+		Given: NewCodeArea(CodeAreaSpec{
+			Abbreviations: func(f func(abbr, full string)) {
+				f("||", " | less")
+			},
+		}),
+		Events:       []term.Event{term.K('{'), term.K('e'), term.K('c'), term.K('h'), term.K('o'), term.K(' '), term.K('x'), term.K('}'), term.K('|'), term.K('|')},
+		WantNewState: CodeAreaState{Buffer: CodeBuffer{Content: "{echo x} | less", Dot: 15}},
+	},
+	{
+		Name: "small word abbreviation expansion space trigger",
+		Given: NewCodeArea(CodeAreaSpec{
+			SmallWordAbbreviations: func(f func(abbr, full string)) {
+				f("eh", "echo hello")
+			},
+		}),
+		Events:       []term.Event{term.K('e'), term.K('h'), term.K(' ')},
+		WantNewState: CodeAreaState{Buffer: CodeBuffer{Content: "echo hello ", Dot: 11}},
+	},
+	{
+		Name: "small word abbreviation expansion non-space trigger",
+		Given: NewCodeArea(CodeAreaSpec{
+			SmallWordAbbreviations: func(f func(abbr, full string)) {
+				f("h", "hello")
+			},
+		}),
+		Events:       []term.Event{term.K('x'), term.K('['), term.K('h'), term.K(']')},
+		WantNewState: CodeAreaState{Buffer: CodeBuffer{Content: "x[hello]", Dot: 8}},
+	},
+	{
+		Name: "small word abbreviation expansion preceding char invalid",
+		Given: NewCodeArea(CodeAreaSpec{
+			SmallWordAbbreviations: func(f func(abbr, full string)) {
+				f("h", "hello")
+			},
+		}),
+		Events:       []term.Event{term.K('g'), term.K('h'), term.K(' ')},
+		WantNewState: CodeAreaState{Buffer: CodeBuffer{Content: "gh ", Dot: 3}},
+	},
+	{
+		Name: "small word abbreviation expansion after backspace preceding char invalid",
+		Given: NewCodeArea(CodeAreaSpec{
+			SmallWordAbbreviations: func(f func(abbr, full string)) {
+				f("h", "hello")
+			},
+		}),
+		Events: []term.Event{term.K('g'), term.K(' '), term.K(ui.Backspace),
+			term.K('h'), term.K(' ')},
+		WantNewState: CodeAreaState{Buffer: CodeBuffer{Content: "gh ", Dot: 3}},
+	},
+	{
 		Name: "overlay handler",
 		Given: codeAreaWithOverlay(CodeAreaSpec{}, func(w *codeArea) Handler {
 			return MapHandler{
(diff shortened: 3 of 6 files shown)


krader1961:
I expect this change to be contentious. But please try it before commenting. The old behavior caused problems with several of the abbreviations I migrated from my fish config. For example, I had a `hr` abbreviation. The old elvish behavior meant that typing `chr` caused the `hr` expansion to occur. This implementation does not do that which is far less surprising. Now I have to type `hr` at the start of the command line or after some punctuation, or a pipe symbol, to trigger the expansion.

I've been using the core logic in this PR for a couple of days and am no longer annoyed, or surprised, by the behavior of elvish abbreviations. Which is not to say the new behavior is going to satisfy everyone. But I think it is an improvement over the simpler implementation this replaces.

xiaq:
I recognize your use case, but the rules here are too complex.

A much simpler way to solve this is to introduce another type of "word abbreviations" that is only expanded when the abbreviation is preceded by and followed by word boundaries, and your `hr` abbreviation can be configured as a word abbreviation instead:

```
edit:word-abbr[hr] = ...
```

krader1961:
> I recognize your use case, but the rules here are too complex.

Fair enough. I've followed your suggestion and instead of changing the existing `$edit:abbr` behavior introduced a `$edit:word-abbr`.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,56 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
 	return true
 }
 
+// Try to expand a non-word abbreviation.
+func (w *codeArea) abbrExpansion() {
```
xiaq, line 204: Give function verbial names: `expandAbbr`.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,56 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
...
+		*c = CodeBuffer{
+			Content: c.Content[:c.Dot-len(abbr)] + full + c.Content[c.Dot:],
+			Dot:     c.Dot - len(abbr) + len(full),
+		}
+		w.resetInserts()
+	}
+}
+
+// Try to expand a word abbreviation.
+func (w *codeArea) wordAbbrExpansion() {
```
xiaq, line 224: Similarly, `expandWordAbbr`.

pkg/cli/codearea.go
```diff
@@ -207,9 +261,18 @@ func (w *codeArea) handleKeyEvent(key ui.Key) bool {
 		}
 		return true
 	}
+
+	// Perform abbreviation expansion if a space, newline, pipe or punctuation
```
xiaq, line 265: A reader would expect the code to expand abbreviations and word abbreviations to be at the same place, but they aren't.

It's worth spending some time refactoring the code so that it is the case.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,56 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
...
+
+	ilen := len(w.inserts)
+	w.WordAbbreviations(func(a, f string) bool {
+		alen := len(a)
+		if alen > ilen { // skip abbr since it can't match the insert buffer
+			return false // check the next abbr
+		}
+		if ilen > alen {
+			rune, _ := utf8.DecodeLastRuneInString(w.inserts[:ilen-alen])
+			if unicode.IsLetter(rune) || unicode.IsDigit(rune) {
```
xiaq, line 236: Elvish's editor has a concept of *word* that is a sequence of non-space characters (space is defined by [unicode.IsSpace](https://godoc.org/unicode#IsSpace)).

It also has a concept of *small word* that is a sequence of letters or numbers (as defined by unicode.IsLetter and unicode.IsNumber).

It would be useful to align the concept of "word abbreviations" to one of those.

krader1961: Small word is what we want here so I'll replace IsDigit() with IsNumber() and add an inline comment.

P.S., Should this use of `IsDigit()` be replaced with `IsNumber()`?

https://github.com/elves/elvish/blob/05156ea239d6c902f7ec10fe6ab74fd4a26c734d/cmd/examples/cli/main.go#L20

pkg/cli/codearea.go
krader1961: Done

pkg/cli/codearea.go
krader1961: Done

pkg/cli/codearea.go
krader1961: It's not clear that doing so would be an improvement. The reason for the current structure is due to treating a newline char as whitespace that triggers a word expansion. The case statements in the `switch key` block below won't be executed if, as currently happens, the newline is handled by the `if w.OverlayHandler.Handle(term.KeyEvent(key))` block.

I'm going to rebase master and address your other points. Then we can revisit this if you still feel strongly that unifying the two cases is both desirable and a net improvement.

krader1961:
@xiaq, What do you think about renaming `$edit:word-abbr` to `$edit:abbr-word`? The latter "reads" weird (at least for a native English speaker) but has the advantage of placing the two closely related definitions adjacent in the documentation.

krader1961:
@xiaq, Hold off on reviewing my most recent forced push. I'm thinking more about your comment regarding big versus small words. I'm pondering whether that should be expressed in a different fashion (e.g., not private to the "edit" package). Also, whether that definition should replace the more fine-grained test in `handleKeyEvent()`.

krader1961:
@xiaq, Please take another look.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,56 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
...
+	}
+}
+
+// Try to expand a small word abbreviation.
+func (w *codeArea) expandSmallWordAbbr() {
+	w.StateMutex.Lock()
+	defer w.StateMutex.Unlock()
+
+	ilen := len(w.inserts)
+	w.WordAbbreviations(func(a, f string) bool {
```
xiaq, line 229: Small word expansion should prefer the longest match too, so this should first walk through all pairs, find the longest match, and then do the expansion.

There is no need to make the callback return a `bool` after you make this change.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,56 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
...
+		}
+		if ilen > alen {
+			rune, _ := utf8.DecodeLastRuneInString(w.inserts[:ilen-alen])
+			if IsSmallWordRune(rune) { // the abbreviation doesn't match a small word
+				return false // check the next abbreviation
+			}
+		}
+		if !strings.HasSuffix(w.inserts, a) {
+			return false // check the next abbreviation
+		}
```
xiaq, line 242: If you move this to be beginning of the closure, there is no need for the length check.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,56 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
...
+		w.resetInserts()
+	}
+}
+
+// Try to expand a small word abbreviation.
+func (w *codeArea) expandSmallWordAbbr() {
+	w.StateMutex.Lock()
+	defer w.StateMutex.Unlock()
+
+	ilen := len(w.inserts)
```
xiaq, line 228: Prefer just writing out `len(w.inserts)`, it isn't so long.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,56 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
...
+	w.WordAbbreviations(func(a, f string) bool {
+		alen := len(a)
```
xiaq, line 230: Prefer just writing out `len(a)`, it isn't so long.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,56 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
...
+		if alen > ilen { // skip abbr since it can't match the insert buffer
+			return false // check the next abbreviation
+		}
+		if ilen > alen {
+			rune, _ := utf8.DecodeLastRuneInString(w.inserts[:ilen-alen])
```
xiaq, line 235: IIUC this only checks the *left border* of the abbreviation. The right border should be checked too, using the to-be-inserted character.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,56 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
...
+			if IsSmallWordRune(rune) { // the abbreviation doesn't match a small word
```
xiaq, line 236: There are two classes of small words: a run of alphanumeric characters (e.g. `abc`) is a small word, but so is a run of non-whitespace, non-alphanumeric characters (e.g. `*/*`), so there is no single definition for what a "small word rune" is; there are two symmetric small-word categories. See [this comment](https://github.com/elves/elvish/blob/1d85b7fced66688367096767d707cc992807add6/pkg/edit/builtins.go#L462] on how this all works.

The correct way to identify whether a word abbreviation should be expanded is to detect *word boundaries*, where the category of two adjacent runes are different. You should do something like:

```
if categorizeSmallWord(r1) != categorizeSmallWord(r2) {
  // hit
}
```

Detecting word boundaries is also useful. As a theoretical example, you can now define `&` as a small-word expansion that only expands when both sides of it are alphanumerical characters. You can even mix and define `|a` as a small-word expansion that only expands when the left side of it is an alphanumerical character, but the right side of it is not.

pkg/cli/codearea.go
```diff
@@ -207,9 +261,19 @@ func (w *codeArea) handleKeyEvent(key ui.Key) bool {
 		}
 		return true
 	}
+
+	// Perform small word abbreviation expansion if a non-small word char is
+	// typed. This mimics how abbreviations work in some editors. This needs
+	// to be done before the overlay handler test to ensure it happens on
+	// newline being typed.
```
xiaq, line 268: I'd rather expansions to *not* to be done when a newline is typed. Expansions are supposed to give users real-time feedback, giving them a chance to correct the expansion if that's not what they need. Expanding an abbreviation when the user types a newline means that the user never gets a chance to actually see what got expanded.

I'd rather live with (and force users to live with) the slight inconvenience of having to say, type a space, to see the expansion.

pkg/edit/insert_api.go
```diff
@@ -8,11 +8,53 @@ import (
...
+// mimics that of abbreviations in editors such as Vim and Emacs. Most notably
+// word abbreviations are expanded only when a non-small word character is
+// typed and the character preceding the abbreviation is not a small word
+// character.
+//
+// Note that the beginning and ending of a typed line of text is considered to
+// be a non-small word character. This means that word abbreviations expand
+// when typed at the start of a line and followed by a non-small word char or
+// when [enter] is pressed.
+//
```
xiaq, line 38: Document what happens when both small-word abbreviation expansion and (normal) abbreviation expansion are both eligible: the former takes precedence.

pkg/cli/app_spec.go
```diff
@@ -16,9 +16,10 @@ type AppSpec struct {
 	Prompt      Prompt
 	RPrompt     Prompt
 
-	OverlayHandler Handler
-	Abbreviations  func(f func(abbr, full string))
-	QuotePaste     func() bool
+	OverlayHandler    Handler
+	Abbreviations     func(f func(abbr, full string) bool)
+	WordAbbreviations func(f func(abbr, full string) bool)
```
xiaq, line 21: Call this SmallWordAbbreviations; same for the edit API: `edit:small-word-abbr`.

For symmetry, there should also be `edit:word-abbr` and `edit:alnum-word-abbr`.

pkg/cli/codearea_test.go
```diff
@@ -265,33 +265,83 @@ var codeAreaHandleTests = []HandleTest{
...
 			},
 		}),
-		Events:       []term.Event{term.K('d'), term.K('n')},
-		WantNewState: CodeAreaState{Buffer: CodeBuffer{Content: "/dev/null", Dot: 9}},
+		Events:       []term.Event{term.K('d'), term.K('n'), term.K(' ')},
+		WantNewState: CodeAreaState{Buffer: CodeBuffer{Content: "/dev/null ", Dot: 10}},
 	},
 	{
-		Name: "abbreviation expansion preferring longest",
+		Name: "abbreviation expansion",
```
xiaq, line 276: Some early comments on the test cases (I haven't reviewed them in detail since they will change a bit after you act on my other comments):

* All test cases should have unique names describing the facet of the behavior they test against.

* Not sure why you removed the test against the behavior of preferring the longest candidate.

xiaq:
Re `edit:word-abbr` vs `edit:abbr-word`, I prefer the former; a cross-reference in the doc should be sufficient.

pkg/cli/codearea.go
krader1961: Due to restricting matches to word boundaries there can never be more than a single match.

pkg/cli/codearea.go
krader1961: Okay, done, but note doing it that way only eliminates one of the length checks and replaces it with the same length check that is implicit in the `HasSuffix()` call.

pkg/cli/codearea.go
krader1961: done

pkg/cli/codearea.go
krader1961: done

pkg/cli/codearea.go
krader1961: Okay, I can live with that. But note that the "expand on newline" behavior is how fish has behaved for many years and it doesn't appear to cause problems for users. I can't recall anyone opening an issue or raising this as a problem on the fish IM channels.

pkg/edit/insert_api.go
krader1961: I've restructured the code so that "regular" abbreviations have higher precedence and documented that fact.

pkg/cli/codearea.go
krader1961: That's done before calling this function. However, your comment caused me to realize there was a corner case this doesn't handle correctly. Consider the abbreviation `gc` => `git commit` and the user types `x`, ` `, [backspace], `g`, `c`, ` `. This implementation would incorrectly expand the abbreviation. I've fixed that by also taking into account  the char, if any, that precedes the current insert buffer.

pkg/cli/app_spec.go
krader1961: I'll include "small" in the respective tokens and adjust the documentation and implementation accordingly. In particular with respect to the idea of supporting arbitrary "small word" transitions per your other comment. It's not clear that supporting "alnum" abbreviations, as I've implemented in the PR up to this point, is justified given the more general "small word" abbreviation mechanism.

pkg/cli/codearea.go
krader1961: I like this idea. It supports my legacy "word" abbreviations imported from the `fish` shell and is more general.

krader1961:
@xiaq, PTAL. I believe this addresses all your prior feedback. But because that required significant revision to the `$edit:small-word-abbr` documentation I'll be surprised if you don't find any problems with that aspect of the P.R.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,71 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
...
+		*c = CodeBuffer{
+			Content: c.Content[:c.Dot-len(abbr)] + full + c.Content[c.Dot:],
+			Dot:     c.Dot - len(abbr) + len(full),
+		}
+		w.resetInserts()
+	}
+}
+
+// Try to expand a small word abbreviation.
+func (w *codeArea) expandSmallWordAbbr(triggerRune rune, triggerLen int) {
```
xiaq, line 223: There is no need to pass `triggerLen` as it can be entirely determined by `triggerRune`.

Furthermore, this function can take the categorizer as an argument to make it work for other types of word boundaries in future.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,71 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
...
+	// We only do small word expansions at the end of the line and only if
+	// there are characters in the insert buffer other than the trigger rune.
```
xiaq, line 225: These two conditions are different:

* Only expanding at the end of the buffer is a design choice and alters the behavior of the code.
* Only expanding when `len(w.inserts) > triggerLen` is purely an optimization

So they should be documented differently.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,71 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
...
+		if len(a) >= len(w.inserts) || len(a) < len(abbr) {
+			return // check the next abbreviation
+		}
+
+		// Verify the abbreviation is in the insert buffer.
+		if !strings.HasSuffix(inserts, a) {
+			return
+		}
+		// Verify the trigger rune creates a word boundary.
+		r2, _ := utf8.DecodeLastRuneInString(a)
```
xiaq, line 244: There is no `r`, so `r2` can be just called `r`.

pkg/cli/codearea.go
```diff
@@ -196,6 +200,71 @@ func (w *codeArea) handlePasteSetting(start bool) bool {
...
+	c := &w.State.Buffer
+	if len(w.inserts) <= triggerLen || c.Dot < len(c.Content) {
+		return
+	}
+
+	inserts := w.inserts[:len(w.inserts)-triggerLen] // ignore abbr trigger rune
+	var abbr, full string
+	// Find the longest matching abbreviation.
+	w.SmallWordAbbreviations(func(a, f string) {
+		if len(a) >= len(w.inserts) || len(a) < len(abbr) {
```
xiaq, line 235: The check `len(a) >= len(w.inserts)` is unnecessary here as its case will be caught by `strings.HasSuffix(inserts, a)` below.

pkg/cli/codearea.go
```diff
@@ -245,21 +316,25 @@ func (w *codeArea) handleKeyEvent(key ui.Key) bool {
...
+func CategorizeSmallWord(r rune) int {
+	switch {
+	case unicode.IsSpace(r):
+		return 0
+	case IsAlnum(r):
+		return 1
+	default:
+		return 2
+	}
+}
```
xiaq, line 340: This is fine, just noting that I'd like to move all the word logic from `pkg/edit` to `pkg/cli` in future.

pkg/cli/codearea_test.go
```diff
@@ -293,6 +293,67 @@ var codeAreaHandleTests = []HandleTest{
...
+	{
+		Name: "abbreviation expansion",
+		Given: NewCodeArea(CodeAreaSpec{
+			Abbreviations: func(f func(abbr, full string)) {
+				f("||", " | less")
+			},
+		}),
+		Events:       []term.Event{term.K('{'), term.K('e'), term.K('c'), term.K('h'), term.K('o'), term.K(' '), term.K('x'), term.K('}'), term.K('|'), term.K('|')},
+		WantNewState: CodeAreaState{Buffer: CodeBuffer{Content: "{echo x} | less", Dot: 15}},
+	},
```
xiaq, line 315: Test names should describe which aspects of the code are being tested.

pkg/cli/codearea.go
krader1961: Yes, this is arguably an instance of premature optimization but given the other, necessary, length check it seemed reasonable to do both at the same time to make it clear to the reader we can't possibly match the abbreviation if the insert buffer is too short and, potentially, shave a few nanoseconds off the run time for the common case.

pkg/cli/codearea.go
krader1961: Yes. This is called `r2` because an earlier revision did have an `r1` var.

pkg/cli/codearea.go
krader1961: I included the rune's encoded length because we had already converted the rune to a UTF-8 sequence and it seemed silly to do it a second time in this function. I didn't benchmark the cost so this could very well be an example of an unneeded optimization.

PR #983, opened 2020-04-24
Augment @cf expansion to work outside builtin: NS

krader1961:
The @cf documentation command currently only works correctly for the
builtin namespace. This augments the expansion of @cf arguments so
that an explicit namespace can be included. This makes it possible to
cross-reference between namespaces. For example, `@cf builtin:splits`
and `@cf str:split`.

### DIFF
--- a/cmd/website/macros/macros.go
+++ b/cmd/website/macros/macros.go
@@ -49,6 +49,26 @@ func expandTtyshot(line string) string {
 	return buf.String()
 }
 
+// Given a `@cf` reference convert it to a HTML link.
+func makeLink(cf string) string {
+	i := strings.IndexRune(cf, ':')
+	if i == -1 {
+		// This handles the @cf uses in the `builtin:` namespace to other
+		// sections in the same namespace.
+		return fmt.Sprintf("[`%s`](#%s)", cf, cf)
+	}
+
+	module := cf[:i]
+	symbol := cf[i+1:]
+	if module == "builtin" {
+		// The `builtin:` namespace is treated differently than other
+		// namespaces with regard to how hash tag references are named.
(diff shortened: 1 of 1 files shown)


cmd/website/macros/macros.go
```diff
@@ -66,7 +66,21 @@ func expandCf(line string) string {
...
 			sep = ", "
 		}
-		fmt.Fprintf(&buf, "%s[`%s`](#%s)", sep, target, target)
+		if i := strings.IndexRune(target, ':'); i == -1 {
+			// This handles the @cf uses in the `builtin:` namespace to other
+			// commands in the same namespace.
+			fmt.Fprintf(&buf, "%s[`%s`](#%s)", sep, target, target)
+		} else {
+			module := target[:i]
+			href := target[i+1:]
```
xiaq, line 75: This variable is not the href, but the unqualified symbol name; call it `symbol`.

cmd/website/macros/macros.go
```diff
@@ -66,7 +66,21 @@ func expandCf(line string) string {
 		} else {
 			sep = ", "
 		}
-		fmt.Fprintf(&buf, "%s[`%s`](#%s)", sep, target, target)
+		if i := strings.IndexRune(target, ':'); i == -1 {
```
xiaq, line 69: It will be much clearer to abstract these into a function:

```go
func makeLink(cf string) string
```

cmd/website/macros/macros.go
krader1961: Oops, yes, agreed.

cmd/website/macros/macros.go
krader1961: Agreed.

xiaq:
You can make `makeLink` only build the href, since the link text is the same. But I'll merge this and do that small change myself.

PR #984, opened 2020-04-24
Fix how [ctrl-v] works on BSD systems

### LINKED ISSUE #917: Need to type Ctrl-V twice (macOS)
On macOS, I need to type Ctrl-V (bound to `edit:insert-raw`) twice to get the desired effect. If I wish to insert a literal Ctrl-V, I need to type it _four_ times.

I only have to run `edit:insert-raw` once in the minibuffer, however.

krader1961:
On BSD systems iexten has to be explicitly disabled. Whereas on Linux
systems it is implicitly ignored if icanon is disabled.

Fixes #917

### DIFF
--- a/pkg/cli/term/setup_unix.go
+++ b/pkg/cli/term/setup_unix.go
@@ -26,6 +26,7 @@ func setup(in, out *os.File) (func() error, error) {
 	savedTermios := term.Copy()
 
 	term.SetICanon(false)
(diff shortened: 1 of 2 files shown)


hanche:
I can confirm that this fixes the problem for me on the mac. No surprise there, I suppose.

xiaq:
Thanks for looking into this 👍

PR #1018, opened 2020-05-11
Add a wildcard [type:xxx] modifier

### LINKED ISSUE #1015: Add more wildcard modifiers to filter on characteristics other than the name
Issue #298 was resolved by augmenting the wildcard modifiers as documented [here](https://elv.sh/ref/language.html#modifiers). I recently opened #1013 proposing a `path:` namespace that would provide commands such as `is-dir` and `is-file`. It would also be extremely useful if wildcards could be modified to match specific file types; e.g., `**[type:dir]` or `**[type:file]`. If #1013 is implemented the proposed modifiers aren't strictly speaking necessary since they can be written like this instead: `each [p]{ if (path:is-file $p) { put $p } } [**]`. Nonetheless, I think there is utility is being able to express a few of the more common cases directly in the wildcard. If for no other reason than it is far more efficient to do the filtering during wildcard expansion than after the fact.

Obviously we don't want this to become a mini-DSL that implements all the features of a tool like `find`. Timestamp comparison, for example, should probably not be implemented via wildcard modifiers. But the equivalent of `find -maxdepth n` would also be extremely useful and is otherwise expensive and hard to do after the wildcard has been expanded.

In addition to, or as an alternative to, the `[type:...]` modifier proposed above leverage the `os.FileInfo` proposal of #1013. Add a global modifier that takes a lambda that is called with a `os.FileInfo` like object and outputs `$true` or `$false` to determine if the path should be included or excluded. That would mostly resolve the performance implications of doing the filtering after the wildcard is expanded. While flexible and relatively efficient it's still very verbose -- especially for common cases such as enumerating just directories or files.

krader1961:
This allows filtering the wildcard expansion by path type. For example,
whether the path is a directory, regular file, or something else. So
instead of `find . -type f` you can now do `**[type:file]`.

Resolves #1015

### DIFF
--- a/pkg/eval/glob.go
+++ b/pkg/eval/glob.go
@@ -3,6 +3,7 @@ package eval
 import (
 	"errors"
 	"fmt"
+	"os"
 	"strings"
 	"unicode"
 
@@ -15,11 +16,17 @@ import (
 // wildcards.
 type GlobPattern struct {
 	glob.Pattern
-	Flags GlobFlag
-	Buts  []string
+	Flags  GlobFlag
+	Buts   []string
+	TypeCb func(os.FileMode) bool
 }
 
 type GlobFlag uint
+
+var typeCbMap = map[string]func(os.FileMode) bool{
+	"dir":     os.FileMode.IsDir,
+	"regular": os.FileMode.IsRegular,
+}
 
 const (
 	NoMatchOK GlobFlag = 1 << iota
@@ -35,9 +42,11 @@ var (
 )
 
 var (
-	ErrMustFollowWildcard   = errors.New("must follow wildcard")
-	ErrModifierMustBeString = errors.New("modifier must be string")
-	ErrWildcardNoMatch      = errors.New("wildcard has no match")
+	ErrMustFollowWildcard    = errors.New("must follow wildcard")
+	ErrModifierMustBeString  = errors.New("modifier must be string")
+	ErrWildcardNoMatch       = errors.New("wildcard has no match")
+	ErrMultipleTypeModifiers = errors.New("only one type modifier allowed")
+	ErrUnknownTypeModifier   = errors.New("unknown type modifier")
 )
 
 var runeMatchers = map[string]func(rune) bool{
@@ -74,6 +83,19 @@ func (gp GlobPattern) Index(k interface{}) (interface{}, error) {
 		}
 		gp.Segments[len(gp.Segments)-1] = glob.Wild{
 			lastSeg.Type, true, lastSeg.Matchers,
+		}
+	case strings.HasPrefix(modifier, "type:"):
+		if gp.TypeCb != nil {
+			return nil, ErrMultipleTypeModifiers
+		}
+		switch {
+		default:
+			typeName := modifier[len("type:"):]
+			cb, ok := typeCbMap[typeName]
+			if !ok {
+				return nil, ErrUnknownTypeModifier
+			}
+			gp.TypeCb = cb
 		}
 	default:
 		var matcher func(rune) bool
@@ -123,6 +145,13 @@ func (gp GlobPattern) Concat(v interface{}) (interface{}, error) {
 		gp.append(rhs.Segments[0])
 		gp.Flags |= rhs.Flags
 		gp.Buts = append(gp.Buts, rhs.Buts...)
+		// This handles illegal cases such as `**[type:regular]x*[type:directory]`.
+		if gp.TypeCb != nil && rhs.TypeCb != nil {
+			return nil, ErrMultipleTypeModifiers
+		}
+		if rhs.TypeCb != nil {
+			gp.TypeCb = rhs.TypeCb
+		}
 		return gp, nil
 	}
 
@@ -135,7 +164,8 @@ func (gp GlobPattern) RConcat(v interface{}) (interface{}, error) {
 		segs := stringToSegments(lhs)
 		// We know gp contains exactly one segment.
 		segs = append(segs, gp.Segments[0])
-		return GlobPattern{glob.Pattern{Segments: segs}, gp.Flags, gp.Buts}, nil
+		return GlobPattern{Pattern: glob.Pattern{Segments: segs}, Flags: gp.Flags,
+			Buts: gp.Buts, TypeCb: gp.TypeCb}, nil
 	}
 
 	return nil, vals.ErrConcatNotImplemented
@@ -208,15 +238,20 @@ func doGlob(gp GlobPattern, abort <-chan struct{}) ([]interface{}, error) {
 	}
 
 	vs := make([]interface{}, 0)
-	if !gp.Glob(func(name string) bool {
+	if !gp.Glob(func(pathInfo glob.PathInfo) bool {
 		select {
 		case <-abort:
 			logger.Println("glob aborted")
 			return false
 		default:
 		}
-		if _, b := but[name]; !b {
-			vs = append(vs, name)
+
+		if _, ignore := but[pathInfo.Path]; ignore {
+			return true
+		}
+
+		if gp.TypeCb == nil || (gp.TypeCb)(pathInfo.Info.Mode()) {
+			vs = append(vs, pathInfo.Path)
 		}
 		return true
 	}) {
--- a/pkg/eval/glob_test.go
+++ b/pkg/eval/glob_test.go
@@ -88,3 +88,19 @@ func TestGlob_But(t *testing.T) {
 		That("put *[but:foobar][but:ipsum]").Puts("bar", "foo", "lorem"),
 	)
 }
+
+func TestGlob_Type(t *testing.T) {
+	_, cleanup := util.InTestDir()
+	defer cleanup()
+
+	mustMkdirAll("d1", "d2", ".d", "b/c")
+	mustCreateEmpty("bar", "foo", "ipsum", "lorem", "d1/f1", "d2/fm")
+
+	Test(t,
+		That("put **[type:dir]").Puts("b/c", "b", "d1", "d2"),
+		That("put **[type:regular]").Puts("d1/f1", "d2/fm", "bar", "foo", "ipsum", "lorem"),
+		That("put **[type:regular]m").Puts("d2/fm", "ipsum", "lorem"),
+		That("put **[type:dir]f*[type:regular]").ThrowsCause(ErrMultipleTypeModifiers),
+		That("put **[type:unknown]").ThrowsCause(ErrUnknownTypeModifier),
+	)
+}
--- a/pkg/glob/glob.go
+++ b/pkg/glob/glob.go
@@ -10,13 +10,24 @@ import (
 
 // TODO: Use native path separators instead of always using /.
 
+// PathInfo is used in glob expansion callbacks to efficiently determine if a
+// given pathname satisfies a particular constraint.
+type PathInfo struct {
+	// `Path` is needed to generate paths consistent with the original glob
+	// pattern. That is, the `Info.Name()` value might be just the final
+	// component of `Path`. Glob modifiers need the complete path in addition
+	// to information about the target of the path.
+	Path string
+	Info os.FileInfo
+}
+
 // Glob returns a list of file names satisfying the given pattern.
-func Glob(p string, cb func(string) bool) bool {
+func Glob(p string, cb func(PathInfo) bool) bool {
 	return Parse(p).Glob(cb)
 }
 
 // Glob returns a list of file names satisfying the Pattern.
-func (p Pattern) Glob(cb func(string) bool) bool {
+func (p Pattern) Glob(cb func(PathInfo) bool) bool {
 	segs := p.Segments
 	dir := ""
 
@@ -49,7 +60,7 @@ func isDrive(s string) bool {
 // glob finds all filenames matching the given Segments in the given dir, and
 // calls the callback on all of them. If the callback returns false, globbing is
 // interrupted, and glob returns false. Otherwise it returns true.
-func glob(segs []Segment, dir string, cb func(string) bool) bool {
+func glob(segs []Segment, dir string, cb func(PathInfo) bool) bool {
 	// Consume non-wildcard path elements simply by following the path. This may
 	// seem like an optimization, but is actually required for "." and ".." to
 	// be used as path elements, as they do not appear in the result of ReadDir.
@@ -63,11 +74,14 @@ func glob(segs []Segment, dir string, cb func(string) bool) bool {
 	}
 
 	if len(segs) == 0 {
-		return cb(dir)
+		if info, err := os.Stat(dir); err == nil {
+			return cb(PathInfo{dir, info})
+		}
+		return true
 	} else if len(segs) == 1 && IsLiteral(segs[0]) {
 		path := dir + segs[0].(Literal).Data
-		if _, err := os.Stat(path); err == nil {
-			return cb(path)
+		if info, err := os.Stat(path); err == nil {
+			return cb(PathInfo{path, info})
 		}
 		return true
 	}
@@ -142,7 +156,12 @@ func glob(segs []Segment, dir string, cb func(string) bool) bool {
 	for _, info := range infos {
 		name := info.Name()
 		if matchElement(segs, name) {
-			if !cb(dir + name) {
+			dirname := dir + name
+			info, err := os.Stat(dirname)
+			if err != nil {
+				return true
+			}
+			if !cb(PathInfo{dirname, info}) {
 				return false
 			}
 		}
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -907,8 +907,10 @@ The following behaviors are default, although they can be altered by modifiers:
 
 Wildcards can be **modified** using the same syntax as indexing. For instance,
 in `*[match-hidden]` the `*` wildcard is modified with the `match-hidden`
-modifier. Multiple matchers can be chained like `*[set:abc][range:0-9]`. There
-are two kinds of modifiers:
+modifier. Multiple matchers can be chained like `*[set:abc][range:0-9]`. In
+which case they are OR'ed together.
+
+There are two kinds of modifiers:
 
 **Global modifiers** apply to the whole pattern and can be placed after any
 wildcard:
@@ -920,11 +922,17 @@ wildcard:
 -   `but:xxx` (where `xxx` is any filename) excludes the filename from the final
     result.
 
-Although global modifiers affect the entire wildcard pattern, you can add it
-after any wildcard, and the effect is the same. For example,
-`put */*[nomatch-ok].cpp` and `put *[nomatch-ok]/*.cpp` do the same thing.
+-   `type:xxx` (where `xxx` is a recognized file type from the list below). Only
+    one type modifier is allowed. For example, to find the directories at any
+    level below the current working directory: `**[type:dir]`.
 
-On the other hand, you must add it after a wildcard, instead of after the entire
+    -   `dir` will match if the path is a directory.
+
+    -   `regular` will match if the path is a regular file.
+
+after any wildcard, and the effect is the same. For example,
+`put */*[nomatch-ok].cpp` and `put *[nomatch-ok]/*.cpp` do the same thing. On
+the other hand, you must add it after a wildcard, instead of after the entire
 pattern: `put */*.cpp[nomatch-ok]` unfortunately does not do the correct thing.
 (This will probably be fixed.)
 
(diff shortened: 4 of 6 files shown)


xiaq:
Thanks for tackling this. There are some open design questions that I don't have good answers for, and I suggest we start with a modest feature set and iterate on this.

Ultimately, I'd like Elvish to have a small, general-purpose query syntax with support for NOT, AND and OR relations, and the glob modifiers syntax should be just an example of it. The same syntax can be used, for example, in the navigation mode to filter files based on their properties (which should use exactly the same syntax as glob modifiers), and in command history listing to filter out history items (when they have metadata such as working directory, duration, etc.).

I haven't reviewed the code in detail, so some design comments:

* Juxtaposition of filters strongly suggests AND: this is how most search engine's query languages work. I'd prefer to start simply forbidding multiple `type:` in this PR before we agree on the design of how to do AND and OR.

* The abbreviated form for `type:dir`, `type:d`, isn't actually very short. I'd just use `/`; which is easy to remember and consistent with `ls -F`. There is no suffix for regular files but I'd copy `.` from zsh's list of single-letter glob modifiers (see section 14.8.7 [here](http://zsh.sourceforge.net/Doc/Release/Expansion.html)). In fact, Elvish's glob modifier feature is inspired by Zsh's, except that it is done right ;-)

* There is no well-defined "other" category; "not a regular file and not a directory" is an arbitrary definition. I suggest dropping it from the PR, and maybe add all the special device classes etc later. Also, what is currently "other" can be expressed in terms of the NOT syntax, when we have that designed.

xiaq:
To see why the AND semantics for juxtaposition is useful and intuitive, consider that we may want to add filters for owners and permissions. It's intuitive to write `**[type:dir owner:xiaq]`, and the underlying relationship is AND.

xiaq:
Some more comments:

* I prefer `type:regular` to `type:file`. Nominally directories are also files, so `type:file` is not clear. "Regular file" is not 100% clear either but definitely an improvement.

* For the short form: not all filters need to have a short form, only when they are used a lot and the short form is very intuitive. So I'd suggest starting with only `/` for `type:dir`, and live without a short form for `type:regular`.

xiaq:
For the general purpose query syntax, the [search query language](https://developers.google.com/issue-tracker/concepts/search-query-language) of Google's issue tracker is a good model, but unfortunately much of it won't work directly with Elvish due to syntax clashes. Elvish's query syntax should play well with the language syntax.

xiaq:
Oh, and perhaps more notably, there is also the [syntax](https://developers.google.com/code-search/reference) of Google's codesearch, which is very similar.

krader1961:
Gah! Please don't implement a query DSL just for use in wildcard expansion. Better to allow a lambda as a callback that takes two args: the path and the `os.FileInfo` object. It can then use Elvish, along with the commands from the proposed `path:` module to do arbitrary filtering.

> I prefer type:regular to type:file

I'm going to disagree. Not only is "regular" longer it is less clear. Most UNIX users would be surprised if you told them directories are just special files. But everyone, including geeks like us, understand

> I'd prefer to start with simply forbidding multiple type:

Works for me. In practice I would not expect anyone to ever use more than one `type:` modifier. And as you note the implicit "or" is inconsistent with how the modifiers otherwise work.

> The abbreviated form for type:dir, type:d, isn't actually very short. I'd just use /; which is easy to remember and consistent with ls -F. There is no suffix for regular files but I'd copy . from zsh's list of single-letter glob modifiers.

I am **not** a fan of zsh single character glob qualifiers. And I say that despite having been a fan of Perl, a language some have described as serial line noise, for many years. Way too magical for my taste. Worse, they require memorizing an otherwise unimportant details (the list of magic single char qualifiers) or having to perform a difficult search to find out what `[/]` means. Clarity is more important in this case than avoiding typing seven characters.

krader1961:
I removed `[type:other]` and support for more than one type modifier. I'd like to leave introduction of single-char modifiers such as `[/]` until there has been more discussion about the merits of going down that path. Please take another look.

xiaq:
A generic functional API and DSL are not conflicting goals. Introducing `/` as a shorthand for `type:dir` does not mean copying Zsh's qualifiers wholesale.

krader1961:
> Introducing / as a shorthand for type:dir does not mean copying Zsh's qualifiers wholesale.

Yes, but my intuition, and decades of experience, is that restricting the expansion to directories is much less common than restricting it to regular files. Consider how often you use a wildcard to generate a list of files to be passed to commands like `grep`. Versus how often you use a wildcard to generate a list of directories. The reason I have a `ff` function is that 99% of the time I want to enumerate all the regular files at or below the CWD.

If either of those two cases deserve a single-char shorthand it is the regular file case. So I'm willing to introduce `[/]` as an alias for `[type:dir]` only if we can agree on a single char alias for `[type:file]`. AFAICT, bash is particularly brain-dead in this regard as it doesn't support restricting matches to regular files and and has a rather obtuse mechanism for matching directories:

```
              *      Matches any string, including the null  string.   When  the  globstar
                     shell  option  is enabled, and * is used in a pathname expansion con-
                     text, two adjacent *s used as a single pattern will match  all  files
                     and zero or more directories and subdirectories.  If followed by a /,
                     two adjacent *s will match only directories and subdirectories.
```

Is there some reason we can't just support `[d]` and `[f]`? AFAICT those do not conflict with any existing wildcard modifiers an if we're going to support single-char modifiers I don't see any reason to prefer `[/]` over `[d]`.

xiaq:
* My experience and intuition are different from you, but I am convinced that restricting wildcards to regular files is also a frequent use case.

* I don't feel particularly strong about `/` as a shorthand for `type:dir`, so I'm fine with leaving it out. It could be useful for other purposes, and having to spell out `type:dir` is probably for the best.

* I remain unconvinced about using "file" to mean *regular* file. I know it's a strong tradition used by both `test(1)` and `find(1)`, but it's worth making a break here.

  It's not so problematic when it comes to the file/directory distinction, but it *is* problematic when it comes to things such as device files. The name "file" makes it really unclear whether it includes device files or not. On the other hand, "regular file" is well-defined in the [C API](https://www.man7.org/linux/man-pages/man2/stat.2.html), in the [Go API](https://godoc.org/os#FileMode.IsRegular), and in other language bindings such as [Python](https://docs.python.org/3/library/stat.html#stat.S_ISREG), where it does not include device files.

* I acknowledge `type:d` and `type:f` as as a nod `find(1)`'s `-type d` and `-type f`, but remain unconvinced about the need for them.

I suggest starting with only `type:dir` and `type:regular` and nothing more for the sake of this PR.

krader1961:
PTAL. I changed "type:file" to "type:regular" and also introduced `[d]` and `[r]` as shorthand for the long forms.

pkg/eval/glob.go
```diff
@@ -15,12 +16,23 @@ import (
...
+	Buts   []string
+	TypeCb func(os.FileMode) bool
 }
 
 type GlobFlag uint
 
+const (
+	typeDir     = "dir"
+	typeRegular = "regular"
+)
```
xiaq, line 29: Remove

pkg/eval/glob.go
```diff
@@ -75,6 +89,24 @@ func (gp GlobPattern) Index(k interface{}) (interface{}, error) {
 		gp.Segments[len(gp.Segments)-1] = glob.Wild{
 			lastSeg.Type, true, lastSeg.Matchers,
 		}
+	case modifier == "d" || modifier == "r" || strings.HasPrefix(modifier, "type:"):
```
xiaq, line 92: `type:dir` and `type:regular` and nothing more; remove "d" and "r".

pkg/eval/glob.go
```diff
@@ -75,6 +89,24 @@ func (gp GlobPattern) Index(k interface{}) (interface{}, error) {
...
+		// This handles cases such as `**x*[r][d]`.
```
xiaq, line 93: Change this comment to use the longhand form, or just remove it as it's quite self-evident what this branch does.

pkg/eval/glob.go
```diff
@@ -75,6 +89,24 @@ func (gp GlobPattern) Index(k interface{}) (interface{}, error) {
...
+		if gp.TypeCb != nil {
+			return nil, ErrMultipleTypeModifiers
+		}
+		switch {
+		case modifier == "d":
+			gp.TypeCb = typeMap[typeDir]
+		case modifier == "r":
+			gp.TypeCb = typeMap[typeRegular]
```
xiaq, line 101: Remove

pkg/eval/glob.go
```diff
@@ -75,6 +89,24 @@ func (gp GlobPattern) Index(k interface{}) (interface{}, error) {
...
+		default:
+			ptype := modifier[len("type:"):]
```
xiaq, line 103: I know you can't use `type` since it's a keyword, but not clear what `ptype` means; `typeName` is better. Or just don't use a variable here, it is only used in the next statement.

pkg/eval/glob.go
```diff
@@ -15,12 +16,23 @@ import (
...
 }
 
 type GlobFlag uint
 
+const (
+	typeDir     = "dir"
+	typeRegular = "regular"
+)
+
+var typeMap = map[string]func(os.FileMode) bool{
```
xiaq, line 31: `typeCbMap` would be a bit clearer.

pkg/eval/glob.go
```diff
@@ -123,6 +155,11 @@ func (gp GlobPattern) Concat(v interface{}) (interface{}, error) {
 		gp.append(rhs.Segments[0])
 		gp.Flags |= rhs.Flags
 		gp.Buts = append(gp.Buts, rhs.Buts...)
+		// This handles cases such as `**[r]x*[d]`.
```
xiaq, line 158: Change this comment to use the longhand form.

pkg/eval/glob.go
```diff
@@ -123,6 +155,11 @@ func (gp GlobPattern) Concat(v interface{}) (interface{}, error) {
...
+		if gp.TypeCb != nil && rhs.TypeCb != nil {
+			return nil, ErrMultipleTypeModifiers
+		}
+		gp.TypeCb = rhs.TypeCb
```
xiaq, line 162: This should only be done if `rhs.TypeCb != nil`.

pkg/eval/glob.go
```diff
@@ -208,15 +246,24 @@ func doGlob(gp GlobPattern, abort <-chan struct{}) ([]interface{}, error) {
...
+		if _, ignore := but[pathInfo.Path]; ignore {
+			return true
+		}
+
+		if gp.TypeCb == nil { // no file type restrictions, common case
+			vs = append(vs, pathInfo.Path)
+		} else {
+			if (gp.TypeCb)(pathInfo.Info.Mode()) {
+				vs = append(vs, pathInfo.Path)
+			}
```
xiaq, line 266: This can be written a bit more succinctly as:

```go
if gp.TypeCb == nil || gp.TypeCb(pathInfo.Info.Mode()) {
    vs = append(vs, pathInfo.Path)
}
```

pkg/eval/glob_test.go
```diff
@@ -88,3 +88,23 @@ func TestGlob_But(t *testing.T) {
...
+	_, cleanup := util.InTestDir()
+	defer cleanup()
+
+	mustMkdirAll("d1", "d2", ".d", "b/c")
+	mustCreateEmpty("bar", "foo", "ipsum", "lorem", "d1/f1", "d2/fm")
+
+	Test(t,
+		That("put **[d]").Puts("b/c", "b", "d1", "d2"),
+		That("put **[type:dir]").Puts("b/c", "b", "d1", "d2"),
+		That("put **[r]f*").Puts("d1/f1", "d2/fm", "foo"),
```
xiaq, line 102: This test doesn't really test the regular modifier since only regular files have `f` in their names, meaning that it fails to capture the bug I mentioned earlier in the review (where the rhs's nil TypeCb overrides the non-nil TypeCb).

pkg/glob/glob.go
```diff
@@ -10,13 +10,18 @@ import (
 
 // TODO: Use native path separators instead of always using /.
 
+type PathInfo struct {
```
xiaq, line 13: Exported symbols must have godoc comments.

pkg/glob/glob.go
```diff
@@ -10,13 +10,18 @@ import (
...
+	Path string
```
xiaq, line 14: A reader might question why we can't just use `Info.Name()` here, so document the reason - `Path` is needed to generate paths consistent with the original glob pattern.

pkg/glob/glob.go
```diff
@@ -63,11 +68,14 @@ func glob(segs []Segment, dir string, cb func(string) bool) bool {
 	}
 
 	if len(segs) == 0 {
-		return cb(dir)
+		if info, err := os.Stat(dir); err == nil {
```
xiaq, line 71: You can save the `FileInfo` the code on line 65 obtained earlier and avoid another `Stat`.

pkg/glob/glob.go
```diff
@@ -63,11 +68,14 @@ func glob(segs []Segment, dir string, cb func(string) bool) bool {
...
+			return cb(PathInfo{dir, info})
+		}
+		return true
 	} else if len(segs) == 1 && IsLiteral(segs[0]) {
 		path := dir + segs[0].(Literal).Data
-		if _, err := os.Stat(path); err == nil {
-			return cb(path)
+		if info, err := os.Stat(path); err == nil {
```
xiaq, line 77: This should use `Lstat` to avoid dropping broken symlinks.

This will also drop files that you can't stat. It is a sound strategy to handle the error, since that indicates that the filesystem is changing under our feet and there isn't much we can do.

However, it's worth calling that out as an conscious choice in a comment, since this may raise a red flag to a reader as not handling the error properly.

pkg/glob/glob.go
```diff
@@ -142,7 +150,12 @@ func glob(segs []Segment, dir string, cb func(string) bool) bool {
 	for _, info := range infos {
 		name := info.Name()
 		if matchElement(segs, name) {
-			if !cb(dir + name) {
+			name = dir + name
+			info, err := os.Stat(name)
+			if err != nil {
+				return true
+			}
```
xiaq, line 157: You already have a `FileInfo`; this is redundant.

pkg/glob/glob.go
```diff
@@ -142,7 +150,12 @@ func glob(segs []Segment, dir string, cb func(string) bool) bool {
 	for _, info := range infos {
 		name := info.Name()
 		if matchElement(segs, name) {
-			if !cb(dir + name) {
+			name = dir + name
```
xiaq, line 153: Do not modify `name` in place; it gets confusing.

website/ref/language.md
```diff
@@ -949,19 +962,24 @@ pattern: `put */*.cpp[nomatch-ok]` unfortunately does not do the correct thing.
...
+Note the following caveats:
+
+-   Local matchers chained together in separate modifiers are OR'ed. For
+    instance, `?[set:aeoiu][digit]` matches all files with the chars `aeoiu` or
+    containing a digit.
 
-    -   Multiple matchers, they are OR'ed. For instance, ?[set:aeoiu][digit]
-        matches `aeoiu` and digits.
+-   Local matchers combined in the same modifier, such as `?[set:aeoiu digit]`,
+    are OR'ed but in a manner something like a SQL outer join. This results in
```
xiaq, line 972: This behavior is more easily and unambiguously explained by saying that `?[set:aeoiu digit]` expands to `?[set:aeiou] ?[digit]`.

website/ref/language.md
```diff
@@ -920,11 +922,22 @@ wildcard:
...
+    level below the current working directory: `**[type:dir]`.
+
+    -   `dir` will match if the path is a directory.
+
+    -   `regular` will match if the path is a regular file.
 
-On the other hand, you must add it after a wildcard, instead of after the entire
+-   `d` is shorthand for `type:dir`.
+
+-   `r` is shorthand for `type:regular`.
```
xiaq, line 935: Remove.

xiaq:
Hi! Sorry for nagging, but do you intend to act on the feedback soon?

If you don't have time to work on this now, I can merge it and do the changes myself.

pkg/glob/glob.go
krader1961: FYI, The existing code violates this best practice. A lot. I installed `golint` a few weeks ago. Simply doing `golint pkg/eval` identifies 24 such instances :-)

pkg/eval/glob.go
krader1961: Renamed. Note: I prefer a temp var in this situation because it's much clearer to a human and shouldn't be any slower.

pkg/eval/glob.go
krader1961: Okay. Done. That makes this more painful to use interactively but presumably you would prefer not to institute these shorthands until more experience using this feature is gained and there has been more discussion about whether shorthands like `/` for `type:dir` is preferable to `d` or some other shorthand.

pkg/glob/glob.go
krader1961: Agreed. Changed to `dirname`.

website/ref/language.md
krader1961: Your example does not match the behavior I'm seeing. Either standalone or bracketed by `*` wildcards. For example:

```
 macpro  ~/tmp  > echo ?[set:aeoiu digit]
1 a 1 a
 macpro  ~/tmp  > echo ?[set:aeoiu] ?[digit]
a 1
 macpro  ~/tmp  > echo *?[set:aeoiu digit]*
1 :3 a a1 b2 d:4 1 :3 a a1 b2 d:4
 macpro  ~/tmp  > echo *?[set:aeoiu]?[digit]*
a1
```

Frankly, I think combining local modifiers in this fashion should not be allowed. At least not until it has a behavior that can be explained to an average shell user. So I'm simply going to modify the comment to say "do not use".

pkg/glob/glob.go
krader1961: I have no idea what you're telling me. There is no usable `FileInfo` object at this point in the code.

pkg/glob/glob.go
krader1961: I have no idea what you're suggesting. If `len(segs) == 0` was true when this function was entered then the preceding `os.Stat()` would not have been run.

pkg/glob/glob.go
krader1961: Since the existing, and surrounding, code uses `Stat` I'd prefer to leave that unchanged with respect to implementing this feature. If this should be changed to `Lstat` there probably needs to be a careful review of the other paths that call `Stat` and appropriate comments added as to why one or the other is used.

Similarly, when `Readdir` fails you have the comment "// XXX Silently drop the error". But that same comment isn't present in the `Stat` error cases. Since adding such comments are orthogonal to this change I'd prefer they be done in a separate change.

krader1961:
PTAL. This addresses all comments other than a) the suggestion to replace an existing `os.Stat` with `os.Lstat`, and b) using the return value of an earlier `os.Stat` since, AFAICT, we can't count on the previous `os.Stat` to have been executed when its output is needed later in the code.

krader1961:
Regarding the documentation of multiple matchers in the same modifier... the current P.R. says to avoid using that form because the behavior is not easy to describe, let alone intuitive. See https://github.com/elves/elvish/issues/1019#issuecomment-635095672 for an example why.

pkg/eval/glob.go
```diff
@@ -208,15 +238,20 @@ func doGlob(gp GlobPattern, abort <-chan struct{}) ([]interface{}, error) {
...
 		default:
 		}
-		if _, b := but[name]; !b {
-			vs = append(vs, name)
+
+		if _, ignore := but[pathInfo.Path]; ignore {
+			return true
+		}
+
+		if gp.TypeCb == nil || (gp.TypeCb)(pathInfo.Info.Mode()) {
```
xiaq, line 253: Unnecessary parentheses around `gp.TypeCb`.

pkg/glob/glob.go
xiaq: Yes, but I unapologetically hold new code to a higher standard :)

pkg/glob/glob.go
xiaq: `len(segs) == 0` will only become true if the loop has been run, which shortens `segs` (`segs = segs[2:]`), and this function is never called with an empty `segs`.

OTOH, it's not easy to reason all that. The code as it is now is easier to reason about.

website/ref/language.md
```diff
@@ -920,11 +922,17 @@ wildcard:
...
-`put */*[nomatch-ok].cpp` and `put *[nomatch-ok]/*.cpp` do the same thing.
+-   `type:xxx` (where `xxx` is a recognized file type from the list below). Only
+    one type modifier is allowed. For example, to find the directories at any
+    level below the current working directory: `**[type:dir]`.
+
+    -   `dir` will match if the path is a directory.
 
-On the other hand, you must add it after a wildcard, instead of after the entire
+    -   `regular` will match if the path is a regular file.
+
```
xiaq, line 932: Removed the first line of this paragraph by mistake?

PR #1022, opened 2020-05-13
Update the docs regarding builtin modules

krader1961:
The existing documentation is somewhat inconsistent with respect to
builtin modules and what constitutes "reference" documentation. This
change attempts to bring a bit more consistency to the documentation and
fill in some gaps.


website/ref/language.md
```diff
@@ -1664,21 +1664,36 @@ The following namespaces have special meanings to the language:
 
 ## Pre-Defined Modules
 
-Namespaces that are not special (i,e. one of the above) are also called
+Namespaces that are not special (i.e. one of the above) are also called
 **modules**. Aside from these special namespaces, Elvish also comes with the
-following modules:
+following modules that are always available, but might need to be imported by a
+`use` command:
```
xiaq, line 1670: It is more productive to just list the names of all the pre-defined modules, as links to their respective reference pages, than repeating the index page of the reference section. It's also worthwhile to split out `edit` as a special case that might change in future.

Something like:

> * The following modules are always available: [exc:](exc.html), ...
> * The [unix](unix.html) module is available on UNIX-like platforms.
> * The [edit](edit.html) module is available in interactive module. As a special case, it does not need importing, but this may change in future.

xiaq:
Oh, the typo correction is in its own commit. Then that's fine.

website/ref/readline-binding.md
```diff
@@ -0,0 +1,19 @@
...
+```elvish
+use readline-binding
+```
+
+Note that this will override some of the standard bindings. For example, Ctrl-L
+will be bound to a function that clears the terminal screen rather than start
+[location mode](../learn/cookbook.html).
+
+See the
+<a href="https://github.com/elves/elvish/blob/master/pkg/eval/bundled/readline-binding.elv.go">source
```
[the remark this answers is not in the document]
krader1961: Simply a brain fart. Fixed.

website/ref/index.toml
```diff
@@ -1,55 +1,47 @@
...
 [[articles]]
-name = "epm"
-title = "The Elvish Package Manager (epm:)"
+name = "readline-binding"
+title = "The Readline-like Key Bindings Module (readline-binding)"
 
 [[articles]]
 name = "store"
-title = "The Elvish Store (store:)"
+title = "The Interactive Command History Module (store)"
```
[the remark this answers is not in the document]
krader1961: Okay, although I'm still a bit perplexed what the `store` module is intended to do beyond provide an API for the interactive command history.

website/ref/index.toml
```diff
@@ -1,55 +1,47 @@
...
 
 [[articles]]
 name = "builtin"
-title = "The Builtin Module"
+title = "The Core Commands Module (builtin)"
 
 [[articles]]
 name = "edit"
-title = "The Editor Module (edit:)"
+title = "The Interactive Editor Module (edit)"
```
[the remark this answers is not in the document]
krader1961: Haha! I agree. I actually spent nearly an hour trying to make that work. By which I mean produce readable output. The lack of vertical alignment of the module title is extremely annoying. That is, it looks like

```
edit: API...
epm: Elvish...
ex: Exception...
```

Yes, I deliberately misspelled the last one. Where what you want is

```
edit: API...
epm:  Elvish...
ex:   Exception...
```

I vote that we leave improving this aspect of the module list to a separate change.

website/ref/index.toml
```diff
@@ -1,55 +1,47 @@
 [[articles]]
 name = "language"
-title = "The Language"
+title = "The Language Specification"
 
 [[articles]]
 name = "builtin"
-title = "The Builtin Module"
+title = "The Core Commands Module (builtin)"
```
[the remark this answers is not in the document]
krader1961: Done

website/ref/language.md
krader1961: Fine by me but note the existing style of this list was created by you. :smile: Changed.

website/ref/index.toml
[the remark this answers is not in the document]
krader1961: I decided to learn a tiny bit of CSS and think I have a solution for keeping this information readable.

krader1961:
Please take another look.

krader1961:
FYI, You can see how the reformatted reference index looks here: https://www.skepticism.us/elvish/ref/

website/ref/index.toml
```diff
@@ -1,55 +1,47 @@
 [[articles]]
 name = "language"
-title = "The Language"
+title = "The Language Specification"
 
 [[articles]]
 name = "builtin"
-title = "The Builtin Module"
+title = "<span class=module>builtin:</span> Builtin Functions and Variables"
```
xiaq, line 7: Please refrain from doing that and just use plain text title.

website/ref/language.md
```diff
@@ -1675,21 +1675,21 @@ The following namespaces have special meanings to the language:
...
```
xiaq, line 1681: "always available" is duplicate with the first bullet and contradicts the second bullet. Better to say:

> Elvish also comes with the following modules that can be imported with `use`:

And make the `edit` bullet not a bullet, just a separate paragraph.

website/ref/readline-binding.md
```diff
@@ -0,0 +1,19 @@
+<!-- toc -->
+
+# Introduction
+
+The `readline-binding` module provides readline-like key bindings. Such as
```
xiaq, line 5: Use a comma instead of period before "such as".

website/ref/readline-binding.md
```diff
@@ -0,0 +1,19 @@
...
+binding Ctrl-A to move the cursor to the start of the line. To use, put the
+following in `~/.elvish/rc.elv`:
+
+```elvish
+use readline-binding
+```
+
+Note that this will override some of the standard bindings. For example, Ctrl-L
```
xiaq, line 13: Wrap keys inside `<span class="key">`.

website/style.css
```diff
@@ -154,7 +154,13 @@ a {
     content: "\00a0";
 }
 
-code, pre {
+.module {
```
xiaq, line 157: No.

krader1961: Okay, but the lack of such a mechanism means that placing the module name at the front of the title severely reduces the readability of the generated index. I agree that using this as in this example is suboptimal:

```
title = "<span class=module>builtin:</span> Builtin Functions and Variables"
```

Ideally there would be an alternative to the `title` attribute that allows specifying a module name as a prefix to the title text and that is rendered in a distinct, fixed with, manner. If I were using HTML or Markdown directly to write the index document I would use a table.

website/ref/index.toml
krader1961: Okay, but that means putting the module name at the front of the title results in a very hard to read index. So I'll revert to putting the module name at the end of the title enclosed in parentheses.

website/ref/language.md
krader1961: Done.

website/ref/readline-binding.md
krader1961: Done.

website/ref/readline-binding.md
krader1961: Done.

krader1961:
PTAL

website/ref/index.toml
```diff
@@ -1,55 +1,47 @@
 [[articles]]
 name = "language"
-title = "The Language"
+title = "The Language Specification"
 
 [[articles]]
 name = "builtin"
-title = "The Builtin Module"
+title = "Builtin Functions and Variables (builtin:)"
```
xiaq, line 7: `builtin:` is not the canonical way to use the builtin module, only for introspection. So I'd rather this be left out of the title. A paragraph inside `builtin.html` can explain the fact that `$builtin:` can be used for introspection.

website/style.css
xiaq, line 157: Revert

website/ref/index.toml
krader1961: Done

website/style.css
krader1961: Done

krader1961:
PTAL. I also clarified the first paragraph of the `builtin` introduction regarding when the namespace prefix needs to be used.

website/ref/builtin.md
```diff
@@ -4,8 +4,21 @@
 
 The builtin module contains facilities that are potentially useful to all users.
 It occupies the `builtin:` namespace. You rarely have to explicitly specify the
-namespace though, since it is one of the namespaces consulted when resolving
-unqualified names.
+namespace since it is one of the namespaces consulted when resolving unqualified
+names. The exception to this rule is if you define a function with the same name
+as a builtin. In which case to use the builtin rather than the function you need
```
xiaq, line 9: I'd always write "which" clauses as part of the same sentence but I'll fix that myself.

website/ref/index.toml
```diff
@@ -1,55 +1,47 @@
...
 
 [[articles]]
 name = "builtin"
-title = "The Builtin Module"
+title = "Builtin Functions and Variables"
 
 [[articles]]
 name = "edit"
-title = "The Editor Module (edit:)"
+title = "API for the Interactive Editor (edit:)"
```
xiaq, line 11: I see you reverted the change of putting the module names at the front; I prefer keeping that change, but I can do that myself too.

krader1961:
> I see you reverted the change of putting the module names at the front; I prefer keeping that change, but I can do that myself too.

That was deliberate. My first version put the module names at the front of the sentence but that made it harder to read. That's why I experimented with a custom CSS class to force the module names to all have the same width.

PR #1027, opened 2020-05-15
Clarify the rules for using styled #RGB values

krader1961:
Document that an unqouted `#` character has special meaning to the Elvish
parser in the context of passing `#RGB` values to the `styled` command.

### DIFF
--- a/pkg/eval/builtin_fn_styled.go
+++ b/pkg/eval/builtin_fn_styled.go
@@ -68,9 +68,17 @@ var errStyledSegmentArgType = errors.New("argument to styled-segment must be a s
 //     -   Any color from the xterm 256-color palette, as `colorX` (such as
 //    `color12`).
 //
-//     -   A 24-bit RGB color, as `#RRGGBB`, such as `#778899`.
+//     -   A 24-bit RGB color written as `#RRGGBB` such as `'#778899'`.
+//
+//		   **NOTE:** You need to quote such values since an unquoted `#` char
+//		   introduces a comment. So use `'bg-#778899'` not `bg-#778899`. If
+//		   you omit the quotes the text after the `#` char is ignored which
+//		   will likely result in an error or unexpected behavior.
 //
 // -   A color name prefixed by `bg-` to set the background color.
+//
+// -   A color name prefixed by `fg-` to set the foreground color. This has
+// the same effect as specifying the color name without the `fg-` prefix.
 //
 // -   A lambda that receives a styled segment as the only argument and returns a
 // single styled segment.


krader1961:
This bit me in the ass when I put this statement in my *~/.elvish/rc.elv* without quoting the RGB value:

```
chain:segment-style[conda-env] =  [bg-#EEEEEE fg-magenta]
```

krader1961:
Should `styled` RGB values use something other than `#` introduce a 24-bit color value? Is there any ambiguity that requires the `#` prefix? Why not simply test whether the string (minus any `fg-` or `bg-` prefix) consists of six hexadecimal characters?

pkg/eval/builtin_fn_styled.go
```diff
@@ -68,7 +68,12 @@ var errStyledSegmentArgType = errors.New("argument to styled-segment must be a s
 //     -   Any color from the xterm 256-color palette, as `colorX` (such as
 //    `color12`).
 //
-//     -   A 24-bit RGB color, as `#RRGGBB`, such as `#778899`.
+//     -   A 24-bit RGB color written as `#RRGGBB` such as `'#778899'`. Note
```
xiaq, line 71: Please put notes in a separate paragraph starting with `NOTE:`. This makes it clear that the information is non-normative.

pkg/eval/builtin_fn_styled.go
```diff
@@ -68,7 +68,12 @@ var errStyledSegmentArgType = errors.New("argument to styled-segment must be a s
...
+//     that you need to quote such values since the `#` character introduces a
+//     comment if not in a quoted string. So `'#'778899`, `'#778899'`,
```
xiaq, line 73: I prefer `'#778899'` to `'#'778899` for the example.

pkg/eval/builtin_fn_styled.go
krader1961: Done

pkg/eval/builtin_fn_styled.go
krader1961: Done.

krader1961:
I also realized the `fg-` color name prefix wasn't documented so added a bullet item for it as well.

pkg/eval/builtin_fn_styled.go
```diff
@@ -68,10 +68,18 @@ var errStyledSegmentArgType = errors.New("argument to styled-segment must be a s
...
+//		   **NOTE:** You need to quote such values since an unquoted `#` char
+//		   introduces a comment. So use `'bg-#778899'` not `bg-#778899`. If
+//		   you omit the quotes the text after the `#` char is ignored which
+//		   will likely result in an error or unexpected behavior.
 //
 // -   A color name prefixed by `bg-` to set the background color.
 //
+// -   A color name prefixed by `fg-` to set the foreground color. This is the
+// default if `bg-` isn't used and therefore optional.
+//
```
xiaq, line 82: This is a parallel bullet with the color name bullet, not a sub-bullet, so it's ambiguous what "the default" and "optional" means here. IMO it's clearer to say:

> This has the same effect as specifying the color name without the `fg-` prefix.

krader1961: Good point. Done.

xiaq:
Didn't see your comments previously - I feel `#rrggbb` is so common elsewhere that it's worth doing the same in Elvish even if that requires quoting.

I don't like omitting the `#` because some hexdecimal sequences can sound like words which can be quite confusing, `decafe` for example.

PR #1037, opened 2020-05-20
Implement platform hostname and machname variables

### LINKED ISSUE #1028: Add the hostname to the platform module
I like to have the machine name in my prompt. I also predicate some startup configuration on the host name. Lastly, I have some scripts that alter their behavior based on the host name. What I do now is set a `HOST` var that is set to the output of the `hostname` command. It would be nice if this, or the output of `uname -n`, was automatically exposed by the `platform` module. Possibly in two forms: the FQDN and the machine name.

krader1961:
Resolves #1028

### DIFF
--- a/pkg/eval/platform/platform.go
+++ b/pkg/eval/platform/platform.go
@@ -3,7 +3,10 @@
 package platform
 
 import (
+	"errors"
+	"os"
 	"runtime"
+	"strings"
 
 	"github.com/elves/elvish/pkg/eval"
 	"github.com/elves/elvish/pkg/eval/vars"
@@ -35,9 +38,41 @@ import (
 // Whether or not the platform is Microsoft Windows.
 // This is read-only.
 
+//elvdoc:fn hostname &strip-domain=false
+//
+// The host name of the system. If option `&strip-domain` is used any domain
+// components are stripped from the host name. Domain components include the
+// first dot (period) and anything that follows. For example,
+// `machname.subdomain.tld` => `machname`.
+//
+// If an error occurs while fetching the host name an exception is thrown.
+// This is implemented using Go's
+// [`os.Hostname`](https://golang.org/pkg/os/#Hostname) function.
+
+var osHostname = os.Hostname // to allow mocking in unit tests
+var errCannotDetermineHostname = errors.New("cannot determine the hostname")
+
+type hostnameOpt struct{ StripDomain bool }
+
+func (o *hostnameOpt) SetDefaultOptions() {}
+
+func hostname(opts hostnameOpt) (string, error) {
+	hostname, err := osHostname()
+	if err != nil {
+		return "", errCannotDetermineHostname
+	}
+	if !opts.StripDomain {
+		return hostname, nil
+	}
+	parts := strings.SplitN(hostname, ".", 2)
+	return parts[0], nil
+}
+
 var Ns = eval.Ns{
 	"arch":       vars.NewReadOnly(runtime.GOARCH),
 	"os":         vars.NewReadOnly(runtime.GOOS),
 	"is-unix":    vars.NewReadOnly(isUnix),
 	"is-windows": vars.NewReadOnly(isWindows),
-}
+}.AddGoFns("platform:", map[string]interface{}{
+	"hostname": hostname,
+})
--- a/pkg/eval/platform/platform_test.go
+++ b/pkg/eval/platform/platform_test.go
@@ -1,15 +1,34 @@
 package platform
 
 import (
+	"errors"
 	"runtime"
 	"testing"
 
 	"github.com/elves/elvish/pkg/eval"
 )
 
+const (
+	testHostname = "mach1.domain.tld"
+	testMachname = "mach1"
+)
+
+var hostnameFail = true
+
+func hostnameMock() (string, error) {
+	if hostnameFail {
+		hostnameFail = false
+		return "", errors.New("hostname cannot be determined")
+	}
+	return testHostname, nil
+}
+
 var That = eval.That
 
 func TestPlatform(t *testing.T) {
+	savedOsHostname := osHostname
+	osHostname = hostnameMock
+	defer func() { osHostname = savedOsHostname }()
 	setup := func(ev *eval.Evaler) { ev.Builtin.AddNs("platform", Ns) }
 	eval.TestWithSetup(t, setup,
 		That(`put $platform:arch`).Puts(runtime.GOARCH),
@@ -17,5 +36,10 @@ func TestPlatform(t *testing.T) {
 		That(`put $platform:is-windows`).Puts(runtime.GOOS == "windows"),
 		That(`put $platform:is-unix`).Puts(
 			runtime.GOOS != "windows" && runtime.GOOS != "plan9" && runtime.GOOS != "js"),
+		// The first time we invoke the mock it acts as if we can't determine
+		// the hostname. Make sure that is turned into the expected exception.
+		That(`platform:hostname`).ThrowsCause(errCannotDetermineHostname),
+		That(`platform:hostname`).Puts(testHostname),
+		That(`platform:hostname &strip-domain`).Puts(testMachname),
 	)
 }


krader1961:
I've written hundreds of mocks in Python. But, as this is my very first attempt to create a Go unit test mock I make no claims that my attempt is anything like optimal.. Feel free to channel your inner Yoda.

xiaq:
I feel this is doing too much, with questionable benefit. Simply exposing `os.Hostname` as a function is good enough, and much simpler. Unit test will also be unnecessary.

That the first component of a hostname is the "machine name" is a purely a convention, I don't find it necessary to expose in Elvish' stdlib. It's easy enough for a user to do `(str:split . (platform:hostname) | take 1)`.

krader1961:
To the extent that the underlying implementation uses the OS `uname()` API caching the value has limited benefit because the syscall is cheap. The problem is that even the Go implementation falls back to reading */proc*. Too, in the future we might decide to do an actual DNS lookup. Which is also normally relatively cheap but could take a long time. Which is why the example in the cookbook caches the information:

```
# "constantly" returns a function that always writes the same value(s) to
# output; "styled" writes styled output.
edit:rprompt = (constantly (styled (whoami)✸(hostname) inverse))
```

As does the actual implementation in *pkg/edit/prompt.go* (see `func getDefaultPromptVals()`).

> It's easy enough for a user to do `(str:split . (platform:hostname) | take 1)`.

Really? That seems like a silly thing to require the user to do given that most users, most of the time, will want just the machine name, not the FQDN, in their prompt.

krader1961:
> That the first component of a hostname is the "machine name" is a purely a convention,

Yes, it is purely a convention but one that is universal today. At least on any OS that Elvish currently supports. If, and when, Elvish supports a platform where machine names cannot be derived from a host name (i.e., FQDN) by simply splitting on `.` characters and extracting the first element we can decide to change the implementation. Perhaps the documentation of `$platform:machname` should be augmented to make that more clear.

xiaq:
Sorry, I am still not convinced. Exposing `os.Hostname` as a function still sounds good enough to me.

krader1961:
> Exposing os.Hostname as a function still sounds good enough to me.

Okay, then any objection to my changing the `getDefaultPromptVals()` implementation to not cache the hostname? I don't think you can have it both ways.

You also said "Unit test will also be unnecessary." I respectfully disagree. We have plenty of unit tests that do nothing more than verify a Elvish command correctly exposes a Go stdlib function. See, for example, most of the `math:` and `str:` unit tests. Why is exposing the `os.Hostname()` function different?

zzamboni:
I don't have a strong opinion on this topic, but just a suggestion: if both FQDN and "simple" hostname are provided, maybe it would make sense to have it as an option of the same function, instead of two separate ones? `hostname` could have a `&trim` option, for example, which removes everything after the first period (if any).

xiaq:
> Okay, then any objection to my changing the `getDefaultPromptVals()` implementation to not cache the hostname? I don't think you can have it both ways.

UI code such as the default prompt and programming API such as `platform:hostname` are different; there is no need to insist that they must be consistent.

> You also said "Unit test will also be unnecessary." I respectfully disagree. We have plenty of unit tests that do nothing more than verify a Elvish command correctly exposes a Go stdlib function. See, for example, most of the `math:` and `str:` unit tests. Why is exposing the `os.Hostname()` function different?

IIRC most of those tests were added in your PRs; I'd say they are not unnecessary too, but I am not against them either as long as they are harmless.

Back to the "machine name" issue:

* Thinking about this, my primary reservation is about the name: there isn't a strong enough convention to call the unqualified name "machine name". FreeBSD's [manpage](https://www.freebsd.org/cgi/man.cgi?query=hostname&sektion=1) doesn't mention this, and worse, GNU uname has a [--machine](https://manpages.debian.org/stretch/coreutils/uname.1.en.html) that is something different (equivalent to Elvish's `$platform:arch`).

* But I like the idea of adding an option to get the unqualified name; `&trim` is not descriptive enough though, I'd prefer either `$trim-domain` or `&unqualified`.

* If you do that the unit test will be necessary again and you'll need to mock out `os.Hostname`: I suggest simply adding a global function `var osHostname = os.Hostname` and substituting it in tests (and reset it afterwards). This might feel dirty if you are used to full-fledged mocks, but I'd rather keep the tests simple this way.

krader1961:
PTAL

pkg/eval/platform/platform.go
```diff
@@ -35,9 +37,46 @@ import (
...
+	comps := strings.SplitN(hostname, ".", 2)
+	return comps[0]
+}
+
 var Ns = eval.Ns{
 	"arch":       vars.NewReadOnly(runtime.GOARCH),
 	"os":         vars.NewReadOnly(runtime.GOOS),
 	"is-unix":    vars.NewReadOnly(isUnix),
 	"is-windows": vars.NewReadOnly(isWindows),
+}.AddGoFns("platform:", fns)
```
xiaq, line 78: Just inline the `fns` map here.

pkg/eval/platform/platform.go
```diff
@@ -35,9 +37,46 @@ import (
 // Whether or not the platform is Microsoft Windows.
 // This is read-only.
 
+//elvdoc:fn hostname &omit-domain=false
+//
+// The host name of the system. If option `&omit-domain` is used any domain
+// components are stripped from the host name. If an error occurs while
```
xiaq, line 43: You documented the option as "any domain components are stripped from the host name"; `&strip-domain` would be more appropriate.

It's also better to document explicitly what is considered the domain name: anything after the first dot.

pkg/eval/platform/platform.go
```diff
@@ -35,9 +37,46 @@ import (
...
+// fetching the host name `$nil` is substituted. This is implemented using
```
xiaq, line 44: It is better to propagate the error as an exception. If you let `hostname` return `(string, error)`, the error is automatically converted to an exception.

pkg/eval/platform/platform.go
```diff
@@ -35,9 +37,46 @@ import (
...
+// Go's [`os.Hostname()`](https://golang.org/pkg/os/#Hostname) function.
```
xiaq, line 45: In Go suffixing function names with `()` isn't really a thing; just say `os.Hostname`.

pkg/eval/platform/platform_test.go
```diff
@@ -7,15 +7,29 @@ import (
...
+	return testHostname, nil
+}
+
 var That = eval.That
 
 func TestPlatform(t *testing.T) {
-	setup := func(ev *eval.Evaler) { ev.Builtin.AddNs("platform", Ns) }
+	setup := func(ev *eval.Evaler) {
+		ev.Builtin.AddNs("platform", Ns)
+		osHostname = hostnameMock
```
xiaq, line 24: This isn't really part of the setup for the `Evaler`, so better put it outside the `setup` function.

Also, it's good practice to reset `osHostname` to its original value, even if that doesn't really matter here. So I'd write the setup as:

```go
func TestPlatform(t *testing.T) {
    osHostname = hostnameMock
    defer func() { osHostname = os.Hostname }
    setup := ...
    ...
}
```

pkg/eval/platform/platform.go
```diff
@@ -35,9 +37,46 @@ import (
...
+
+//elvdoc:var machname
+//
+// The machine name; i.e., the first component of the hostname.
+// If an error occurs while fetching the machine name `$nil` is substituted.
+// This is implemented using Go's
+// [`os.Hostname()`](https://golang.org/pkg/os/#Hostname) function.
+// This is read-only.
```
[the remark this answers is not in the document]
krader1961: Done.

pkg/eval/platform/platform.go
```diff
@@ -35,9 +37,46 @@ import (
...
+
+var osHostname = os.Hostname // to allow mocking in unit tests
+
+type hostnameOpt struct{ OmitDomain bool }
+
+func (o *hostnameOpt) SetDefaultOptions() { o.OmitDomain = false }
```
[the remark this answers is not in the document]
krader1961: Okay

pkg/eval/platform/platform.go
krader1961: I'm not a fan of that style from a readability standpoint. Especially as more functions get added in the future. But, okay, done.

pkg/eval/platform/platform.go
krader1961: Tomato, Tomahto :-) I think `omit-domain` is perfectly clear but it's not worth arguing. Done.

Good point about explaining what it means to omit or strip the domain. Done.

pkg/eval/platform/platform.go
krader1961: Done.

pkg/eval/platform/platform.go
krader1961: Done, although I am ambivalent. It's not clear that this can ever fail in practice but if it does should that be fatal?

pkg/eval/platform/platform_test.go
krader1961: Done, although I'll point out it should save the original value in a temp var and restore that in the deferred function rather than hardcoding `os.Hostname`. There is no reason for the unit test to be that tightly coupled to the code being tested.

krader1961:
FYI, this passes on my macOS server. I've been seeing every change fail on the Github macOS environment for a few days with this in the log:

```
curl -s https://codecov.io/bash -o codecov && bash codecov -f coverage -t $CODECOV_TOKEN
Exit status: 7
```

Possibly due to out of date security tokens.

krader1961:
I see from the `curl` man page that exit status 7 means "Failed to connect to host." However it's unclear if that is from the `curl` or running the downloaded `codecov` command. If the latter it might still be due to an expired or revoked `$CODECOV_TOKEN`.

pkg/eval/platform/platform.go
```diff
@@ -35,9 +38,41 @@ import (
 // Whether or not the platform is Microsoft Windows.
 // This is read-only.
 
+//elvdoc:fn hostname &strip-domain=false
+//
+// The host name of the system. If option `&strip-domain` is used any domain
```
xiaq, line 43: Function docs should start with a verb, so this should say "Outputs the host name of the system".

pkg/eval/platform/platform.go
```diff
@@ -35,9 +38,41 @@ import (
...
+var errCannotDetermineHostname = errors.New("cannot determine the hostname")
+
+type hostnameOpt struct{ StripDomain bool }
+
+func (o *hostnameOpt) SetDefaultOptions() {}
+
+func hostname(opts hostnameOpt) (string, error) {
+	hostname, err := osHostname()
+	if err != nil {
+		return "", errCannotDetermineHostname
```
xiaq, line 62: It's better to just preserve the original error.

krader1961: I think it's better to wrap the original error but I didn't know how to do that. If the person using the `platform:hostname` command wants to trap just failures to determine the hostname they shouldn't have to know about all the possible ways the underlying implementation can fail.

PR #1038, opened 2020-05-20
Handle SGR codes in byte output in prompt

### LINKED ISSUE #814: Support prompts that output ANSI color sequence
Elvish can parse the ANSI color sequences and turn them into `styled` objects internally.

pontaoski:
This allows prompts such as Starship that output SGR colour sequences to render properly.

ksnip_20200520-003028

Left is how Starship renders without this patch.
Right is how Starship renders with this patch.

Fixes #814.

### DIFF
--- a/pkg/edit/prompt.go
+++ b/pkg/edit/prompt.go
@@ -175,21 +175,17 @@ func parseBytesIntoTokens(b []byte) (ret []rawToken) {
 	return
 }
 
-func parseTokensIntoText(rt []rawToken) ui.Text {
+func parseTokensIntoText(rt []rawToken) []ui.Segment {
 	segments := []ui.Segment{}
 	currentStyle := ui.Style{}
 	for _, token := range rt {
 		if !token.isSGRCode {
 			segments = append(segments, ui.Segment{Style: currentStyle, Text: string(token.data)})
 		} else {
-			currentStyle = ui.StyleFromSGR(strings.TrimPrefix(strings.TrimSuffix(string(token.data), "m"), "\033["))
+			ui.StylingFromSGR(strings.TrimPrefix(strings.TrimSuffix(string(token.data), "m"), "\033[")).Transform(&currentStyle)
 		}
 	}
-	ret := ui.Text{}
-	for idx := range segments {
-		ret = append(ret, &segments[idx])
-	}
-	return ret
+	return segments
 }
 
 // Calls a function with the given arguments and closed input, and concatenates
@@ -223,8 +219,8 @@ func callForStyledText(nt notifier, ev *eval.Evaler, ctx string, fn eval.Callabl
 		if err != nil {
 			nt.notifyf("error reading prompt byte output: %v", err)
 		}
-		if len(parsed) > 0 {
-			add(parsed)
+		for _, seg := range parsed {
+			add(seg)
 		}
 	}
 
--- a/pkg/edit/prompt.go
+++ b/pkg/edit/prompt.go
@@ -4,6 +4,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/user"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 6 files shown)


xiaq:
I prefer a different approach: parse SGR-bracketed texts into [ui.Segment](https://github.com/elves/elvish/blob/8af76083827a4d968d5f35ace2e89dc4dcc573e6/pkg/ui/text_segment.go#L13) objects.

This is because the outputs of prompts can be subject to further manipulations such as [stale transform](https://elv.sh/ref/edit.html#stale-prompt). In the implementation of this PR, the SGR sequences are opaque strings and do not get transformed.

xiaq:
FWIW, there is some code for parsing SGR sequences [here](https://github.com/elves/elvish/blob/8af76083827a4d968d5f35ace2e89dc4dcc573e6/pkg/ui/style.go#L112); it was added for parsing `$LS_COLORS` and can be used to parse SGR sequences.

The approach of the code there isn't entirely right. There are two related types, `ui.Style`, which describes the style and `ui.Styling`, which describes how to change the style. SGR sequences should be parsed into `ui.Styling`, not `ui.Style`, since they can cascade. The difference was irrelevant for parsing `$LS_COLORS` but becomes important for parsing SGR-bracketed texts.

pkg/ui/style_test.go
```diff
@@ -92,7 +92,7 @@ func TestMergeFromOptions(t *testing.T) {
 }
 
 func TestStyleFromSGR(t *testing.T) {
-	tt.Test(t, tt.Fn("StyleFromSGR", StyleFromSGR), tt.Table{
+	tt.Test(t, tt.Fn("StyleFromSGR", StylingFromSGR), tt.Table{
```
xiaq, line 95: Update string here and function name to say `StylingFromSGR`

pkg/ui/style.go
```diff
@@ -109,38 +109,50 @@ var sgrStyling = map[int]Styling{
...
+
 // StyleFromSGR builds a Style from an SGR sequence.
 func StyleFromSGR(s string) Style {
-	style := Style{}
+	return StyleFromStyling(StylingFromSGR(s))
+}
+
+// StylingFromSGR builds a Styling from an SGR sequence.
+func StylingFromSGR(s string) Styling {
+	style := jointStyling{}
```
xiaq, line 126: Rename variable to `styling`.

pkg/ui/style.go
```diff
@@ -109,38 +109,50 @@ var sgrStyling = map[int]Styling{
 	7: Inverse,
 }
 
+// StyleFromStyling builds a Style from a Styling
```
xiaq, line 112: End this comment with a period.

pkg/ui/style.go
```diff
@@ -109,38 +109,50 @@ var sgrStyling = map[int]Styling{
...
 			consume = 3
 		case code == 38 && len(codes) >= 5 && codes[1] == 2:
-			style.Foreground = trueColor{
-				uint8(codes[2]), uint8(codes[3]), uint8(codes[4])}
+			style = append(style, setForeground{trueColor{uint8(codes[2]), uint8(codes[3]), uint8(codes[4])}})
 			consume = 5
 		case code == 48 && len(codes) >= 5 && codes[1] == 2:
-			style.Background = trueColor{
-				uint8(codes[2]), uint8(codes[3]), uint8(codes[4])}
+			style = append(style, setBackground{trueColor{uint8(codes[2]), uint8(codes[3]), uint8(codes[4])}})
```
xiaq, line 155: It's better style to introduce a variable `moreStyling`, assign to it, and centralize the `append` call at the end:

```go
if moreStyling != nil {
    styling = append(styling, moreStyling)
}
```

pkg/edit/prompt.go
```diff
@@ -126,6 +127,67 @@ func defaultStaleTransform(original ui.Text) ui.Text {
...
+				})
+				currentSGRCode = []byte{}
+				currentSGRCode = append(currentSGRCode, '\033', '[')
+				continue
+			}
+		}
+		currentPlaintext = append(currentPlaintext, byt)
+	}
+	return
+}
```
xiaq, line 176: Instead of parsing everything into a token list, I prefer implementing a stateful tokenizer type that produces one token at a time:

```go
func (st sgrTokenizer) next() sgrToken {
}
```

This method is about as easy to use as a list of tokens, but the implementation will be simpler - you can branch into an is-sgr-token branch vs. not-sgr-token branch right away, and avoid the state machine bookkeeping needed in the current implementation.

Please also take care to remove all escape characters that do not form valid SGR sequences.

pkg/edit/prompt.go
```diff
@@ -126,6 +127,67 @@ func defaultStaleTransform(original ui.Text) ui.Text {
 	return ui.StyleText(original, ui.Inverse)
 }
 
+type rawToken struct {
```
xiaq, line 130: This type is declared in the `edit` package, where `rawToken` is not descriptive enough; `sgrToken` is better.

pkg/edit/prompt.go
zzamboni: I would vote for leaving all escape characters unchanged. This would make it possible to emit custom escape sequences from the prompt function, for example to enable Terminal-specific functionality like iTerm2's Shell Integration. Example: https://github.com/zzamboni/elvish-modules/blob/master/iterm2.org

krader1961: @zzamboni, It is not clear why your proposal is needed. Doesn't the solution for issue #1039 address your use-case in a simpler, clearer, fashion? Also, this block of code does not, AFAIK, handle arbitrary escape sequences -- only SGR sequences which should not encompass the sort of terminal specific sequence you're thinking of.

zzamboni: @krader1961 maybe I commented in the wrong place - I am not too familiar with the different types of escape sequences. What I know is what, at the moment, if I try to print some of iTerm2's [custom escape codes](https://www.iterm2.com/documentation-escape-codes.html) without redirecting them to `/dev/tty`, they get printed "escaped", like this:
image

As I understand, the proposal from #1039 would not change this if Elvish is still "escaping" those escape sequences instead of passing them unmodified.

zzamboni: To clarify, and as mentioned in my latest comment in #1060, the problem with having to redirect with `/dev/tty` is that the codes are printed to the terminal when Elvish executes the prompt function to capture its output, and not when that output is actually printed to the screen. In the case of escape codes whose effect depends on where the cursor is when they are printed (like iTerm2's "end of prompt" marker), this makes a difference.

krader1961: @zzamboni, The solution for your situation is to implement a solution for issue #1039. Any escape sequences in a zero-width styled object are ignored by elvish -- as is any other text in such a segment. When you use a zero-width styled object you are promising that it does not change the location of the cursor. Thus there is no need for elvish to interpret the bytes within such a segment in an attempt to determine the width of the output produced by the segment.

xiaq:
@pontaoski Hi! Are you interested in continuing to work on this PR?

If you don't mind, I'll just implement this feature with the approach I proposed instead.

pkg/edit/prompt.go
zzamboni: @krader1961 thanks for the explanation. I agree zero-width segments solve this problem. See my latest comment in #1060 - I still think having a post-prompt hook would be useful but agree that zero-width segments would be far more important.

xiaq:
I have reimplemented the approach used in this PR, closing.

Thanks for your contribution.

PR #1040, opened 2020-05-21
errata for greet-and-bye.elv

kybin:


krader1961:
LGTM (trivial fix of a typographical mistake).

PR #1048, opened 2020-06-01
Make the 0.14 release notes visible

krader1961:
It's useful for people to be able to see the changes included in the
forthcoming release.


xiaq:
It is only intended to be published when 0.14.0 is released so I am going to reject this.

As an alternative, I added links to the draft release notes in both /get and /blog in https://github.com/elves/elvish/commit/af7fe6c596222d175380acb9daddec6ff8d7a8b1.

PR #1052, opened 2020-06-17
Don't start interactive mode if stdin is not a tty

### LINKED ISSUE #661: Do not launch the line editor when reading from standard input
I think it shouldn't launch the editor in this case.
```bash
echo "echo 'string'" | elvish
```
If stdin is not a tty, `bash`, `sh`, and `zsh` reads the script from it:
```bash
echo "echo 'string'" | zsh
```

krader1961:
If stdin is not a tty start a non-interactive shell. The bytes from
stdin will be treated as a script if there are no CLI arguments or the
first argument is "-". Otherwise the stdin stream is treated as data.

Fixes #661

## This will need more changes
The new section on "Running elvish" in the "fundamentals" page will likely benefit from some wordsmithing :smile:

### DIFF
--- a/NEXT-RELEASE.md
+++ b/NEXT-RELEASE.md
@@ -112,6 +112,10 @@ Other improvements:
 
 # Notable bugfixes
 
+-   Running elvish with stdin not attached to a tty no longer attempts to start
+    an interactive shell ([#661](https://b.elv.sh/661)). Depending on the CLI
+    arguments it may treat stdin as a script or data.
+
 -   Invalid option names or values passed to builtin functions now correctly
     trigger an exception, instead of being silently ignored.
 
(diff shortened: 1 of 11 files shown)


krader1961:
I made the single, interactive shell, unit test depend on the replacement for the `github.com/kr/pty` package that was already used elsewhere in this project. I also replaced that sole use of `github.com/kr/pty` with its successor. I could have instead added a second use of `github.com/kr/pty` but since it is no longer being maintained this seemed like a good time to update that dependency.

krader1961:
Closing because I broke something. I was adding tests of the new "shell" pkg functionality and noticed that the tests passed even when I deliberately introduced a bug in a test (something I always due to help ensure my new test has no false-negatives).

PR #1055, opened 2020-06-25
Panic during shell unit tests should be failures

krader1961:
While debugging my fix for issue #661 I noticed that a panic that invokes
the `rescue` function causes Go to think the test passed due to launching
a rescue shell. This introduces a mechanism for elvish unit tests to
request that panics be fatal; i.e., exit with a non-zero status rather
than launch a rescue shell.

P.S., I verified that including this change with my preliminary, incomplete, fix for issue #661 does result in the test failing with the panic info being displayed.

### DIFF
--- a/pkg/shell/script.go
+++ b/pkg/shell/script.go
@@ -24,7 +24,6 @@ type ScriptConfig struct {
 
 // Script executes a shell script.
 func Script(fds [3]*os.File, args []string, cfg *ScriptConfig) int {
-	defer rescue()
 	ev, cleanup := setupShell(fds, cfg.Paths, cfg.SpawnDaemon)
 	defer cleanup()
 
--- a/pkg/shell/shell.go
+++ b/pkg/shell/shell.go
@@ -2,16 +2,13 @@
 package shell
 
 import (
-	"fmt"
 	"os"
 	"os/signal"
 	"strconv"
-	"syscall"
 
 	"github.com/elves/elvish/pkg/cli/term"
 	"github.com/elves/elvish/pkg/eval"
 	"github.com/elves/elvish/pkg/prog"
-	"github.com/elves/elvish/pkg/sys"
 	"github.com/elves/elvish/pkg/util"
 )
 
@@ -90,17 +87,4 @@ func saveEnv(name string) func() {
 		return func() { os.Setenv(name, v) }
 	}
 	return func() { os.Unsetenv(name) }
-}
-
-// Global panic handler.
-func rescue() {
-	r := recover()
-	if r != nil {
-		println()
-		print(sys.DumpStack())
-		println()
-		fmt.Println(r)
-		println("\nExecing recovery shell /bin/sh")
-		syscall.Exec("/bin/sh", []string{"/bin/sh"}, os.Environ())
-	}
 }
--- a/pkg/shell/interact.go
+++ b/pkg/shell/interact.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"syscall"
 	"time"
 
 	"github.com/elves/elvish/pkg/cli"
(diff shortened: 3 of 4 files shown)


pkg/shell/shell.go
```diff
@@ -20,6 +20,11 @@ var logger = util.GetLogger("[shell] ")
 // Program is the shell subprogram.
 var Program prog.Program = program{}
 
+// This is set to true by shell unit tests so that a panic doesn't appear to
+// result in a successful test due to the default behavior of launching a
+// rescue shell.
```
xiaq, line 25: The first sentence of a godoc should always start with the symbol being described, and describe what it is or what it does in general, even if there is only one use case.

In this case, the godoc should start with something like:

```go
// PanicIsFatal controls whether a panic in the shell is considered fatal...
```

Also, this flag is kind of backward - panic being fatal is the default in Go. So instead of a `PanicIsFatal` flag, it's better to introduce a `RescuePanic` flag that defaults `true`, set it to `false` in tests, and make the `defer rescue` call conditional:

```go
if RescuePanic {
    defer rescue()
}
```

krader1961: I'll eventually learn to write Go doc strings correctly :-) I'll fix that wording.

Regarding the var name and how it is used.... My solution doesn't alter the main path and provides a way, in theory, to test that panics are handled correctly; i.e., by running the rescue function modulo not launching a rescue shell. It also opens the possibility of making the rescue shell behavior optional in normal use via a CLI flag or some other mechanism (e.g., a builtin var) while still allowing a lot of Elvish specific context being output when a panic occurs. For example, the current statement being executed, open files, etc. That might prove useful although I'll readily admit I can't provide a real world use case as I write this.

krader1961: I've pushed a change that improves the doc string. Also, looking at how `defer rescue()` is used it seems to me the use in *pkg/shell/script.go* is problematic. At least with respect to the default behavior of launching a rescue shell. That should only occur for interactive shells -- not when executing arbitrary scripts. Under what circumstances does it make sense to launch */bin/sh* as a rescue shell if something like `elvish -c 'put WTF'` results in a panic? I can't help but wonder if that path shouldn't also set `PanicIsFatal` to true. It could, of course, simply not `defer rescue()` but that would eliminate the possibility of providing more context than the default Go panic output.

krader1961: I ran a quick and dirty test by adding a `panic("WTF")` to function `evalInTTY` and ran `./elvish -c 'echo yes'`. Sure enough, it showed me the panic info and also left me in a rescue shell. It's not clear why the `Script()` path is calling `evalInTTY()` but I can't see how dropping into a rescue shell as a result of a panic is correct for that case. I added `PanicIsFatal = true` as the first statement of function `Script` and observed the expected behavior; i.e., elvish specific panic output while the shell still exits with a non-zero status rather than launching a rescue shell. So I'm going to add that to this P.R.

krader1961: Having slept on this I've decided to rename the bool. Primarily because panics are always fatal so whether `PanicIsFatal` is true or false doesn't change the fact the elvish process is terminated when a panic occurs. However, I still feel that it's better to keep the `defer rescue()` and simply predicate the launching of an actual rescue shell on whether the shell is interactive so additional context about the panic can be output (at least in theory).

pkg/shell/script.go
```diff
@@ -24,7 +24,10 @@ type ScriptConfig struct {
 
 // Script executes a shell script.
 func Script(fds [3]*os.File, args []string, cfg *ScriptConfig) int {
-	defer rescue()
+	// For non-interactive shells we want elvish specific output from a panic
+	// but do not want to launch a rescue shell.
+	PanicRescueShell = false
+	defer handlePanic()
```
xiaq, line 30: It's simpler to just not do `defer handlePanic` here. Setting a global here is also fragile.

In case you are keeping this because we *might* want to add more handling than the default panic handler in future, [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

pkg/shell/shell.go
```diff
@@ -20,6 +20,11 @@ var logger = util.GetLogger("[shell] ")
 // Program is the shell subprogram.
 var Program prog.Program = program{}
 
+// PanicRescueShell should be set to false by shell unit tests or any
+// execution path that is not an interactive shell. This keeps a panic from
+// causing a test false positive due the rescue shell.
```
xiaq, line 25: First document what this boolean this, and then its use case. Something like:

```go
// PanicRescueShell determines whether a rescue shell is launched by the panic handler in interactive mode.
//
// It should be set to false ...
```

pkg/shell/script.go
krader1961: I'm familiar with the YAGNI principal. The first major project I worked on at Google had a couple of "features" that greatly complicated the implementation and were not usable for various reasons. Those features were implemented because the author thought they might be useful. I eliminated one of them almost immediately after joining the project because the feature was inherently unusable and causing problems when people tried to use it. The design of the feature didn't address what the users needed and the implementation guaranteed unpredictable behavior.

My hope in doing it this way was to eventually add additional context to the panic data in a subsequent change. Most notably including an Elvish call stack. But after pondering the issue some more I see that isn't going to be practical for several reasons.

Having said all that this particular global flag is not "fragile". Not every global var, or singleton object, is inherently problematic.

xiaq: It *is* problematic - if a program calls `Script` and then `Interact`, the panic handler is turned off for the second `Interact` call.

PR #1063, opened 2020-07-02
website/ref/language.md: Remove mention of exc module

hanche:
A teeny weeny little doc update!

PS. I noticed that the page https://elv.sh/ref/exc.html is still there. Maybe the removal of defunct pages ought to be automated?

### DIFF
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -1753,7 +1753,7 @@ Namespaces that are not special (i.e. one of the above) are also called
 following modules that can be imported by `use`:
 
 -   The following modules are always available: [daemon](daemon.html),
-    [epm](epm.html), [exc](exc.html), [math](math.html),
+    [epm](epm.html), [math](math.html),
     [platform](platform.html), [str](str.html), [re](re.html),
     [readline-binding](readline-binding.html), [store](store.html).
 


hanche:
I think the failed test is not my fault. Rather, it appears to be a problem with the test setup itself?

krader1961:
@hanche, Mind adding the documentation for the `show` command to the builtin module as part of this change? That too was overlooked by the previous refactoring.

xiaq:
@hanche the test failure means that the Markdown file is formatted properly, see https://github.com/elves/elvish/blob/master/CONTRIBUTING.md#formatting-source-files

xiaq:
@krader1961 that's https://elv.sh/ref/builtin.html#show, did I miss anything?

krader1961:
> that's https://elv.sh/ref/builtin.html#show, did I miss anything?

No, I did. I thought I had refreshed my local copy of the webified docs but hadn't. Still, @xiaq, the documentation of the command needs some work as it doesn't explain why someone would use that command. Also, wouldn't it be better to simply special-case exception objects in the `pprint` command? Do we really need another command in addition to `put`, `echo`, `print` and `pprint`?

hanche:
>  the test failure means that the Markdown file is formatted properly

That's a strangely inverted logic … I guess you forgot a “not”. Does the style checker insist on refilling the paragraph? I can't imagine what else could have gone wrong. And IMO, refilling paragraphs is exactly what you don't want to do with source control, as it messes up diffs.

I don't want to install node.js and npm just for this. So I am abandoning this PR. If I were to make more substantial contributions in the future, I'll reconsider.

xiaq:
@hanche Sorry that this hurdle has made you close this PR, I'll just apply this change myself.

Would it be better if I switch to a formatter implemented in Go instead? I imagine you (and most contributors) would find it easier and more acceptable to install a Go binary?

Re reflowing lines: yes, losing clean diffs is not good, but on the other hand:

* I don't want to end up with staggered lines in markdown files

* I don't want to have to pay extra attention to maintaining clean diffs when editing markdown files

Hence I prefer letting the formatter reflow the lines; it's a tradeoff.

xiaq:
@krader1961 `show` is the only command that will write out terminal escape sequences.

Re. the other printing commands:

* `put` is the only one that writes to value output

* `echo` and `print` both prints out stringification of values, they only differ in the presense or lack of a trailing newline. Maybe `print` should be removed in favor of an option on `echo`; this perhaps warrants a poll.

* `repr` prints out **repr**esentation, which is different from the stringification for some values, notably strings. While `pprint` is a pretty version of `repr`, using indentation for nested values. Perhaps `pprint` should be removed in favor of a `pprint` option of `repr`.

hanche:
@xiaq I did it as a pull request because I thought it would be easier for you. Instead, it turned out to be more work for both of us. I suppose I should have read the CONTRIBUTING file first. Sorry about neglecting that.

I think a formatter implemented in go would be more acceptable, yes. If you can find a good one …

xiaq:
The Elvish repo now has its own Markdown formatter, and there is no requirement other than Go for using it now: https://github.com/elves/elvish/blob/master/CONTRIBUTING.md#formatting

PR #1066, opened 2020-07-06
Remove PPA from website

zhsj:
It hasn't been updated since 2018


krader1961:
Would it be better to find a way to continue providing elvish via the Ubuntu PPA mechanism? Would the lack of a PPA for elvish be likely to decrease the number of users to try using it?

xiaq:
An outdated PPA provides negative value, so removing it sounds good.

Would be great if we can add this back when we have an up-to-date PPA though.

PR #1071, opened 2020-07-10
Add gccgo support

zhsj:


krader1961:
What's the rationale for this change?

zhsj:
> What's the rationale for this change?

On debian/ppc64, we build package with gccgo, not gc. It seems it's not too much to afford.

krader1961:
> On debian/ppc64, we build package with gccgo, not gc. It seems it's not too much to afford.

Then that should be the commit comment.

.cirrus.yml
```diff
@@ -33,7 +44,7 @@ test_task:
         # Linux, so losing this part of the data makes very little difference.
         SKIP_UPLOAD_COVERAGE: 1
   go_version_script: go version
-  test_script: go test -race ./...
+  test_script: go test $TEST_FALG ./...
```
krader1961, line 47: Typo here and above: s/TEST_FALG/TEST_FLAG/. You're also replacing `-race` with `-trace` which appears to be another typo.

krader1961:
> On debian/ppc64, we build package with gccgo, not gc. It seems it's not too much to afford.

Since gccgo isn't needed to build elvish anywhere else why do we need a dependency on gccgo for Debian ppc64? That warrants a comment in the relevant section of the *.cirrus.yml* file. Also, why can't the tests be run with race detection enabled on that platform? That should either be fixed so the tests can be run with race detection or a comment added explaining why it is disabled.

.cirrus.yml
```diff
@@ -15,6 +16,16 @@ test_task:
...
+    - name: Test on Linux (Gccgo)
+      container:
+        image: debian:unstable-slim
+      setup_script:
+        - apt-get -y update
+        - apt-get -y install ca-certificates gccgo-go
+      env:
+        # This should have the same coverage as the previous task, so skip it.
+        SKIP_UPLOAD_COVERAGE: 1
+        TEST_FLAG: ""
```
xiaq, line 28: Please add comment explaining why this needs to be disabled.

pkg/eval/exception_unix_test.go
```diff
@@ -17,11 +17,11 @@ func TestExternalCmdExit_Error(t *testing.T) {
 		tt.Args(ExternalCmdExit{0x100, "ls", 1}).Rets("ls exited with 1"),
 		// Note: all Unix'es have SIGINT = 2 and the syscall package has same
 		// string for it ("interrupt").
```
xiaq, line 19: Update this comment, it's no longer true.

.cirrus.yml
```diff
@@ -15,6 +16,17 @@ test_task:
       env:
         # This should have the same coverage as the previous task, so skip it.
         SKIP_UPLOAD_COVERAGE: 1
+    - name: Test on Linux (Gccgo)
```
krader1961, line 19: Maybe name this "Test on Debian using gccgo". Too, add a comment explaining that all programs written in Go are built using gccgo on that platform. Thus the desire to add a CI test to verify that using gccgo to build elvish works.

.cirrus.yml
```diff
@@ -15,6 +16,17 @@ test_task:
...
+      container:
+        image: debian:unstable-slim
+      setup_script:
+        - apt-get -y update
+        - apt-get -y install ca-certificates gccgo-go
+      env:
+        # This should have the same coverage as the previous task, so skip it.
+        SKIP_UPLOAD_COVERAGE: 1
+        # Gccgo doesn't support race test
```
krader1961, line 28: This needs more detail. I googled "gccgo race detection" and could find very few documents that support your assertion gccgo doesn't support race detection. One of those I did find was this comment in the Go source: https://go.googlesource.com/go/+/master/src/cmd/go/go_test.go#244. Adding the term "debian" resulted in no obvious explanations for why this needs to be disabled.

krader1961:
@zhsj, FWIW, I think this change looks good at this point. Notwithstanding my two minor nits about adding comments to better document why this change is needed. It's impressive that your report to the Go project that the [bool](https://github.com/golang/go/issues/40152) was resolved so quickly. No doubt due to the concise reproducer you included in that bug report.

.cirrus.yml
xiaq: The title is fine; the fact that it uses Debian is an implementation detail; the intention is to test gccgo on Linux.

> Too, add a comment explaining that all programs written in Go are built using gccgo on that platform.

That's not true.

.cirrus.yml
xiaq: > Adding the term "debian" resulted in no obvious explanations for why this needs to be disabled.

I can't quite parse this; this comment seems fine to me, it's just documenting a fact, which the source you linked also documented.

PR #1073, opened 2020-07-14
Test io deadlock

krader1961:


pkg/prog/progtest/progtest.go
```diff
@@ -34,7 +56,8 @@ func (f *Fixture) Cleanup() {
 	f.dirCleanup()
 }
 
-// Fds returns the file descriptors in the fixture.
+// Fds returns the file descriptors in the fixture suitable for use by the
+// shell, not the test framework.
```
xiaq, line 60: What's exact meaning of "not the test framework" here - what kind of misuse do you have in mind?

pkg/shell/shell_test.go
```diff
@@ -68,3 +69,21 @@ func testSHLVL(t *testing.T, wantSHLVL string) {
 		t.Errorf("SHLVL existence not restored, %v -> %v", oldOK, newOK)
 	}
 }
+
+// This test is unusual in that its primary purpose is to verify we don't
+// deadlock if the shell produces more output than can be buffered by a pipe
+// or tty.
+func TestScript_Deadlock(t *testing.T) {
+	t.Helper()
```
xiaq, line 77: Remove

pkg/shell/shell_test.go
```diff
@@ -68,3 +69,21 @@ func testSHLVL(t *testing.T, wantSHLVL string) {
 		t.Errorf("SHLVL existence not restored, %v -> %v", oldOK, newOK)
 	}
 }
+
+// This test is unusual in that its primary purpose is to verify we don't
+// deadlock if the shell produces more output than can be buffered by a pipe
+// or tty.
+func TestScript_Deadlock(t *testing.T) {
```
xiaq, line 76: * This test should be in `progtest` since it tests the fixture defined in that package.

  The use of `Script` here is really irrelevant for the purpose of testing that the pipe in the test fixture is able to swallow data without deadline.

* "`TestScript_Deadlock`" is not very informative regarding what it really tests; there are many possibilities of deadlock. Instead of adding a comment on what a test does, use a longer function name: `TestFixture_PipeAcceptsLargeInput`.

pkg/prog/progtest/progtest.go
krader1961: I was just being pedantic that the return value of this method is meant as a suitable (stdin, stdout, stderr) triplet for `shell.Script` and related functions. When I embarked on this work (really the fix for #661) I was initially confused by this method. Since I'm no longer confused I'll just revert this part of the change.

pkg/shell/shell_test.go
krader1961: TL;DR:

Once there is a "real" elvish test that produces a huge amount of output (more than can be buffered by the pipe or tty) this test is redundant.

Longer answer:

I'm happy to give it a better name but I can't see much point in moving this to a *progtest_test.go* module that doesn't currently exist. While this test is to validate the behavior of the test framework its primary purpose is to verify we don't deadlock in the context of an actual elvish shell instance that produces a huge amount of output.

Ideally this would be run before practically any other test but I couldn't find a way to arrange for that to happen. If that prioritization were possible then that would justify moving this into a *progtest_test.go* module. This is, obviously, meant to tell us today whether future tests might deadlock due to producing a large amount of output. If there were such a test, such as I wrote for #661 involving output to a tty, it would deadlock if this test also deadlocked. So in a very real sense this test is redundant. It exists solely to provide confidence that any future elvish test that produces a huge amount of output (such as the one I wrote for issue #661) won't deadlock.

I'd rather drop this test than go to the trouble of moving it to a test module that doesn't currently exist since its value will be short lived.

xiaq: > Once there is a "real" elvish test that produces a huge amount of output (more than can be buffered by the pipe or tty) this test is redundant.

There will always be some redundancy between unit tests and integration tests (the "real" test you refer to). This particular test is a unit test against the `progtest` package.

> I'm happy to give it a better name but I can't see much point in moving this to a progtest_test.go module that doesn't currently exist. While this test validates the behavior of the test framework its primary purpose is to verify we don't deadlock in the context of an actual elvish shell instance that produces a huge amount of output.

Many the unit tests serve double purposes - validating some local property, and ensuring that this local property is useful in a wider context. That doesn't change the fact that unit tests are unit tests and not integration tests.

> I'd rather drop this test than go to the trouble of moving it to a test module that doesn't currently exist since its value will be short lived.

This test will still be useful when integration tests exist; in general redundant unit tests and integration tests *are* useful, because unit test failures are more focused. Imagine that the IO deadlock issue is re-introduced into `progtest`, which will cause both this test (a unit test) and a "real" elvish test (an integration test) to fail. The integration test will be harder to debug since it exercises more code paths, whereas the failure in the unit test provides very focused diagnostics that it's because the `progtest` package is misbehaving.

krader1961: Fair enough. I've moved that test to a new *progtest_test.go* file.

krader1961:
Note that the code coverage failure is because prior to the most recent update to this PR there were no tests of the `progtest` package. This PR introduces partial test coverage that results in a -0.01% overall coverage decrease. Obviously, it would be a good idea to add more `progtest` unit tests but that is outside the scope of this change.

xiaq:
Thanks, merging now.

Re test coverage, I am actually considering excluding all the packages that only provide test utilities from coverage calculation:

* On the one hand, It's not actually possible to test the failure paths. It is impossible to construct a valid `testing.T` outside the testing package (I consider this a defect of the testing package), so you have to pass the instance the test function gets - consequently, exercising the failure path will just fail the test.

* On the other hand, the happy paths are already exercised in actual tests.

An alternative is to let all the test utilities take an interface that `testing.T` satisfies. The tests against test utilities can then pass a mock. This would work if not for [testing.T.Run](https://godoc.org/testing#T.Run) which references `testing.T` itself (which is used by many test utilities). There might be ways to work around this, but it's probably not worth the trouble.

Issue #1060, opened 2020-07-01
New hook `edit:after-prompt`


zzamboni:
The existing `edit:before-readline` and `edit:after-readline` are executed at the very beginning of a new command line (before the prompt), and after the user presses Enter on a complete command to execute it. I would like to propose a new hook `edit:after-prompt` which gets executed after the prompt is drawn and before the start of the command:
```
[before-readline]prompt[after-prompt] command
[after-readline]
```
This can be used to, for example, emit special escape sequences to notify the terminal emulator about where a new command starts (e.g. iTerm2).

This can be achieved at the moment by extending the prompt function itself, but I think a hook would have the following advantages:

- Guarantee that the hooks are executed only once, regardless of how many times the prompt is updated. This is at the moment hard to do, requiring logic in both `edit:prompt` and `edit:after-readline` to keep state and determine when the prompt function is being called for the first time.
- Makes it easier for modules to add code to the hook without having to redefine the `edit:prompt` function, potentially conflicting with themes, other modules, or user configuration.
- Is in line with existing `edit:` hooks.

krader1961:
> This can be used to, for example, emit special escape sequences to notify the terminal emulator about where a new command starts (e.g. iTerm2).

Doesn't the zero-width styled segment proposal in #1039 address this need? If the prompt width changes, thus changing where the command begins, and the proposed `after-prompt` isn't run again wouldn't that confuse the terminal?

zzamboni:
@krader1961 in this particular case, you are correct. In fact, at the moment this is how [iterm2.elv](https://github.com/zzamboni/elvish-modules/blob/master/iterm2.org) works.

(side comment: unfortunately iTerm2 does not seem to handle redrawn prompts gracefully, and even if the codes are re-emitted, it still captures the command wrong if the prompt changes length in the middle of typing. But this is an iTerm2 bug and not related to Evish)

Even if the hook gets executed when the prompt is redrawn (maybe: re-execute it only when the prompt length changes?), I still think it's more elegant for modules to add code to a hook variable than to have to modify the prompt function.

zzamboni:
One more point for this: iTerm2 now supports (at my request, see https://gitlab.com/gnachman/iterm2/-/issues/8996) changing the length of the prompt on the fly. However, this does not work at the moment due to a combination of factors:

[...]

Issue #1039, opened 2020-05-20
Implement a zero-width styled segment


krader1961:
There was a question on Gitter/IM earlier today about how to include sequences in the prompt that don't affect the width of the prompt and are present solely for their side-effects. Such as changing the terminal title or marking the end of the prompt in a manner recognized by the terminal. An example of the latter is the `FTCS_COMMAND_START` sequence recognized by [iTerm2](https://www.iterm2.com/documentation-escape-codes.html).

Zsh does provides for this use case via its `%{...%}` bracketed sequences. The Elvish equivalent would be a special styled segment that indicates the segment promises not to change the location of the cursor.

My suggestion is to add a `zero-width` style transformer; e.g., `styled $obj zero-width`. That transformer cannot be combined with any other transformer because constructs like `styled $obj green zero-width` are nonsensical. @xiaq, however, expressed a preference for a discrete command (`zero-width`?) to make the lack of composability with normal styled objects clearer.

zzamboni:
After all the discussion and clarification in multiple issues (thanks @krader1961 and @xiaq!) I realize this would be the most useful feature at the moment for iTerm2 integration as implemented by [iterm2.elv](https://github.com/zzamboni/elvish-modules/blob/master/iterm2.org).

In fact, given that iTerm2 now allows variable-length prompts by re-emitting the "prompt end" escape code (https://gitlab.com/gnachman/iterm2/-/issues/8996), I'm afraid the next release of iTerm2 will break the current implementation of `iterm2.elv` without the zero-width segments :(

So this issue has a big vote from me 😄

krader1961:
@xiaq, Do you have a strong opinion as to whether this should be implemented via a new `edit:zero-width` command or a `zero-width` transformer to the`builtin:styled` command? I prefer the latter since the point is to make it clear that the result is simply a styled segment that has no visible output; other than as a result of side-effects due to magic escape sequences. If this is implemented in terms of a new `zero-width` command (whatever it is named) that introduces a new data type that has to be handled by `styled`, `echo`, etc. If that command just produces a `styled` object with a "zero width" attribute why use a new command to produce it? If it produces a new, distinct, object then everywhere that consumes a `sytyled-object` has to also handle the new type.

I do not think introducing a distinct "zero-width" type, to indicate lack of composability with the `styled-segment` type, is worth the cost. There are undoubtedly many, existing, cases where combining attributes in a `styled-segment` do not make sense. A zero-width segment is just one such example.

PR #1074, opened 2020-07-15
Proof-of-concept implementation of edit:after-prompt hook

zzamboni:
Initial attempt to address #1060.

**This is in no way ready for merge**, it's meant as an initial proof-of-concept with the intention to promote discussion. I am also not well versed in Go nor in the Elvish codebase. **Any** feedback is very welcome.

The new variable `edit:after-prompt` can contain a list of functions that will be executed every time the prompt is redrawn. Any output produced by the functions (either in the byte or the value stream) must be strings, and these strings will be concatenated and added to the output of `edit:prompt`.

I know there has been some pushback against #1060, which is why I thought it might be a good idea to get a feeling for how the feature might work. As a point of reference, the [`iterm2:init`](https://github.com/zzamboni/elvish-modules/blob/master/iterm2.org#shell-integration) function becomes much cleaner, not having to modify the `edit:prompt` function: 

```
fn init {
  edit:after-prompt = [
    { set-currentdir $pwd >/dev/tty }
    { ftcs-command-start >/dev/tty }
  ]
  edit:before-readline = [
    {
      ftcs-command-finished
      set-remotehost $E:USER (platform:hostname)
      ftcs-prompt
    }
    $@edit:before-readline
  ]
  edit:after-readline = [
    $ftcs-command-executed~
    $@edit:after-readline
  ]
}
```

Other notes:
- At the moment, the output from the functions in `edit:after-prompt` still has to be redirected to `/dev/tty`, otherwise the characters get escaped and displayed instead of sent to the terminal. Implementation of #1039 would make it possible to avoid this by simply returning the corresponding zero-width segments to be displayed after the prompt.
- Unclear what to do about rprompt - at the moment I only added this hook for the regular prompt.

### DIFF
--- a/pkg/cli/app.go
+++ b/pkg/cli/app.go
@@ -48,6 +48,7 @@ type app struct {
 	MaxHeight         func() int
 	RPromptPersistent func() bool
 	BeforeReadline    []func()
+	AfterPrompt       []func() string
 	AfterReadline     []func(string)
 	Highlighter       Highlighter
 	Prompt            Prompt
@@ -87,6 +88,7 @@ func NewApp(spec AppSpec) App {
 		MaxHeight:         spec.MaxHeight,
 		RPromptPersistent: spec.RPromptPersistent,
 		BeforeReadline:    spec.BeforeReadline,
+		AfterPrompt:       spec.AfterPrompt,
 		AfterReadline:     spec.AfterReadline,
 		Highlighter:       spec.Highlighter,
 		Prompt:            spec.Prompt,
--- a/pkg/cli/app_spec.go
+++ b/pkg/cli/app_spec.go
@@ -10,6 +10,7 @@ type AppSpec struct {
 	MaxHeight         func() int
 	RPromptPersistent func() bool
 	BeforeReadline    []func()
+	AfterPrompt       []func() string
 	AfterReadline     []func(string)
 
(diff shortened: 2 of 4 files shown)


krader1961:
I actually like the idea of defining the prompt (whether left or right) in terms of a sequence of functions, rather than a single function, each of whose output is concatenated to compose the prompt. I'm not thrilled about introducing an additional variable that affects how the left prompt is constructed. Especially since it is only useful for output that is implicitly treated as not affecting the visible prompt. See my [comment](https://github.com/elves/elvish/issues/1060#issuecomment-659150358) on your `edit:after-prompt` issue for a counter proposal. So I'm going to punt on reviewing the code in this P.R. until we have some sort of consensus on whether to generalize `edit:prompt` or introduce `edit:after-prompt`.

xiaq:
I'm inclined to rejecting this proposal, but some notes on implementation:

* The after-prompt hook functions should be expected to output `ui.Style` objects, consistent with the prompt itself.

* This API is better implemented entirely on the level of the `edit` package instead of the `cli` package - just make it so that the "prompt" seen by `cli` is really prompt + after-prompt.

zzamboni:
Thanks for the notes! I'm closing this in the spirit of the latest comments at #1060.

PR #1077, opened 2020-07-20
dummy change to trigger CI

krader1961:


krader1961:
Closing since this served its purpose of confirming that the recently added Gccgo CI environment is broken independent of any substantive change to the code.

PR #1079, opened 2020-07-23
Document double-quoted string escape sequences

krader1961:


website/ref/language.md
```diff
@@ -55,30 +55,70 @@ some values. (The traditional terms for the two levels are "commands" and
...
+Unlike most other shells, double-quoted strings in elvish do not support
+interpolation. For instance, `"$name"` simply evaluates to the string `$name`.
+To get a similar effect, simply concatenate strings: instead of
+`"my name is $name"`, write `"my name is "$name`. Under the hood this is a
 [compound expression](#compound-expression-and-braced-lists).
 
-### Barewords
+These are the recognized escape sequences in double-quoted strings:
+
+-   `\cX` converts `X` to an ASCII control character. For example, `\cI` is the
```
xiaq, line 82: Defining this in terms of control characters introduces an additional concept unnecessarily, and I'd prefer this just to be defined in terms of the mechanism (substract 0x40 from the codepoint), and say that this corresponds to control characters as a non-normative comment:

> \cX, where `X` is a character with codepoint between 0x40 and 0x5F, represents the codepoint that is 0x40 lower than X. This corresponds to control character annotations in Unix.

website/ref/language.md
```diff
@@ -55,30 +55,70 @@ some values. (The traditional terms for the two levels are "commands" and
...
+
+-   `\x..` is a Unicode code point represented by two hexadecimal digits.
+
+-   `\u....` is a Unicode code point represented by four hexadecimal digits.
+
+-   `\U......` is a Unicode code point represented by eight hexadecimal digits.
+
+-   The following single character escape sequences:
+
+    -   `\a` is also `\007` or `\x07`, the "bel" character.
```
xiaq, line 103: "is also" is a bit of weird way to *define* something. I'd prefer this to say:

> `\a` is the "bel" character, equivalent to `\007` or `\x07`.

website/ref/language.md
```diff
@@ -55,30 +55,70 @@ some values. (The traditional terms for the two levels are "commands" and
...
-effect, simply concatenate strings: instead of `"my name is $name"`, write
-`"my name is "$name`. Under the hood this is a
+Unlike most other shells, double-quoted strings in elvish do not support
+interpolation. For instance, `"$name"` simply evaluates to the string `$name`.
+To get a similar effect, simply concatenate strings: instead of
+`"my name is $name"`, write `"my name is "$name`. Under the hood this is a
 [compound expression](#compound-expression-and-braced-lists).
 
-### Barewords
+These are the recognized escape sequences in double-quoted strings:
```
xiaq, line 80: An opinionated comment on style: rewrite "there is/are N Ved" to "N is Ved":

> The following escape sequences are recognized in double-quoted strings:

website/ref/language.md
krader1961: Your wording is arguably better but less flexible. That the implementation defines it in terms of subtracting 0x40 is not consistent with the historical definition. Specifically, masking the character with 0x1F so that `\ca` and `\cA` are equivalent. I was surprised when looking at the implementation to find that `\ca` isn't legal. My original text did mention masking with 0x1F. I probably should have retained that text. I would argue that the implementation should be relaxed to allow lowercase letters.

krader1961: Note that bash, as just one example, uses `#define TOCTRL(x) ((x) == '?' ? 0x7f : (TOUPPER(x) & 0x1f))` and `#define control_character_mask 0x1f`. Historically, various terminals disallowed applying the control modifier to some characters (such as `;`) and, as bash does, treat `\c?` specially as a synonym for 0x7F (del) rather than just masking `?` with 0x1F to produce 0x0F (si).

krader1961: Unsurprisingly, Wikipedia has a pretty decent [article](https://en.wikipedia.org/wiki/Control_character) concerning ASCII control characters. My opinion is that Elvish should handle this similar to bash but with `\c ` (i.e., control space) as a special-case that maps to `\c@` (aka nul). Just as `\c?` is a special-case for 0x7F (aka del). That should provide maximal compatibility with existing implementations and least surprise for users. Especially with respect to upper versus lowercase letters.

krader1961:
PTAL. Rather than simply change the wording for `\c` sequences I opted to modify the implementation. Specifically, to do a `&0x1F` mask operation on any non-control ASCII character and special case `\c?` as every other program (including terminal emulators) do.

xiaq:
> Specifically, to do a &0x1F mask operation on any non-control ASCII character and special case \c? as every other program (including terminal emulators) do.

I find it unnecessary to extend the accepted characters. There are only 32 control characters (0x0 to 0x1F), and each of them has a traditional caret notation using their counterparts from 0x40 to 0x5F. Is there any prior art for allowing characters outside 0x40 to 0x5F?

I don't dispute supporting mapping ? to DEL though.

krader1961:
> Is there any prior art for allowing characters outside 0x40 to 0x5F?

Yes. Starting with any terminal you've ever used. You don't have to press [shift-ctrl-a] as that is equivalent to [ctrl-a]. On most (all?) terminals pressing [ctrl-space] is the same as [ctrl-@]. Although it is true that terminals are inconsistent since [ctrl-backtick] should also be equivalent to those two examples.

See also the Wikipedia [article](https://en.wikipedia.org/wiki/Control_character) on control characters. Specifically this sentence:

> The other implementation is to take the ASCII code produced by the key and bitwise AND it with 31, forcing bits 6 and 7 to zero.

Other shells also allow using the full range of non-control ASCII chars. For example, bash:

```
bash-5.0$ echo $'\c!' | god -tx1z
0000000 01 0a                                            >..<
0000002
bash-5.0$ echo $'\ca' | god -tx1z
0000000 01 0a                                            >..<
0000002
bash-5.0$ echo $'\cA' | god -tx1z
0000000 01 0a                                            >..<
0000002
```

The Korn shell is weird. Lower and upper case letters are equivalent but punctuation chars get converted to lowercase letters:

```
KSH PROMPT:3: echo $'\c!' | god -tx1z
0000000 61 0a                                            >a.<
0000002
KSH PROMPT:4: echo $'\ca' | god -tx1z
0000000 01 0a                                            >..<
0000002
KSH PROMPT:5: echo $'\cA' | god -tx1z
0000000 01 0a                                            >..<
0000002
```

krader1961:
My new control literal implementation is backward compatible with the existing implementation. There is no good reason to allow `\cA` while disallowing `\ca`. That distinction suggests that number literals like `0xab` should be treated differently than `0xAB`.

xiaq:
Thanks for the historical context, but I am still not convinced that the new behavior is necessary.

> Yes. Starting with any terminal you've ever used. You don't have to press [shift-ctrl-a] as that is equivalent to [ctrl-a]. On most (all?) terminals pressing [ctrl-space] is the same as [ctrl-@]. Although it is true that terminals are inconsistent since [ctrl-backtick] should also be equivalent to those two examples.

The correspondence only applies to letters; `Ctrl-0` does not generate 0x10, for example.

Systematic correspondence between the caret notation and the keyboard is a lost cause in 2020.

It really sounds to me that mapping Ctrl to "& 0x1F" was a smart trick invented by authors of terminal drivers or emulators to make sure that Ctrl-[Any key] has a meaning. But the trick is a bit too smart for my taste.

> The Korn shell is weird. Lower and upper case letters are equivalent but punctuation chars get converted to lowercase letters

That convinces me that since there is no actual agreement in other shells on how characters outside 0x40 to 0x5F behaves, Elvish should just stick to the smaller range.

> There is no good reason to allow \cA while disallowing \ca.

There isn't really a good reason to allow both; `^A` is a canonical way to write 0x1, understood everywhere, whereas `^a` not nearly so. It is just an extra thing for users to learn.

krader1961:
> But the trick is a bit too smart for my taste.

Honestly, I find the "subtract 64" to be the trickier, and less general, of the two implementations. This is a case of "you say tomato, I say tomahto" (lyric from a famous song). I'll revert to the subtract 64 on a narrower range since it's more important to document what is supported than implement the solution I think makes more sense. :smile:

krader1961:
The "check the style of .md files" test failed yesterday with this reason:

```
make checkstyle-md
Markdown files that need formatting:
  website/ref/language.md
Makefile:34: recipe for target 'checkstyle-md' failed
make: *** [checkstyle-md] Error 1

Exit status: 2
```

The `prettier` command I had installed on my workstation was version 2.0.2. I upgrade to prettier version 2.0.5 with significant effort since I don't normally use the `npm` command. Running the upgraded `prettier` against my change produced this diff:

```
diff --git a/website/ref/language.md b/website/ref/language.md
index 616d261f..e6e9e200 100644
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -80,8 +80,8 @@ To get a similar effect, simply concatenate strings: instead of
 The following escape sequences are recognized in double-quoted strings:

 -   `\cX`, where _X_ is a character with codepoint between 0x40 and 0x5F,
-    represents the codepoint that is 0x40 lower than _X_. For example, `\cI`
-    is the tab character: 0x49 (`I`) - 0x40 = 0x09 (tab). There is one special
+    represents the codepoint that is 0x40 lower than _X_. For example, `\cI` is
+    the tab character: 0x49 (`I`) - 0x40 = 0x09 (tab). There is one special
     case: A question-mark is converted to del; i.e., `\c?` or `\^?` is
     equivalent to `\x7F`.
```

Force pushing that change resolved the test "failure". Notice that the updated (2.0.2 to 2.0.5) `prettier` command did nothing more than move a single word to the previous line.

@xiaq, This is a pretty painful mechanism for contributors. There needs to be more clarity regarding the `prettier` version being used by the CI environment and how a contributor can verify the version they are using and how to upgrade their version.

krader1961:
@xiaq, Consider adding this link to the contributing documentation or a condensed version of the information within it: `https://docs.npmjs.com/updating-packages-downloaded-from-the-registry`. I "solved" my problem by deleting the installed `prettier` package and reinstalling it. But even that blunt force solution took additional steps due to the */usr/local/bin/prettier* symlink changing which the reinstall could not handle.

xiaq:
Huh, that's some bad practice from prettier - I hope it doesn't become too common. I'll document the version of prettier to use and pin the version in CI too.

PR #1080, opened 2020-07-23
Remove two obsolete TODO comments and resolve one

krader1961:


pkg/eval/builtin_fn_num.go
```diff
@@ -9,7 +9,18 @@ import (
...
+//
+// ```elvish
+// float64 123
+// float64 NaN
+// float64 Inf
+// ```
+//
+// Explicitly convert a string to a `float64` data type.
+//
+// It is very rare that this command is needed. See the discussion of the
```
xiaq, line 22: "Very rare" is a bit too strong; I'd rather this says "this command is often not needed".

krader1961: Fair enough.

PR #1081, opened 2020-07-23
Add str:from-utf8-bytes, str:to-utf8-bytes and move builtin ord and chr to str module

raetiacorvus:


pkg/eval/str/str.go
```diff
@@ -331,6 +389,37 @@ func split(fm *eval.Frame, opts maxOpt, sep, s string) {
 // ▶ ABC!123
 // ```
 
+//elvdoc:fn to-utf8-bytes
```
krader1961, line 420: It looks like you refactored the change into two commits and this portion of this commit was meant to be in the second commit. Oops! :smile: Since I know what a PITA this type of refactoring is, and both commits are small and closely related I would just squash them into a single commit.

pkg/eval/str/str_test.go
```diff
@@ -69,6 +73,9 @@ func TestStr(t *testing.T) {
 		That(`str:to-codepoints a`).Puts("0x61"),
 		That(`str:to-codepoints 你好`).Puts("0x4f60", "0x597d"),
 
+		That(`str:to-utf8-bytes a`).Puts("0x61"),
+		That(`str:to-utf8-bytes 你好`).Puts("0xe4", "0xbd", "0xa0", "0xe5", "0xa5", "0xbd"),
```
krader1961, line 77: A string can contain bytes that are not part of a valid UTF8 sequence. There should be a test for that case.

krader1961: Ignore my previous comment. While what I wrote is technically true it is also true that the Go runtime already handles this and has its own tests that it is handled correctly. Adding an elvish test for that case wouldn't provide any benefit.

krader1961:
LGTM other than the code refactoring to either move the `to-utf8-bytes` implementation into the second commit or just squash merge the commits.

pkg/eval/str/str.go
raetiacorvus: I am really sorry for this, probably happened when i tried to fix the formatting with a rebase

pkg/eval/str/str.go
```diff
@@ -92,6 +95,62 @@ import (
...
+// ~> str:from-utf8-bytes 0xe4 0xbd 0xa0 0xe5 0xa5 0xbd
+// ▶ 你好
+// ```
+//
+// @cf str:to-utf8-bytes
+
+func fromUtf8Bytes(nums ...int) (string, error) {
+	var b bytes.Buffer
+	for _, num := range nums {
+		b.WriteByte(byte(num))
```
xiaq, line 146: Check if num is within the range of byte, and return an [OutOfRange](https://github.com/elves/elvish/blob/9858d5a5455ab157934d84b118a4d63c45dbfcd4/pkg/eval/errs/errs.go#L9) error if it isn't. Also add a test for this.

pkg/eval/str/str.go
```diff
@@ -92,6 +95,62 @@ import (
...
+// ▶ a
+// ~> str:from-codepoints 0x4f60 0x597d
+// ▶ 你好
+// ```
+//
+// @cf str:to-codepoints
+
+func fromCodepoints(nums ...int) (string, error) {
+	var b bytes.Buffer
+	for _, num := range nums {
```
xiaq, line 117: Check if num is within the range of rune, and return an [OutOfRange](https://github.com/elves/elvish/blob/9858d5a5455ab157934d84b118a4d63c45dbfcd4/pkg/eval/errs/errs.go#L9) error if it isn't. Also add a test for this.

pkg/eval/str/str_test.go
```diff
@@ -31,6 +31,14 @@ func TestStr(t *testing.T) {
 		That(`str:equal-fold abc ABC`).Puts(true),
 		That(`str:equal-fold abc A`).Puts(false),
 
+		That(`str:from-codepoints 0x61`).Puts("a"),
+		That(`str:from-codepoints 0x4f60 0x597d`).Puts("你好"),
+		That(`str:from-codepoints -1`).ThrowsAny(),
```
xiaq, line 36: Don't use ThrowAny; test the exact exception thrown.

pkg/eval/str/str_test.go
```diff
@@ -31,6 +31,14 @@ func TestStr(t *testing.T) {
...
+
+		That(`str:from-utf8-bytes 0x61`).Puts("a"),
+		That(`str:from-utf8-bytes 0xe4 0xbd 0xa0 0xe5 0xa5 0xbd`).Puts("你好"),
+		That(`str:from-utf8-bytes -1`).ThrowsAny(),
```
xiaq, line 40: Same.

pkg/eval/str/str.go
raetiacorvus: Done. I used int(byte(num)) = num and int(rune(num)) num for the overflow check, not sure if your would preferred `<` and `>` checks for the range boundaries.

pkg/eval/str/str_test.go
krader1961: I wave my magic wand and mutate the existing 42 uses of `ThrowsAny()` to instead check for a specific exception. :smile: On a more serious note I'll point out that the "throw" related test methods in *pkg/eval/testutils.go* would benefit from some documentation. If nothing else that `ThrowsAny()` is deprecated.

pkg/eval/str/str.go
xiaq: Normally I wouldn't have an opinion, but since you need to spell out the range in the error anyway I'd prefer using comparison operators.

pkg/eval/str/str.go
```diff
@@ -92,6 +95,74 @@ import (
...
+		if int(rune(num)) != num {
+			return "", errs.OutOfRange{
+				What:     "codepoint",
+				ValidLow: -2147483648, ValidHigh: 2147483647,
+				Actual: strconv.Itoa(num)}
+		}
+		if !utf8.ValidRune(rune(num)) {
+			return "", fmt.Errorf("invalid codepoint: %d", num)
```
xiaq, line 125: Return a BadValue struct here (https://github.com/elves/elvish/blob/a3f66b52af3c659af21596f0d44b76b2f41a0048/pkg/eval/errs/errs.go#L27)

Yoy can use "argument to str:from-codepoints" as "What" and "valid Unicode codepoint" as "Valid".

pkg/eval/str/str.go
```diff
@@ -92,6 +95,74 @@ import (
...
+		if int(byte(num)) != num {
+			return "", errs.OutOfRange{
+				What:     "byte",
+				ValidLow: 0, ValidHigh: 255,
+				Actual: strconv.Itoa(num)}
+		}
+		b.WriteByte(byte(num))
+	}
+	if !utf8.Valid(b.Bytes()) {
+		return "", fmt.Errorf("invalid utf8 bytes: %d", b.Bytes())
```
xiaq, line 161: Return a BadValue struct here. You can use "arguments to str:from-utf8-bytes" as What and "valid UTF-8 sequence" as Valid.

pkg/eval/str/str.go
```diff
@@ -92,6 +96,80 @@ import (
...
+// @cf str:to-codepoints
+
+func fromCodepoints(nums ...int) (string, error) {
+	var b bytes.Buffer
+	for _, num := range nums {
+		if num < 0 || num > unicode.MaxRune {
+			return "", errs.OutOfRange{
+				What:     "codepoint",
+				ValidLow: 0, ValidHigh: unicode.MaxRune,
+				Actual: strconv.Itoa(num)}
```
xiaq, line 123: Be consistent: always format codepoints in hexadecimal.

rautesamtr:
> I have a few minor comments, but I'll merge this PR and do a small fixup myself.
> 
> A lot of thanks, and congrats on your first contribution :)

Thanks for all your patience in getting this PR into shape. Sorry for the remaining problems.

PR #1082, opened 2020-07-25
Document `break`, `continue`, and `return`

krader1961:
Since `break`, `continue`, and `return` are builtin commands, rather
than language keywords, they should be documented on the builtin module
reference page.

### DIFF
--- a/pkg/eval/builtin_fn_flow.go
+++ b/pkg/eval/builtin_fn_flow.go
@@ -50,7 +50,85 @@ import (
 
 // TODO(xiaq): Document "multi-error".
 
-// TODO(xiaq): Document "return", "break" and "continue".
+//elvdoc:fn break
+//
+// Raises the special "break" exception. When raised inside a loop it is
+// captured and causes the loop to terminate.
+//
+// Because `break` raises an exception it can be caught by a
+// [`try`](language.html#exception-control-try) block. If not caught, either
+// implicitly by a loop or explicitly, it causes a failure like any other
+// uncaught exception.
+//
+// See the discussion about [flow commands and exceptions](language.html#exception-and-flow-commands)
+//
+// **Note**: You can create a `break` function and it will shadow the builtin
+// command. If you do so you should explicitly invoke the builtin. For example:
+//
+// ```elvish-transcript
+// > fn break []{ put 'break'; builtin:break; put 'should not appear' }
+// > for x [a b c] { put $x; break; put 'unexpected' }
+// ▶ a
+// ▶ break
+// ```
+
+//elvdoc:fn continue
+//
+// Raises the special "continue" exception. When raised inside a loop it is
+// captured and causes the loop to begin its next iteration.
+//
+// Because `continue` raises an exception it can be caught by a
+// [`try`](language.html#exception-control-try) block. If not caught, either
+// implicitly by a loop or explicitly, it causes a failure like any other
+// uncaught exception.
+//
+// See the discussion about [flow commands and exceptions](language.html#exception-and-flow-commands)
+//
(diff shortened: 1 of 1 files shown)


pkg/eval/builtin_fn_flow.go
```diff
@@ -50,7 +50,73 @@ import (
 
 // TODO(xiaq): Document "multi-error".
 
-// TODO(xiaq): Document "return", "break" and "continue".
+//elvdoc:fn break
+//
+// Raise the special "break" exception. Loops capture this exception and cause
+// the loop to terminate. Because it is an exception it can be caught by a
```
xiaq, line 56: > Raise the special "break" exception

Raise -> Raises. See https://github.com/elves/elvish/blob/master/CONTRIBUTING.md#reference-docs: The first sentence of the description should start with a verb in 3rd person singular (i.e. ending with a "s"), as if there is an implicit subject "this function".

> Loops capture this exception and cause the loop to terminate

The subject of this sentence is "loops", so the second half of this sentence is grammatically equivalent to "loops cause the loop to terminate". Here are some alternatives:

* Loops capture this exception, which causes the loop to terminate

* When raised inside a loop, it is captured by the loop and causes the loop to terminate

> Because it is an exception

Break the paragraph before this sentence: the first sentence alone is a complete specification of the behavior of `break`; the rest are comments.

pkg/eval/builtin_fn_flow.go
```diff
@@ -50,7 +50,73 @@ import (
...
+// > fn break []{ put break; builtin:break; put 'should not appear' }
+// > for x [a b c] { put $x; break }
+// ▶ a
+// ▶ break
+// ```
+
+//elvdoc:fn continue
+//
+// Raise the special "continue" exception. Loops capture this exception and
+// cause the loop to begin its next iteration. Because it is an exception it
```
xiaq, line 76: Same 3 comments as above.

pkg/eval/builtin_fn_flow.go
```diff
@@ -50,7 +50,73 @@ import (
...
+// ▶ b
+// ▶ continue
+// ▶ c
+// ▶ continue
+// ```
+
+//elvdoc:fn return
+//
+// Raise the special "return" exception. Functions capture this exception and
+// cause the function to terminate. Because it is an exception it can be
```
xiaq, line 100: Same 3 comments as above.

Moreover, `return` exceptions are only captured by `fn`-defined functions, **not** anonymous functions; document that.

pkg/eval/builtin_fn_flow.go
```diff
@@ -50,7 +50,73 @@ import (
...
+// not appear in the output:
+//
+// ```elvish-transcript
+// > fn return []{ put return; builtin:return; put 'should not appear' }
+// > fn test-return []{ put yes; return; put no }
+// > test-return
+// ▶ yes
+// ▶ return
+// ▶ no
+// ```
```
xiaq, line 119: This example really shows two things: (1) the fact that you can shadow the builtin `return`, and that (2) you can't create a wrapper for the `return` command using a `fn`-defined function, since the inner function itself will capture the exception. It's better to break this into two examples.

krader1961: That is true. Although probably not in the fashion you expect. Using a lambda results in an infinite loop:

```
> return~ = []{ put return; builtin:return; put 'should not appear' }
> fn test-return []{ put yes; return; put no }
> test-return
▶ yes
▶ return
▶ return
▶ return
[ad infinitum]
```

That seems like a bug. Nonetheless, I don't see the point in creating two examples since this is a "**do not do this**" scenario as explained by the preceding paragraph.

Issue #849, opened 2019-08-11
Support a `[count:x]` wildcard qualifier


xiaq:
Should work similarly to regular expression's `{n}`, matching number of codepoints, with support for ranges. Examples:

* `*[count:5]` matches 5 codepoints
* `*[count:5-10]` matches 5 to 10 codepoints
* `*[count:5-]` matches 5 or more codepoints
* `*[count:-5]` matches 5 or fewer codepoints

I don't quite like the last example though, as `-5` looks like negative five. It is technically unambiguous in this case, but still weird.

xiaq:
Idea and proposed syntax courtesy of @hanche

krader1961:
Rather than introducing another, seldom used, modifier I think allowing a modifier to be a lambda is preferable. Yes, I know I have a PR #1018 (issue #1015) under review to introduce a `type:` modifier :smile: But I expect that to be used at least two orders of magnitude more often than filtering on the length of of the file name. Allowing a lambda lets the user use Elvish to filter the list of pathnames.

xiaq:
This is a local qualifier on how many characters a wildcard matches, not a global modifier on the length of the full filename.

krader1961:
> This is a local qualifier on how many characters a wildcard matches, not a global modifier on the length of the full filename.

Understood, but it is still a niche feature that can be implemented via a callback. The callback can easily test the length of the full path or just the final component.

krader1961:
Note that it appears that most of these niche wildcard modifiers exist in bash/ksh/zsh are because those shells don't support a callback to decide if the expansion should be accepted or rejected.

hanche:
If this is not implemented, I think at least I'd prefer a replacement for likely the most common occurrence: `[count:1-]` meaning “at least one”. So `*[set:ab][nonempty]` could match one or more `a` or `b`, while `?[set:ab][nonempty]` could match exactly one of those.

hanche:
Here's another issue. How do you write a wildcard pattern that will match filenames containing at least one digit? At present, I don't think you can. But, assuming that a quoted empty string will separate adjacent wildcards, you could say `*''*[digit][count:1]*` or `*''?[digit][nonempty]*`. To make it look a little less hacky, we might allow empty modifiers instead: `*[]*[digit][count:1]*`.

In other shells, this is trivial, as in `*[0-9]*`. (I have to admit that the wildcards are my least favourite part of the elvish language. But the added power of modifiers is nice, so I am conflicted about them.)

[...]

PR #1084, opened 2020-07-27
Introduce new path: module

kolbycrouch:
Try to introduce a new `path:` module for path evaluating built-ins.

Starts by simply moving `path-/*` and `eval-symlinks` from the built-in namespace.

### DIFF
--- a/NEXT-RELEASE.md
+++ b/NEXT-RELEASE.md
@@ -26,6 +26,19 @@ or compiled, even if it is not executed:
 
 -   The `-source` command is now deprecated. Use `eval` instead.
 
+-   The `eval-symlinks` command is now deprecated. Use `path:eval-symlinks`
+    instead.
+
+-   The `path-abs` command is now deprecated. Use `path:abs` instead.
+
+-   The `path-base` command is now deprecated. Use `path:base` instead.
+
+-   The `path-clean` command is now deprecated. Use `path:clean` instead.
+
+-   The `path-dir` command is now deprecated. Use `path:dir` instead.
+
+-   The `path-ext` command is now deprecated. Use `path:ext` instead.
+
 The following deprecated features trigger a warning when the code is evaluated:
 
 -   Using `:` in slice indicies is deprecated. Use `..` instead.
@@ -44,6 +57,8 @@ New features in the standard library:
 
 -   A new `eval` command supports evaluating a dynamic piece of code in a
     restricted namespace.
+
+-   A new `path:` module has been introduced.
 
 # Notable bugfixes
 
--- a/pkg/eval/builtin_fn_fs.go
+++ b/pkg/eval/builtin_fn_fs.go
@@ -26,6 +26,8 @@ var ErrStoreNotConnected = errors.New("store not connected")
 //
 // See [godoc of path/filepath](https://godoc.org/path/filepath). Go errors are
 // turned into exceptions.
+//
+// These functions are deprecated; use [path:](path.html)
 
 // TODO(xiaq): Document eval-symlinks.
 
--- a/pkg/eval/path/path.go
+++ b/pkg/eval/path/path.go
@@ -1,0 +1,94 @@
+// Package path exposes functionality from Go's path package as an Elvish
+// module.
+package path
+
+import (
+	"path/filepath"
+
+	"github.com/elves/elvish/pkg/eval"
+)
+
+//elvdoc:fn abs
+//
+// ```elvish
+// path:abs $path
+// ```
+//
+// Outputs the absolute path of `$path`.
+//
+// ```elvish-transcript
+// ~> path:abs ~/bin
+// ▶ /home/user/bin
+// ```
+
+//elvdoc:fn base
+//
+// ```elvish
+// path:base $path
+// ```
+//
+// Outputs the base path of `$path`.
+//
+// ```elvish-transcript
+// ~> path:base ~/bin
+// ▶ bin
+// ```
+
+//elvdoc:fn clean
+//
+// ```elvish
+// path:clean $path
+// ```
+//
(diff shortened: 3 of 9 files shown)


xiaq:
Please make the code or test compatible with Windows.

codyopel:
It is unnecessary to prefix functions with `path-` when they are in a separate namespace.

pkg/eval/builtin_fn_fs.go
```diff
@@ -40,21 +39,6 @@ var ErrStoreNotConnected = errors.New("store not connected")
 // ▶ [&path=/Users/foo/.elvish &score=96.79928]
 // ```
 
-//elvdoc:fn path-\*
```
krader1961, line 43: You should not remove the documentation for these builtins. You're deprecating them in this commit; not removing them.

pkg/eval/builtin_fn_fs.go
```diff
@@ -84,13 +68,7 @@ func init() {
 		"dir-history": dirs,
 
 		// Path
-		"path-abs":      filepath.Abs,
```
krader1961, line 87: Do not remove existing builtins when moving the functionality to a new namespace. With this change they should be deprecated; not removed. That provides a transition period so as not to needlessly break existing users of the Elvish shell.

pkg/eval/path/path.go
```diff
@@ -0,0 +1,41 @@
...
+// eval-symlinks $path
+// ```
+//
+// Output the path name of $path after the evaluation of symbolic links.
+
+var Ns = eval.Ns{}.AddGoFns("path:", fns)
+
+var fns = map[string]interface{}{
+	"eval-symlinks": filepath.EvalSymlinks,
+	"path-abs":      filepath.Abs,
```
krader1961, line 36: Avoid "stuttering". That is, rather than `path:path-abs` prefer `path:abs`.

krader1961:
@kolbycrouch, I hope you have not abandoned this pull-request. It is a worthwhile change. It simply has a few problems that are to be expected for the initial version of any change of this nature.

pkg/eval/path/path.go
```diff
@@ -0,0 +1,94 @@
...
+// module.
+package path
+
+import (
+	"path/filepath"
+
+	"github.com/elves/elvish/pkg/eval"
+)
+
+//elvdoc:fn abs
```
xiaq, line 11: Most of these doc strings are not detailed enough, compared to the godoc counterparts. Either add links to the godoc, or adapt the godoc.

pkg/eval/path/path_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"testing"
+
+	"github.com/elves/elvish/pkg/eval"
+)
+
+var That = eval.That
+
+func TestStr(t *testing.T) {
```
xiaq, line 11: TestPath

website/ref/path.md
```diff
@@ -0,0 +1,10 @@
+<!-- toc -->
+
+# Introduction
+
+The `path:` module provides path evaluation functions.
```
xiaq, line 5: "Path evaluation functions" doesn't sound right, "evaluating a path" isn't a well-defined thing. "Path manipulation functions" would be better.

pkg/eval/builtin_fn_fs.go
```diff
@@ -26,6 +26,8 @@ var ErrStoreNotConnected = errors.New("store not connected")
 //
 // See [godoc of path/filepath](https://godoc.org/path/filepath). Go errors are
 // turned into exceptions.
+//
+// These functions are deprecated; use [path:](path.html)
```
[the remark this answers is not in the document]
xiaq: +1

NEXT-RELEASE.md
```diff
@@ -26,6 +26,19 @@ or compiled, even if it is not executed:
 
 -   The `-source` command is now deprecated. Use `eval` instead.
 
+-   The `eval-symlinks` command is now deprecated. Use `path:eval-symlinks`
```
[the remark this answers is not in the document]
xiaq: Hmm, I've seen both in release notes of other software and to me "is now deprecated" signifies that "this feature is deprecated since this release" (as opposed to already deprecated before).

Googling ["release notes" "is deprecated"](https://www.google.co.uk/search?q=%22release+notes%22+%22is+deprecated%22) and ["release notes" "is now deprecated"](https://www.google.co.uk/search?q=%22release+notes%22+%22is+now+deprecated%22) suggests that release notes do seem to say "is now deprecated" (178k vs 328k results).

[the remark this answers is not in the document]
krader1961: > I've seen both in release notes of other software...

Yes, because English is a slippery language with lots of quirks; perhaps more so than any other human language. However, in this case the "now" in "is now deprecated" formulation is redundant. If it was deprecated in an earlier release but not documented then this should say something like "was deprecated, albeit not formally documented, in version $whatever." But is obviously silly since it doesn't matter that the intent was to deprecate it in a prior release if it wasn't formally documented as being deprecated in that prior release. You can't tell users today to pretend you documented this was deprecated six months ago.

My name is Don Quixote :smile:

xiaq:
Thanks for the interest in contributing the `path` module, but since this PR hasn't been active for a while, I am closing this.

Please do feel free to re-open or open another PR if you decide to resume working on this.

krader1961:
I am sad that @kolbycrouch abandoned this change. I really want to see something like this change as a stepping stone towards resolving issue #849 without introducing more wildcard qualifiers and instead using a more general callback mechanism. I'll create a P.R. based on this change that addresses the unresolved feedback on this P.R.

PR #1092, opened 2020-08-04
Fix `exec` handling of $E:SHLVL

### LINKED ISSUE #1091: `exec` command needs to special-case `$E:SHLVL`
I was moving more of my pre-Elvish interactive configuration into my Elvish *~/.elvish/rc.elv* file and noticed that doing `exec elvish` incremented `$E:SHLVL`. That is wrong since I'm replacing the current shell with a different shell at the same level. The `exec` command should decrement the value of `$E:SHLVL`. This is what bash does (see the `exec_builtin()` function):

```
  /* Decrement SHLVL by 1 so a new shell started here has the same value,
     preserving the appearance.  After we do that, we need to change the
     exported environment to include the new value.  If we've already forked
     and are in a subshell, we don't want to decrement the shell level,
     since we are `increasing' the level */

  if (cleanenv == 0 && (subshell_environment & SUBSHELL_PAREN) == 0)
    adjust_shell_level (-1);
```

Notice the subtlety involving command capture (i.e., subshells).

krader1961:
Fixes #1091

### DIFF
--- a/pkg/eval/builtin_fn_cmd.go
+++ b/pkg/eval/builtin_fn_cmd.go
@@ -4,6 +4,9 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
+
+	"github.com/elves/elvish/pkg/util"
 )
 
 // Command and process control.
@@ -65,7 +68,8 @@ import (
 // exec $command?
 // ```
 //
-// Replace the Elvish process with an external `$command`, defaulting to `elvish`.
+// Replace the Elvish process with an external `$command`, defaulting to
+// `elvish`. This decrements `$E:SHLVL` before starting the new process.
 
 //elvdoc:fn exit
 //
@@ -118,7 +122,19 @@ func exit(fm *Frame, codes ...int) error {
 	panic("os.Exit returned")
 }
 
+// decSHLVL decrements $E:SHLVL. It's primary purpose is to ensure that
+// $E:SHLVL var is correct when running an `exec` command since that replaces
+// the current shell level with a new command.
+func decSHLVL() {
+	i, err := strconv.Atoi(os.Getenv(util.EnvSHLVL))
+	if err != nil {
+		return
+	}
+	os.Setenv(util.EnvSHLVL, strconv.Itoa(i-1))
+}
+
 func preExit(fm *Frame) {
+	decSHLVL()
 	if fm.DaemonClient != nil {
 		err := fm.DaemonClient.Close()
 		if err != nil {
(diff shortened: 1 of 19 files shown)


krader1961:
@xiaq, This is incomplete in so far as it does not include a unit test. But before I go to the trouble of writing that, especially since there appear to be no tests of `exec`, I want to reach an agreement on whether env vars used or manipulated by Elvish should be referred to by a string or a const symbol. The code you wrote that added support fo `$E:SHLVL` was inconsistent on that point. I opted to replace the literal strings with a symbolic const because such symbols are easier to search for; especially in an IDE.

krader1961:
Also, if we decide to consistently use a symbolic constant, rather than string literals, for uses of "SHLVL" that also applies to other env vars (e.g., "HOME", "PWD") used by Elvish. That transformation should be done as a separate commit; one for each env var being used via a const symbol rather than a literal string.

xiaq:
Changing all the env names to constants sounds good to me.

I don't have a good idea on how to test `exec`, though.

krader1961:
Testing the `exec` command should be possible if the Go unit test framework provides a way to derive the path to the program being tested. But I'm going to punt that problem down the road since I think there are better ways to spend the available time. I'll add a second commit that changes the other env string literals to `eval` package constants.

krader1961:
I have converted the other easily found literal env var name strings to Go constants. Whether the indirection is justified is hard to say. Even though I am old school, and don't use an IDE, I can appreciate how such symbols are preferable to literal strings with such significance.

krader1961:
Regardless of the ultimate decision regarding the reference to environment variables by literal string or Go constant the bug regarding the manipulation of `$E:SHLVL` when executing an `exec` command is still germane.

pkg/eval/builtin_fn_cmd.go
```diff
@@ -118,7 +122,19 @@ func exit(fm *Frame, codes ...int) error {
 	panic("os.Exit returned")
 }
 
+// decSHLVL decrements $E:SHLVL. It's primary purpose is to ensure that
```
xiaq, line 125: Two nitpicks:

* Doc comment for unexported symbols: https://github.com/elves/elvish/blob/master/CONTRIBUTING.md#comment-for-unexported-go-types-and-functions. This is new, BTW, and there is probably some existing code that doesn't follow this.

* It's -> Its

pkg/eval/builtin_fn_cmd.go
```diff
@@ -118,7 +122,19 @@ func exit(fm *Frame, codes ...int) error {
...
+func decSHLVL() {
+	i, err := strconv.Atoi(os.Getenv(util.EnvSHLVL))
+	if err != nil {
+		return
+	}
+	os.Setenv(util.EnvSHLVL, strconv.Itoa(i-1))
+}
+
 func preExit(fm *Frame) {
+	decSHLVL()
```
xiaq, line 137: Since only `exec` needs this, just call it from `execFn` instead.

Issue #1062, opened 2020-07-01
Improve unit test coverage to 85% before 0.15.0 release


xiaq:

xiaq:
Test coverage has not reached 85% yet but 0.15.0 is scheduled to be released soon, so I'm retrospectively amending the scope and keeping this closed.

krader1961:
FWIW, I don't understand why the title of this issue was changed. As I write this the current master branch (at commit 141a6d49558e663f976f6154effb4a6e51c7c93d) has 87.6% coverage as reported by `go tool cover -func=cover`. It is true that https://codecov.io/gh/elves/elvish reports 82.2% coverage, but that is because it incorrectly handles statements such as `import` and `var`; i.e., it considers such statements as not covered even though the compiler ensures they are "executed" at compilation time.

xiaq:
Hmm there is indeed a weird discrepancy, but re

> it incorrectly handles statements such as import and var; i.e., it considers such statements as not covered even though the compiler ensures they are "executed" at compilation time.

I wonder how did you draw that conclusion? If you open any file in codecov the `import` lines are shown without any coloring, indicating that it is ignored for coverage; see this one for example: https://codecov.io/gh/elves/elvish/src/master/pkg/tt/tt.go

If import lines are always considered uncovered, no file with an import line would have 100% coverage in codecov; yet this file shows 100% coverage.

xiaq:
It seems that the difference is in how `if` branches that were never executed is counted. Codecov will not only count the branch body as uncovered, but also the branch condition; for example here: https://codecov.io/gh/elves/elvish/src/master/pkg/cli/addons/histwalk/histwalk.go#L119. Codecov shows the line containing the branch condition as yellow and doesn't consider it to be covered in its coverage calculation.

Indeed, if I manually include the "yellow" lines in the coverage calculation I get a number much closer to the 87.6% that `go tool cover` outputs. Currently for all of `pkg/`, Codecov is showing:

* 11,193 total lines
* 9,201 green lines (fully covered)
* 434 yellow lines (condition lines that never evaluate to `true`)

The coverage number codecov reports by default only includes green lines, which is 9201/11193 = 82.2%. If I include the yellow lines, it's (9201+434)/11193 = 86.1%. It's still not exactly 87.6% but the difference is much smaller now.

Overall I find codecov's more strict calculation to be better. If a branch condition never evaluates to `true`, the condition line should indeed be considered uncovered even if it is executed.

[...]

PR #1093, opened 2020-08-05
Increase umask code coverage from 62% to 100%

krader1961:
Related #1062

### DIFF
--- a/pkg/eval/errs/errs.go
+++ b/pkg/eval/errs/errs.go
@@ -12,6 +12,7 @@ type OutOfRange struct {
 	ValidLow  int
 	ValidHigh int
 	Actual    string
+	ValidFmt  string
 }
 
 func (e OutOfRange) Error() string {
@@ -19,8 +20,12 @@ func (e OutOfRange) Error() string {
 		return fmt.Sprintf(
 			"out of range: %v has no valid value, but is %v", e.What, e.Actual)
 	}
+	validFmt := e.ValidFmt
+	if validFmt == "" {
+		validFmt = "%d"
+	}
 	return fmt.Sprintf(
-		"out of range: %v must be from %v to %v, but is %v",
+		"out of range: %s must be from "+validFmt+" to "+validFmt+", but is %s",
 		e.What, e.ValidLow, e.ValidHigh, e.Actual)
 }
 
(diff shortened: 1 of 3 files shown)


krader1961:
Notice this tiny change surprisingly increases overall coverage by 1.22% to 81.45%.

pkg/eval/unix/umask.go
```diff
@@ -70,22 +74,22 @@ func (UmaskVariable) Set(v interface{}) error {
...
 		}
 		umask = int(intPart)
 	default:
-		return errors.New("umask value must be a string or float64")
+		return ErrUmaskInvalid
 	}
 
 	if umask < 0 || umask > 0o777 {
-		return errors.New("umask value outside the range [0..0o777]")
+		return ErrUmaskInvalid
```
xiaq, line 92: Returning the same error gives less information to the user. Use errs.BadValue and errs.OutOfRange to build the error.

krader1961: Done.

krader1961:
@xiaq, PTAL. Note that I augmented `errs.OutOfRange` to include a `ValidFmt` member that defaults to `%d`. I did so because it's confusing to display the actual value in decimal rather than octal notation. There may be other errors that would benefit from setting that structure member (e.g., to display the value in hex) but that's obviously outside the scope of this change.

pkg/eval/errs/errs.go
```diff
@@ -12,15 +12,20 @@ type OutOfRange struct {
 	ValidLow  int
 	ValidHigh int
 	Actual    string
+	ValidFmt  string
```
xiaq, line 15: This is fine from the perspective of code building `OutOfRange`, but it introduces information that is not useful for consumers of this data type; I'd change `ValidLow` and `ValidHigh` to strings.

xiaq (approved):

pkg/eval/errs/errs.go
krader1961: > I'd change ValidLow and ValidHigh to strings.

Sure, that works too, but my approach avoids having to add two `fmt.Sprintf` calls at every site which constructs this object. And since the vast majority will continue to use the default decimal representation they don't even need to pass a format string to the constructor.

PR #1094, opened 2020-08-07
Eliminate sys.Ioctl

krader1961:
I started by writing tests for sys.Ioctl but after I had 100% coverage
of that function it occurred to me the function could be eliminated.

Related #1062


pkg/sys/isatty_unix.go
```diff
@@ -4,12 +4,11 @@ package sys
...
 import (
 	"os"
-	"unsafe"
 )
 
 // IsATTY returns true if the given file is a terminal.
 func IsATTY(file *os.File) bool {
-	var term Termios
-	err := Ioctl(int(file.Fd()), getAttrIOCTL, uintptr(unsafe.Pointer(&term)))
+	//_, err := unix.IoctlGetTermios(int(file.Fd()), getAttrIOCTL)
```
xiaq, line 11: Do not comment out code, simply remove it.

pkg/sys/tc.go
```diff
@@ -3,20 +3,15 @@
 package sys
 
 import (
-	"unsafe"
-
 	"golang.org/x/sys/unix"
 )
 
+/* This function is currently unused.
```
xiaq, line 9: Remove

pkg/sys/termios.go
```diff
@@ -15,19 +15,16 @@ import (
...
-	var term Termios
-	if err := Ioctl(fd, getAttrIOCTL, uintptr(unsafe.Pointer(&term))); err != nil {
-		return nil, err
-	}
-	return &term, nil
+// TermiosForFd returns a pointer to a Termios structure if the file
+// descriptor is open on a terminal device.
+func TermiosForFd(fd int) (*Termios, error) {
+	term, err := unix.IoctlGetTermios(fd, getAttrIOCTL)
+	return (*Termios)(unsafe.Pointer(term)), err
```
xiaq, line 22: The conversion to unsafe.Pointer is not needed.

pkg/sys/isatty_unix.go
krader1961: Agreed. That was leftover from when I was dithering on whether to rewrite this function in terms of Go's `unix.IoctlGetTermios` or `TermiosForFd`.

pkg/sys/termios.go
krader1961: Yep, again, leftover from previous changes as I explored various refactorings.

pkg/sys/tc.go
krader1961: That works for me. I left it in case you had a good reason for writing it in the first place.

xiaq:
Unfortunately I had to revert this change since it resulted in "failed to put myself in foreground" errors.

PR #1095, opened 2020-08-08
Validate a for loop corner case

krader1961:
This exercises the first failure condition in the `evalForVar` function.
Thus bringing coverage of pkg/eval/value_helper.go to 100%.

Related #1062

### DIFF
--- a/pkg/eval/builtin_special_test.go
+++ b/pkg/eval/builtin_special_test.go
@@ -87,6 +87,9 @@ func TestBuiltinSpecial(t *testing.T) {
 				What:     "iterator",
 				ValidLow: 1, ValidHigh: 1, Actual: 2},
 			"x,y"),
+		// Invalid for loop lvalue. You can't use a var in a namespace other
+		// than the special "local:" namespace as the lvalue in a for loop.
+		That("for no-such-namespace:x [a b] { }").ThrowsMessage("new variables can only be created in local scope"),
 		// Exception when evaluating iterable.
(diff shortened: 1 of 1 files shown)


krader1961:
This is a subtle case in as much as it took me ~4 hours of puzzling over, and instrumenting, the code before I found a case that would exercise the code that handles `for ...` statements wherein the lvalue is invalid. Maybe I'm just getting old since it seems like I should have identified this test case in far less than four hours. Or, perhaps, this suggests that the documentation would benefit from including a definition of the term "lvalue" since it appears in an error exception. Not to mention that the `for` command documentation should probably mention this corner case.

krader1961:
FWIW, the term "lvalue" occurs in these errors:

```
pkg/eval/compile_effect.go:var ErrMoreThanOneRest = errors.New("more than one @ lvalue")
pkg/eval/compile_lvalue.go:     rest, opFunc := cp.lvalueBase(n, "must be an lvalue or a braced list of those")
pkg/eval/compile_lvalue.go:                     cp.errorpf(cn, "must be an lvalue")
pkg/eval/compile_lvalue.go:             rest, opFuncs[i] = cp.lvalueBase(cn.Indexings[0], "must be an lvalue ")
pkg/eval/compile_lvalue.go:                             cp.errorpf(cn, "only the last lvalue may have @")
```

xiaq:
Re lvalue: I raised #1112.

PR #1097, opened 2020-08-09
Add explicit `cd` unit tests

krader1961:
This is to ensure we have good test coverage of the `cd` implementation.

Related #1062

### DIFF
--- a/pkg/eval/builtin_fn_fs_test.go
+++ b/pkg/eval/builtin_fn_fs_test.go
@@ -1,11 +1,20 @@
 package eval
 
 import (
+	"errors"
+	"fmt"
+	"os/user"
 	"path/filepath"
 	"testing"
 
 	"github.com/elves/elvish/pkg/parse"
+	"github.com/elves/elvish/pkg/util"
 )
+
+// For error injection into the util.GetHome function.
+func currentUser() (*user.User, error) {
+	return nil, fmt.Errorf("user unknown")
+}
 
 func TestBuiltinFnFS(t *testing.T) {
 	tmpHome, cleanup := InTempHome()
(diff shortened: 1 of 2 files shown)


xiaq:
The FreeBSD test environment seems to lack a valid HOME. Fix this with `eval.InTempHome`.

krader1961:
> The FreeBSD test environment seems to lack a valid HOME. Fix this with eval.InTempHome.

Ugh! I don't see a failure executing `go test ./...` on FreeBSD (but not OpenBSD). Presumably because I'm doing so via a scripted interactive shell (thanks to tmux) which results in `$E:HOME` being set. On Windows (both my Windows VM and the Appveyor CI) it fails thusly:

```
--- FAIL: TestBuiltinCd (0.01s)
    --- FAIL: TestBuiltinCd/old_=_$pwd;_cd_/;_put_$pwd;_cd_$old;_eq_$old_$pwd (0.00s)
        builtin_fn_fs_test.go:41: got value out [C:\ $true], want [/ $true]
```

Both failures can be addressed by using directories relative to `eval.InTempHome()` so I'll make that change to this PR. However, the Windows failure suggests a lot more thought needs to go into making Elvish work on that platform.

krader1961:
> Both failures can be addressed by using directories relative to....

Note: that is probably only true if we embed the literal value returned by `eval.InTempHome` and otherwise write the comparison of the test result to be independent of the path separator. It's not clear how to do so given the current test framework without having platform specific versions of these tests.

krader1961:
@xiaq, PTAL. This now passes on all CI environments as well as my local machines.

pkg/eval/builtin_fn_fs_test.go
```diff
@@ -23,3 +32,36 @@ func TestBuiltinFnFS(t *testing.T) {
...
+	os.Setenv("HOME", d1Path)
+
+	// We install this mock for all tests, not just the one that needs it,
+	// because it should not be invoked by any of the other tests.
+	util.CurrentUser = currentUser
+	defer func() { util.CurrentUser = user.Current }()
+
+	Test(t,
+		That(`old = $pwd; cd `+d2Path+`; put $pwd; cd $old; eq $old $pwd`).Puts(d2Path, true),
+		That(`old = $pwd; cd /; cd `+d1Path+`; eq $pwd $E:HOME; cd $old`).Puts(true),
```
xiaq, line 54: It's not clear which aspects of the behavior of `cd` these two tests intend to test; it would be useful to document that in comments.

pkg/eval/builtin_fn_fs_test.go
```diff
@@ -23,3 +32,36 @@ func TestBuiltinFnFS(t *testing.T) {
 		That(`-is-dir ~/file`).Puts(false),
 	)
 }
+
+func TestBuiltinCd(t *testing.T) {
```
xiaq, line 36: This test has quite a bit of setup, but not every test case uses those setups. It'd be cleaner to break up this test into multiple independent tests.

pkg/util/gethome.go
```diff
@@ -9,6 +9,9 @@ import (
 	"github.com/elves/elvish/pkg/env"
 )
 
+// For unit test error injection.
```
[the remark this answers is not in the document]
krader1961: Done

pkg/eval/builtin_fn_fs_test.go
```diff
@@ -23,3 +32,36 @@ func TestBuiltinFnFS(t *testing.T) {
...
+	tmpHome, cleanup := InTempHome()
+	defer cleanup()
+
+	mustMkdirAll("d1")
+	mustMkdirAll("d2")
+	d1Path := tmpHome + string(os.PathSeparator) + "d1"
+	d2Path := tmpHome + string(os.PathSeparator) + "d2"
```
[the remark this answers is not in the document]
krader1961: Thanks for the pointer. I presumed there was something akin to Python's `os.path.join` but didn't spot it when I skimmed the Go docs. Note that there are other places in the Elvish code that should be using the `filepath` functions; e.g., *pkg/cli/addons/location/location_test.go* and *pkg/edit/navigation_test.go*.

pkg/eval/builtin_fn_fs_test.go
krader1961: That seems counterproductive. The setup is done once -- not per test. If the tests are split up then any shared setup, such as creating a temp dir, gets done more than once. If there were a large number of individual tests it might be worthwhile from a clarity viewpoint, but there are just a handful of tests.

pkg/eval/builtin_fn_fs_test.go
```diff
@@ -23,3 +32,36 @@ func TestBuiltinFnFS(t *testing.T) {
...
+	os.Chdir(d1Path)
+	os.Setenv("HOME", d1Path)
```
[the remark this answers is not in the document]
krader1961: /face palm :smile: Yeah, I just copy/pasted without looking at what `InTempHome` actually did. Furthermore, the `os.Chdir` is also not needed since having it be the value of `tmpHome` is fine for my purposes.

pkg/eval/builtin_fn_fs_test.go
krader1961: Good point. Done.

krader1961:
@xiaq, PTAL. I implemented all your feedback with the exception of the suggestion to split each individual `cd` test into two or more test functions. I also greatly simplified the test function.

PR #1102, opened 2020-08-10
Use str: equivalents for the deprecated builtins

krader1961:
I noticed that the builtin `joins` function did not have 100% test coverage
while the equivalent `str:join` did have 100% coverage. This replaces
the builtin string commands with those in the `str` module in order to
help ensure anyone using the legacy builtin is actually using the `str`
version. This ensures we don't have two implementations of each command
that behave differently. Bonus: Dedundant code is removed.

This, unfortunately, decreases test coverage of pkg/eval/builtin_fn_str.go
by 1.8% because it removes a bunch of lines that were exercised while
leaving the `eawk` function which has poor coverage. That will be remedied
by improving coverage of the `eawk` function in a separate change.

Related #1062

### DIFF
--- a/pkg/eval/builtin_fn_cmd.go
+++ b/pkg/eval/builtin_fn_cmd.go
@@ -76,7 +76,7 @@ import (
 // Exit the Elvish process with `$status` (defaulting to 0).
 
 func init() {
-	addBuiltinFns(map[string]interface{}{
+	AddBuiltinFns(map[string]interface{}{
 		// Command resolution
 		"external":        external,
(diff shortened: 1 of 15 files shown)


xiaq:
The commands `joins`, `splits` and `replaces` have been deprecated in 0.14.0, so we can just remove them.

The commands `has-prefix` and `has-suffix` has not been deprecated yet, but they also don't contribute anything to the test coverage. I'll mark them as deprecated and we'll remove them after the release of 0.15.0.

With these in mind this PR is not necessary.

PR #1103, opened 2020-08-10
Document builtin `sleep` and add tests

krader1961:
I was reviewing test coverage and noticed that the `esleep`
implementation was undocumented and had no tests. This

a) renames the command to just `sleep`,

b) uses the Go time.ParseDuration function rather than assuming a
simple number of seconds,

c) documents the command,

d) adds tests of the function.

Related #1062

### DIFF
--- a/pkg/eval/builtin_fn_misc.go
+++ b/pkg/eval/builtin_fn_misc.go
@@ -3,11 +3,13 @@ package eval
 // Misc builtin functions.
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"math/rand"
 	"net"
 	"path/filepath"
+	"strconv"
 	"sync"
 	"time"
 	"unicode/utf8"
@@ -17,8 +19,6 @@ import (
 )
 
 // Builtins that have not been put into their own groups go here.
-
-// TODO(xiaq): Document esleep.
 
 func init() {
 	addBuiltinFns(map[string]interface{}{
@@ -34,6 +34,7 @@ func init() {
 
 		// Time
 		"esleep": sleep,
+		"sleep":  sleep,
 		"time":   timeCmd,
 
 		"-ifaddrs": _ifaddrs,
@@ -348,12 +349,85 @@ func readFileUTF8(fname string) (string, error) {
 	return string(bytes), nil
 }
 
-func sleep(fm *Frame, t float64) error {
-	d := time.Duration(float64(time.Second) * t)
+// This allows for unit tests to efficiently test the behavior of the `sleep`
+// command; both by eliminating an actual sleep and verifying the duration was
+// properly parsed.
+var timeAfter func(fm *Frame, d time.Duration) <-chan time.Time = func(
+	fm *Frame, d time.Duration) <-chan time.Time {
+	return time.After(d)
+}
+
+//elvdoc:fn sleep
+//
+// ```elvish
+// sleep $duration
+// ```
+//
+// Pause for the specified duration. The pause granularity depends on the
+// platform and might be as long as 100 milliseconds. The actual pause might
+// be longer than requested for reasons outside the control of Elvish; e.g.,
+// the systems being so busy it can't schedule the elvish process to run.
+//
+// A duration can be a simple [number](../language.html#number) (with optional
+// fractional value) without an explicit unit suffix or a [float64](#float64).
+// In this case the implied time unit is seconds.
+//
+// A duration can also be a string written as a sequence of decimal numbers,
+// each with optional fraction, plus a unit suffix. For example, "300ms",
+// "1.5h" or "1h45m7s". Valid time units are "ns", "us" (or "µs"), "ms", "s",
+// "m", "h".
+//
+// See the [Go documentation](https://golang.org/pkg/time/#ParseDuration) for
+// more information about how durations are parsed. Negative durations raise a
+// "sleep duration must be >= zero" exception since it doesn't make sense to
+// sleep for a negative interval. This is different than the typical BSD or
+// GNU `sleep` command that silently exits with a success status, and no
+// pause, when given a negative duration.
+//
+// This only affects the current Elvish context. It does not affect any other
+// contexts that might be executing in parallel as a consequence of a command
+// such as [`peach`](#peach).
+//
+// Examples:
+//
+// ```elvish-transcript
+// ~> sleep 0.1    # sleeps 0.1 seconds
+// ~> sleep 100ms  # sleeps 0.1 seconds
+// ~> sleep 1.5m   # sleeps 1.5 minutes
+// ~> sleep 1m30s  # sleeps 1.5 minutes
+// ~> sleep -1
+// Exception: sleep duration must be >= zero
+// [tty 8], line 1: sleep -1
+// ```
+
+func sleep(fm *Frame, duration interface{}) error {
+	var d time.Duration
+
+	switch duration := duration.(type) {
+	case float64:
+		d = time.Duration(float64(time.Second) * duration)
+	case string:
+		f, err := strconv.ParseFloat(duration, 64)
+		if err == nil { // it's a simple number assumed to have units == seconds
+			d = time.Duration(float64(time.Second) * f)
+		} else {
+			d, err = time.ParseDuration(duration)
+			if err != nil {
+				return errors.New("invalid sleep duration")
+			}
+		}
+	default:
+		return errors.New("invalid sleep duration")
+	}
+
+	if d < 0 {
+		return fmt.Errorf("sleep duration must be >= zero")
+	}
+
 	select {
 	case <-fm.Interrupts():
 		return ErrInterrupted
-	case <-time.After(d):
+	case <-timeAfter(fm, d):
 		return nil
 	}
 }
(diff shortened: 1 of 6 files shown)


krader1961:
Note that this version of the builtin `sleep` is compatible with GNU and BSD sleep in that it supports simple decimal seconds (the only form BSD sleep supports) and a superset of the time unit suffixes recognized by GNU sleep.

krader1961:
@xiaq, Note the second commit in this PR does verify that `sleep` can be interrupted and code coverage shows the `sleep` function now has 100% coverage. Whether we want to pay the negligible cost of setting up interrupt handling for every test is an open question. The existing `TestWithSetup` function doesn't support this scenario. We could, obviously, change the `TestWithSetup` signature to include a bool that indicates whether signals should be captured; defaulting to false. There aren't that many places that call that function so that would be the simplest way to avoid the overhead of setting up interrupt handling for every test when 99.99% of them don't need it.

krader1961:
Grumble... The AppVeyor failure is because according to the Go docs: "Sending Interrupt on Windows is not implemented.". So the question is: How do we send a signal programmatically on Windows? Off to Google....

krader1961:
Sending an interrupt (aka SIGINT on UNIX) on Windows is far more work than it's worth. See https://github.com/golang/go/blob/master/src/os/signal/signal_windows_test.go for how to do it. I'm just going to refactor the second commit to run the sleep interrupt test only on UNIX. If it works on UNIX we will simply trust the Go runtime to provide the same behavior on Windows.

pkg/eval/builtin_fn_misc.go
```diff
@@ -309,12 +333,39 @@ func readFileUTF8(fname string) (string, error) {
...
+		// time unit so append "s" if the string is a simple decimal integer.
+		// We don't want to convert something like "1m3" to "1m3s" since that
+		// is arguably wrong.
+		for _, r := range duration {
+			if !unicode.IsDigit(r) && r != '.' {
+				return fmt.Errorf("invalid sleep duration")
+			}
+		}
+		duration += "s"
+	}
```
xiaq, line 359: What about this:

* Try to parse `duration` as a number using `strconv.ParseFloat` first, and if successful, use `time.Duration(float64(time.Second) * t)` (`t` being the parsed float) as the duration

* Otherwise, parse it as a duration using `time.ParseDuration`.

pkg/eval/builtin_fn_misc.go
```diff
@@ -309,12 +333,39 @@ func readFileUTF8(fname string) (string, error) {
...
-func sleep(fm *Frame, t float64) error {
-	d := time.Duration(float64(time.Second) * t)
+// This allows for unit tests to efficiently test the behavior of the `sleep`
+// command; both by eliminating an actual sleep and verifying the duration was
+// properly parsed.
+//var TimeAfter func(d time.Duration) <-chan time.Time = func (d time.Duration, fm *Frame) {time.After
+var TimeAfter func(fm *Frame, d time.Duration) <-chan time.Time = func(
+	fm *Frame, d time.Duration) <-chan time.Time {
+	return time.After(d)
+}
```
xiaq, line 343: Do not export this.

pkg/eval/builtin_fn_misc_test.go
```diff
@@ -31,6 +33,26 @@ func TestBuiltinFnMisc(t *testing.T) {
 	)
 }
 
+func TimeAfterMock(fm *Frame, d time.Duration) <-chan time.Time {
```
xiaq, line 36: Do not export this.

pkg/eval/builtin_fn_misc_unix_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+package eval
+
+import (
+	"os"
+	"testing"
+	"time"
+)
+
+func InterruptedTimeAfterMock(fm *Frame, d time.Duration) <-chan time.Time {
```
xiaq, line 11: Do not export this.

pkg/eval/builtin_fn_misc.go
```diff
@@ -160,7 +161,30 @@ import (
...
-// TODO(xiaq): Document esleep.
+//elvdoc:fn sleep
+//
+// ```elvish
+// sleep $duration
+// ```
+//
+// Sleep for the specified duration.  A duration string is a possibly signed
+// sequence of decimal numbers, each with optional fraction and a unit suffix,
+// such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or
```
xiaq, line 172: Since you mentioned a negative duration as an example, also document the behavior when given a negative duration.

pkg/eval/builtin_fn_misc.go
```diff
@@ -160,7 +161,30 @@ import (
...
 // ```
 
-// TODO(xiaq): Document esleep.
+//elvdoc:fn sleep
+//
+// ```elvish
+// sleep $duration
+// ```
+//
+// Sleep for the specified duration.  A duration string is a possibly signed
```
xiaq, line 170: Sleep -> Sleeps. I'd also prefer this to avoid using "sleep" (which is not very rigorous). Go's doc uses "pauses", which is good. Also document that this pauses *at least* the given duration, especially since you mentioned that the OS may have a minimal granularity.

Don't use double spaces after the period.

pkg/eval/builtin_fn_misc.go
```diff
@@ -160,7 +161,30 @@ import (
...
+// sequence of decimal numbers, each with optional fraction and a unit suffix,
+// such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or
+// "µs"), "ms", "s", "m", "h". The default time unit is "s" (seconds). Note
```
xiaq, line 173: > The default time unit is "s" (seconds)

I'd prefer this to be documented more distinctly, since we only support a single number when there is no unit; `1m2` is forbidden intentionally, for example. Something like:

> Alternatively, the duration can be given as a single number, in which case it is interpreted as a number of seconds.

pkg/eval/builtin_fn_misc.go
```diff
@@ -223,9 +247,9 @@ func init() {
 		"-source": source,
 
 		// Time
-		"esleep": sleep,
-		"time":   timeCmd,
-		"-time":  timeCmd,
+		"sleep": sleep,
```
xiaq, line 250: Don't remove `esleep`; make it an alias of `sleep` and deprecate it. We can remove it after 0.15.0.

pkg/eval/builtin_fn_misc.go
krader1961: SGTM. I originally went with this approach because I hoped that people would start using explicit durations rather than simple number with implied units of seconds. My original approach makes it clear we normally expect a time unit suffix. But even if people start writing `sleep 3s` rather than `sleep 3` I agree it isn't worth the somewhat convoluted logic. And, sadly, I expect that most people will continue to use simple numbers without a time unit suffix.

pkg/eval/builtin_fn_misc.go
krader1961: Done

pkg/eval/builtin_fn_misc_test.go
krader1961: Done

pkg/eval/builtin_fn_misc_unix_test.go
krader1961: Done

pkg/eval/builtin_fn_misc.go
krader1961: Done; although, I'm perplexed why it should be deprecated since anyone using it is not just using an experimental feature -- they're using an undocumented, halt-and-catch-fire, feature.

pkg/eval/builtin_fn_misc.go
krader1961: Done

pkg/eval/builtin_fn_misc.go
krader1961: Agreed, but I think a different approach is preferable. Especially since I noticed, as a consequence of thinking about your comment, that the current implementation does not support `float64` values. These two examples should be identical but the second one raises a "wrong type: need string, got number" exception:

```
> sleep 1
> sleep (float64 1)
```

krader1961:
@xiaq, PTAL.

krader1961:
I pushed some minor improvements to the documentation and the wording of the exception for negative durations.

xiaq (approved):
I have a few nitpicks regarding the documentation, but I'll merge this and make those changes myself.

xiaq:
Hmm, NEXT-RELEASE now has a merge conflict; please resolve it.

krader1961:
> Hmm, NEXT-RELEASE now has a merge conflict; please resolve it.

Done. I also reworked the `sleep` documentation which might address some of your implied criticisms with my prior text.

xiaq:
One question: do you have a source for the example of 100ms granularity?

krader1961:
> One question: do you have a source for the example of 100ms granularity?

That should have been 10ms and I haven't seen a system with that sleep granularity since the early 1990's.

PR #1105, opened 2020-08-13
Improve test coverage of pkg/eval/external_cmd.go

krader1961:
Improve coverage of pkg/eval/external_cmd.go from 58.3% to 86.1% as
measured by `go test -covermode` on macOS.

This would have been a smaller change but I felt it was important to
actually validate the exception raised when an external command fails
rather than simply using `.ThrowsAny()`. That necessitated augmenting
the pkg/eval/testutils.go module.

Related #1062

### DIFF
--- a/pkg/eval/external_cmd.go
+++ b/pkg/eval/external_cmd.go
@@ -2,6 +2,7 @@ package eval
 
 import (
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


krader1961:
Closing pending further work to make this pass on Windows. I've fixed the `syscall.WaitStatus` compilation bug but the test of `EachExternal` is failing on Windows. Apparently I need to do more than create empty files with the execute bit set on Windows.

Issue #997, opened 2020-05-01
Test failures 


krader1961:
I sometimes see the following test failure of *pkg/edit/prompt_test.go* on some of my virtual machines:

```
ok      github.com/elves/elvish/pkg/diag        0.028s
--- FAIL: TestPromptStaleThreshold (0.23s)
    prompt_test.go:72: Wanted buffer not shown
    prompt_test.go:72: Want: Width = 50, Dot = (0, 5)
        ┌──────────────────────────────────────────────────┐
        │???> $                                            │
        └──────────────────────────────────────────────────┘
    prompt_test.go:72: Last buffer: Width = 50, Dot = (0, 5)
        ┌──────────────────────────────────────────────────┐
        │???> $                                            │
        └──────────────────────────────────────────────────┘
FAIL
FAIL    github.com/elves/elvish/pkg/edit        4.331s
ok      github.com/elves/elvish/pkg/edit/complete       0.043s
```

Immediately rerunning the unit test succeeds. It looks like the `edit:prompt-stale-threshold = 0.05` in `TestPromptStaleThreshold()` in *pkg/edit/prompt_test.go* causes flakey behavior because it depends on the performance of the platform on which the unit test is running.

xiaq:
I changed the hardcoded value to a value scaled by `ELVISH_TEST_TIME_SCALE` (which was just [documented](https://github.com/elves/elvish/blob/master/CONTRIBUTING.md#elvish_test_time_scale)). You should set this environment variable to a value greater than 1 when running in VMs.

Are you seeing this particular fail disproportionately more often other tests? If so, we can also bump the base value for the times.

krader1961:
> I changed the hardcoded value to a value scaled by ELVISH_TEST_TIME_SCALE...

Fantastic. I'll modify the test runner script I wrote to run the tests in parallel on all my VMs to use that env var and report back soon with the results. When I was running fish unit tests on the same systems I had tweak a similar knob by a small amount to minimize false positive failures.

> Are you seeing this particular fail disproportionately more often other tests?

Yes. While I sometimes see other tests fail intermittently this particular test is flakey more frequently than any other test. I'll take a look at whether tweaking its timeout independent of setting `ELVISH_TEST_TIME_SCALE` reduces the flakiness.

krader1961:
Setting `ELVISH_TEST_TIME_SCALE=2` still results in noticeable flakiness on a couple of my VMs, while five seems to eliminate the problem. Thx.

PR #1107, opened 2020-08-15
Increase options.go coverage to 100%

krader1961:
Related #1062

### DIFF
--- a/pkg/eval/options_test.go
+++ b/pkg/eval/options_test.go
@@ -1,28 +1,47 @@
 package eval
 
-import "testing"
+import (
+	"errors"
+	"testing"
+)
 
 type opts struct {
 	FooBar string
 	POSIX  bool `name:"posix"`
 	Min    int
+	ignore bool // this should be ignored since it isn't exported
 }
 
 var scanOptionsTests = []struct {
 	rawOpts  RawOptions
 	preScan  opts
 	postScan opts
+	err      error
 }{
 	{RawOptions{"foo-bar": "lorem ipsum"},
-		opts{}, opts{FooBar: "lorem ipsum"}},
+		opts{}, opts{FooBar: "lorem ipsum"}, nil},
 	{RawOptions{"posix": true},
-		opts{}, opts{POSIX: true}},
+		opts{}, opts{POSIX: true}, nil},
+	// Since "ignore" is not exported it will result in an error when used.
+	{RawOptions{"ignore": true},
+		opts{}, opts{ignore: false}, errors.New("unknown option ignore")},
 }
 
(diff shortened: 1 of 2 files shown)


krader1961:
The "Test on Linux" environment failed with this error:

```
--- FAIL: TestLocationAddon_Workspace (6.46s)
    listing_nonwindows_test.go:73: Wanted buffer not shown
    listing_nonwindows_test.go:73: Want: Width = 50, Dot = (0, 11)
        ┌──────────────────────────────────────────────────┐
        │~/ws1/bin> $                                      │
        └──────────────────────────────────────────────────┘
    listing_nonwindows_test.go:73: Last buffer: Width = 50, Dot = (1, 11)
        ┌──────────────────────────────────────────────────┐
        │~/ws1/tmp> $                                      │
        │�[1;37;45m LOCATION �[m $                                      │
        │�[7m 10 ws/bin                                        �[m│
        │ 10 /usr/bin$                                     │
        └──────────────────────────────────────────────────┘
FAIL
FAIL	github.com/elves/elvish/pkg/edit	86.464s
```

That's a failure I'm seeing very infrequently on my local systems. See https://github.com/elves/elvish/issues/997 where a very similar failure was discussed. It looks like this is another test that needs to be given more time before reporting a failure.

xiaq:
Hmm, this particular failing test depends on some IO (changing directory), so it's no wonder that among all the tests this one fails. In fact, looking at #997 the one that failed for you also requires some IO (getting current directory).

In the long term these tests can be made more deterministic by using an IO abstraction layer in the spirit of [afero](https://godoc.org/github.com/spf13/afero) (afero doesn't abstract pwd, though). In the meanwhile I'll just bump the test time scale even further to 20.

pkg/eval/options_test.go
```diff
@@ -1,28 +1,45 @@
...
+		t.Errorf("Scan should have reported invalid arg")
+	}
+
 	for _, test := range scanOptionsTests {
 		opts := test.preScan
-		scanOptions(test.rawOpts, &opts)
+		err := scanOptions(test.rawOpts, &opts)
+		if err != nil && test.err == nil {
+			t.Errorf("Scan %v => %v, unexpected error %v", test.rawOpts, opts, err)
+		}
```
xiaq, line 42: * This isn't checking against the actual value of `test.err`. Either check that (you can check the error message), or just specify a boolean in the test struct.

* This doesn't cover the case where `err == nil && test.err != nil`.

krader1961: Yeah, I was taking an unwarranted shortcut based on the knowledge there is currently only one error condition. Fixed so the test is now future proof.

PR #1118, opened 2020-08-19
Support for darwin/386

enty8080:
Support for darwin/386 pre-compiled (iMacs).


krader1961:
Note: As of macOS Catalina 32-bit apps are no longer supported. Apple announced in 2018 that macOS Mojave would be the last version to support 32-bit apps. So the binary that results from this change will have a limited user base.

xiaq:
Go 1.15 has removed support for darwin/386 (https://github.com/golang/go/issues/37610), so I am rejecting this PR.

PR #1119, opened 2020-08-20
Implement search-external &all

### LINKED ISSUE #1111: Add a &all option to search-external
A feature I really miss from other shells is `type -a` to enumerate all the `PATH` entries that contain a particular command in the order they appear in `PATH`. For example:

```
bash-5.0$ type -a git
git is /usr/local/bin/git
git is /usr/bin/git
```

I propose adding a `&all` option to `search-external` that defaults to false.

krader1961:
Fixes #1111

### DIFF
--- a/NEXT-RELEASE.md
+++ b/NEXT-RELEASE.md
@@ -45,6 +45,9 @@ New features in the standard library:
 -   A new `eval` command supports evaluating a dynamic piece of code in a
     restricted namespace.
 
+-   The `search-external` builtin has a new `&all` option to mimic `type -a` in
+    other shells.
+
 # Notable bugfixes
 
 -   Using large lists that contain `$nil` no longer crashes Elvish.
--- a/go.mod
+++ b/go.mod
@@ -5,8 +5,8 @@ require (
 	github.com/kr/pty v1.1.4
 	github.com/mattn/go-isatty v0.0.12
 	github.com/xiaq/persistent v0.0.0-20200820214153-3175cfb92e14
-	go.etcd.io/bbolt v1.3.4
-	golang.org/x/sys v0.0.0-20200323222414-85ca7c5b95cd
+	go.etcd.io/bbolt v1.3.5
+	golang.org/x/sys v0.0.0-20200820212457-1fb795427249
 )
 
 go 1.14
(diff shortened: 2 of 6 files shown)


krader1961:
I force-pushed an update to resolve the conflict in the *NEXT-RELEASE.md* file.

xiaq:
I am not merging this in its current form - the existing implementation just uses `exec.LookPath`, which supports Windows, while the implementation in this PR breaks that.

Since this is going to reimplement `exec.LookPath` anyway, I'd rather fork the entire implementation in [lp_unix.go](https://golang.org/src/os/exec/lp_unix.go) and [lp_windows.go](https://golang.org/src/os/exec/lp_windows.go), and remove all dependency on `exec.LookPath` to ensure consistency. The forked implementation can live in a new package too - maybe `eval/ext` (ext for "external command").

krader1961:
> which supports Windows, while the implementation in this PR breaks that.

Yes, but only when using the `&all` option. Which I thought I had documented. Regardless, I agree it would be better if it "just worked" on both UNIX and Windows whether or not the `&all` option was used. This was a stepping stone since I, personally, don't use MS Windows.

krader1961:
FYI, I'm still interested in getting this feature merged. I just got side-tracked by other issues. I've left this P.R. open to remind me to revisit the change.

I read the respective licenses and don't see any obvious reason we can't fork the golang.org source for the `os.exec` package. However, I'm ambivalent about forking that code. Since the feature implemented by this issue is not performance critical it may make more sense to implement the `&all` behavior in terms of the Go stdlib behavior by trimming the `PATH` env var in a loop when the `&all` option is used.

On the other hand, the Go code we're talking about was last modified in a substantive fashion five years ago and therefore seems to be stable. Therefore, the maintenance cost is likely small if we fork it. On the other hand... Do we care about abandoning support for platforms such as "Plan9" if we fork the unix/windows implementation but none of the others?

krader1961:
Closing because implementing this feature in a fashion that also supports MS Windows will require forking the Go stdlib code. That's because the functionality needed by `search-external &all` is also needed by the `EachExternal()` method that is used by the completion code; whether directly or indirectly thru the `EachVariableInTop()` method. Since I don't care about MS Windows support, and there are many problems with Elvish on Windows, I don't know if, or when, I'll bother to do the necessary work.

PR #1121, opened 2020-08-21
Improve test coverage of pkg/eval/external_cmd.go

krader1961:
Improve coverage of pkg/eval/external_cmd.go from 58.3% to 86.1% as
measured by `make cover` on macOS.

This also fixes a, presumably, copy/paste bug in `func (ExternalCmd)
Kind()` by correcting its returned value: "fn" should be "external".

This would have been a smaller change but I felt it was important to
actually validate the exception raised when an external command fails
rather than simply using `.ThrowsAny()`. That necessitated augmenting
the pkg/eval/testutils.go module.

Related #1062

### DIFF
--- a/pkg/eval/external_cmd.go
+++ b/pkg/eval/external_cmd.go
@@ -69,8 +69,8 @@ func (e ExternalCmd) Call(fm *Frame, argVals []interface{}, opts map[string]inte
 
 	args := make([]string, len(argVals)+1)
 	for i, a := range argVals {
-		// NOTE Maybe we should enfore string arguments instead of coercing all
-		// args into string
+		// TODO: Maybe we should enforce string arguments instead of coercing
+		// all args to strings.
 		args[i+1] = vals.ToString(a)
 	}
 
@@ -83,22 +83,25 @@ func (e ExternalCmd) Call(fm *Frame, argVals []interface{}, opts map[string]inte
 
 	sys := makeSysProcAttr(fm.background)
 	proc, err := os.StartProcess(path, args, &os.ProcAttr{Files: files, Sys: sys})
-
 	if err != nil {
 		return err
 	}
 
 	state, err := proc.Wait()
-
 	if err != nil {
+		// This should be a can't happen situation. Nonetheless, treat it as a
+		// soft error rather than panicing since the Go documentation is not
+		// explicit that this can only happen if we make a mistake. Such as
+		// calling `Wait` twice on a particular process object.
 		return err
 	}
 	return NewExternalCmdExit(e.Name, state.Sys().(syscall.WaitStatus), proc.Pid)
 }
 
-// EachExternal calls f for each name that can resolve to an external
-// command.
-// TODO(xiaq): Windows support
+// EachExternal calls f for each name that can resolve to an external command.
+//
+// TODO(xiaq): Windows support. See https://golang.org/pkg/os/#Chmod for why
+// this doesn't work on Windows as currently written.
 func EachExternal(f func(string)) {
 	for _, dir := range searchPaths() {
 		// TODO(xiaq): Ignore error.
--- a/pkg/eval/external_cmd_test.go
+++ b/pkg/eval/external_cmd_test.go
@@ -1,0 +1,31 @@
+package eval
+
+import (
+	"os"
+	"testing"
+)
+
+func TestBuiltinFnExternal(t *testing.T) {
+	tmpHome, cleanup := InTempHome()
+	defer cleanup()
+
+	os.Setenv("PATH", tmpHome+":"+os.Getenv("PATH"))
+	Test(t,
+		That(`e = (external true); kind-of $e`).Puts("fn"),
+		That(`e = (external true); put (repr $e)`).Puts("<external true>"),
+		That(`e = (external false); m = [&$e=true]; put (repr $m)`).Puts("[&<external false>=true]"),
+		// This group tests the `ExternalCmd.Call` method.
+		That(`e = (external true); $e`).DoesNothing(),
+		That(`e = (external true); $e &option`).Throws(ErrExternalCmdOpts, "$e &option"),
+		That(`e = (external false); $e`).ThrowsCmdExit(
+			ExternalCmdExit{CmdName: "false", WaitStatus: exitWaitStatus(1)}),
+
+		// TODO: Modify the ExternalCmd.Call method to wrap the Go error in a
+		// predictable Elvish error so we don't have to resort to using
+		// ThrowsAny in the following tests.
+		//
+		// The command shouldn't be found when run so we should get an
+		// exception along the lines of "executable file not found in $PATH".
+		That(`e = (external true); E:PATH=/ $e`).ThrowsAny(),
+	)
+}
--- a/pkg/eval/external_cmd_unix_test.go
+++ b/pkg/eval/external_cmd_unix_test.go
@@ -1,0 +1,37 @@
+// +build !windows,!plan9,!js
+
+package eval
+
+import (
+	"os"
+	"reflect"
+	"sort"
+	"testing"
+)
+
+// TODO: When EachExternal is modified to work on Windows either fold this
+// test into external_cmd_test.go or create an external_cmd_windows_test.go
+// that performs an equivalent test on Windows.
+func TestEachExternal(t *testing.T) {
+	tmpHome, cleanup := InTempHome()
+	defer cleanup()
+
+	mustMkdirAll("dir")
+	mustCreateEmpty("cmdx")
+	mustWriteFile("cmd1", []byte("#!/bin/sh"), 0755)
+	mustWriteFile("cmd2", []byte("#!/bin/sh"), 0755)
+	mustWriteFile("cmd3", []byte(""), 0755)
+	mustCreateEmpty("file")
+
+	os.Setenv("PATH", "/argle:"+tmpHome+":/bargle")
+	wantCmds := []string{"cmd1", "cmd2", "cmd3"}
+	gotCmds := []string{}
+	EachExternal(func(filename string) {
+		gotCmds = append(gotCmds, filename)
+	})
+
+	sort.Strings(gotCmds)
+	if !reflect.DeepEqual(wantCmds, gotCmds) {
+		t.Errorf("EachExternal want %q got %q", wantCmds, gotCmds)
+	}
+}
--- a/pkg/eval/testutils.go
+++ b/pkg/eval/testutils.go
@@ -120,6 +120,24 @@ func (e errWithMessage) matchError(e2 error) bool {
 	return e2 != nil && e.msg == e2.Error()
 }
 
+// An errorMatcher for an ExternalCmdExit error that ignores the `Pid` member.
+// We only match the command name and exit status because at run time we
+// cannot know the correct value for `Pid`.
+type errCmdExit struct{ v ExternalCmdExit }
+
+func (e errCmdExit) Error() string {
+	return e.v.Error()
+}
+
+func (e errCmdExit) matchError(gotErr error) bool {
+	if gotErr == nil {
+		return false
+	}
+
+	ge := gotErr.(*Exception).Reason.(ExternalCmdExit)
+	return e.v.CmdName == ge.CmdName && e.v.WaitStatus == ge.WaitStatus
+}
+
 // The following functions and methods are used to build Test structs. They are
 // supposed to read like English, so a test that "put x" should put "x" reads:
 //
@@ -180,6 +198,13 @@ func (t TestCase) ThrowsCause(err error) TestCase {
 // Whenever possible, use ThrowsCause instead of this method.
 func (t TestCase) ThrowsMessage(msg string) TestCase {
 	return t.throws(errWithMessage{msg})
+}
+
+// ThrowsCmdExit returns an altered TestCase that requires the source code to
+// throw an an ExternalCmdExit error that matches the given error, ignoring
+// the PID.
+func (t TestCase) ThrowsCmdExit(err ExternalCmdExit) TestCase {
+	return t.throws(errCmdExit{err})
 }
 
 // ThrowsAny returns an altered TestCase that requires the source code to throw
--- a/pkg/eval/testutils_unix.go
+++ b/pkg/eval/testutils_unix.go
@@ -1,0 +1,14 @@
+// +build !windows,!plan9,!js
+
+package eval
+
+import (
+	"syscall"
+)
+
+func exitWaitStatus(exit uint32) syscall.WaitStatus {
+	// The exit<<8 is gross but I can't find any exported symbols that would
+	// allow us to construct WaitStatus. So assume legacy UNIX encoding
+	// for a process that exits normally; i.e., not due to a signal.
+	return syscall.WaitStatus(exit << 8)
+}
--- a/pkg/eval/builtin_fn_misc_test.go
+++ b/pkg/eval/builtin_fn_misc_test.go
@@ -99,5 +99,6 @@ func TestResolve(t *testing.T) {
 		That("fn f { }; resolve f").Puts("$f~"),
 		That("use mod; resolve mod:func").Puts("$mod:func~"),
 		That("resolve cat").Puts("(external cat)"),
+		That(`resolve external`).Puts("$external~"),
 	)
 }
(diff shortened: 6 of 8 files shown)


pkg/eval/testutils.go
```diff
@@ -120,6 +120,31 @@ func (e errWithMessage) matchError(e2 error) bool {
 	return e2 != nil && e.msg == e2.Error()
 }
 
+// An errorMatcher for an ExternalCmdExit error that ignores the Pid member.
+// That is, we only match the command name and raw exit status.
+type errCmdExit struct{ v error }
+
+func (e errCmdExit) Error() string { return fmt.Sprintf("%#v", e.v) }
```
xiaq, line 127: This should return a description of what is expected. Since this only matches CmdName and WaitStatus, it should return something like `ExternalCmdExit with CmdName = ... and WaitStatus = ...`

pkg/eval/testutils.go
```diff
@@ -120,6 +120,31 @@ func (e errWithMessage) matchError(e2 error) bool {
...
+		return false
+	}
+	we, ok := want_err.v.(ExternalCmdExit)
+	if !ok { // this should never happen
+		panic("unexpected want_err type passed to errCmdExit.matchError")
+	}
+
+	ge, ok0 := got_err.(*Exception).Reason.(ExternalCmdExit)
+	if !ok0 {
+		fmt.Println("WTF e0 !ok")
```
xiaq, line 140: Remove

pkg/eval/external_cmd.go
```diff
@@ -29,7 +30,7 @@ type ExternalCmd struct {
 }
 
 func (ExternalCmd) Kind() string {
-	return "fn"
+	return "external"
```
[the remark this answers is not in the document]
krader1961: I can't grok what you're saying since "loose interface" has no meaning. Furthermore, it seems to me that if I do `cmd = (external true); kind-of $cmd` it should tell me `$cmd` is an external command. Nonetheless, I'll revert this change since it isn't really the point of this PR.

pkg/eval/external_cmd.go
```diff
@@ -83,22 +84,23 @@ func (e ExternalCmd) Call(fm *Frame, argVals []interface{}, opts map[string]inte
...
 		return err
 	}
 
 	state, err := proc.Wait()
-
 	if err != nil {
-		return err
+		// This is a "can't happen" situation unless we do something invalid
+		// such as call proc.Wait() twice.
+		panic(fmt.Sprintf("proc.Wait returned unexpected error: %v", err))
```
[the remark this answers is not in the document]
krader1961: Okay, I'll change this to return the error but retain an (expanded) comment about why we return the error rather than panicing.

pkg/eval/testutils.go
```diff
@@ -120,6 +120,31 @@ func (e errWithMessage) matchError(e2 error) bool {
 	return e2 != nil && e.msg == e2.Error()
 }
 
+// An errorMatcher for an ExternalCmdExit error that ignores the Pid member.
+// That is, we only match the command name and raw exit status.
+type errCmdExit struct{ v error }
+
+func (e errCmdExit) Error() string { return fmt.Sprintf("%#v", e.v) }
+
+func (want_err errCmdExit) matchError(got_err error) bool {
```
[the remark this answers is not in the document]
krader1961: Old habits die hard. Fixed. However, I'm ambivalent about using `e` rather than `wantErr` here. I only used `e` as the receiver var in the `Error()` definition to match the other code in this module. In this method it seems to me using `wantErr` is clearer. Nonetheless, I'll use `e` rather than `wantErr`.

pkg/eval/testutils.go
krader1961: Yep, leftover debug print. Removed.

pkg/eval/testutils.go
```diff
@@ -192,6 +217,15 @@ func (t TestCase) ThrowsMessage(msg string) TestCase {
 	return t.throws(errWithMessage{msg})
 }
 
+// ThrowsCmdExit returns an altered TestCase that requires the source code to
+// throw an ExternalCmdExit error that ignores the pid.
+func (t TestCase) ThrowsCmdExit(err error) TestCase {
```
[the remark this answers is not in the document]
krader1961: The problem is that the pre-existing `NewExternalCmdExit` function returns an `error`. I'll investigate whether it is appropriate, and makes more sense, for that function to return an `ExternalCmdExit`.

pkg/eval/testutils.go
krader1961: SGTM as far as being more explicit about the important attributes of the value. But note that `errCmdExit` is a generic struct that wraps an `error`, not an `ExternalCmdExit`. Which necessitates writing `e.v.(ExternalCmdExit).WaitStatus` rather than `e.WaitStatus`. That is obviously hideous. This is closely related to your comment about the `ThrowsCmdExit` method.

pkg/eval/testutils.go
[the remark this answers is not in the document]
krader1961: I decided to modify the unit test to directly instantiate a `ExternalCmdExit` struct rather than call `NewExternalCmdExit` (which returns an `error` object). Thus allowing me to make this method require a `ExternalCmdExit` rather than a generic `error` object as you requested. Still, it's not clear this is an improvement due to the disconnect with the value returned by `NewExternalCmdExit`.

pkg/eval/testutils.go
krader1961: I opted to change the `errCmdExit` type to wrap `ExternalCmdExit` rather than `error`. That, and a related change, eliminated most of the ugly reflection code.

krader1961:
@xiaq, PTAL. I verified some of the testutils error paths work and produce useful output by deliberately injecting errors into the unit test. Such as running `fail me` rather than an external command that would result in an `ExternalCmdExit` error.

pkg/eval/testutils.go
```diff
@@ -120,6 +120,28 @@ func (e errWithMessage) matchError(e2 error) bool {
...
+
+func (e errCmdExit) matchError(gotErr error) bool {
+	if gotErr == nil {
+		return false
+	}
+
+	ge, ok := gotErr.(*Exception).Reason.(ExternalCmdExit)
+	if !ok {
+		panic(fmt.Sprintf("errCmdExit.matchError expected eval.ExternalCmdExit, got %T", gotErr))
+	}
```
xiaq, line 140: Since this error is handled by a panic anyway, use the single-lhs form of type assertion, which panics when the type doesn't match.

pkg/eval/testutils.go
```diff
@@ -182,6 +204,12 @@ func (t TestCase) ThrowsMessage(msg string) TestCase {
 	return t.throws(errWithMessage{msg})
 }
 
+// ThrowsCmdExit returns an altered TestCase that requires the source code to
+// throw an ExternalCmdExit error that ignores the PID.
```
xiaq, line 208: > an ExternalCmdExit error that ignores the PID

->

> an ExternalCmdExit error that matches the given error, ignoring the PID

pkg/eval/testutils_unix.go
```diff
@@ -0,0 +1,14 @@
+// +build !windows,!plan9,!js
+
+package eval
+
+import (
+	"syscall"
+)
+
+func ExitWaitStatus(exit uint32) syscall.WaitStatus {
```
xiaq, line 9: No need to export this

pkg/eval/testutils_unix.go
```diff
@@ -0,0 +1,14 @@
...
+	// The exit<<8 is gross but I can't find any exported symbols that would
+	// allow us to construct WaitStatus. So assume legacy UNIX encoding
```
xiaq, line 11: > legacy UNIX encoding

AFAIK there isn't a "modern UNIX" encoding.

pkg/eval/external_cmd_unix_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	mustMkdirAll("dir")
+	mustCreateEmpty("cmdx")
+	mustWriteFile("cmd1", []byte("#!/bin/sh"), 0755)
+	mustWriteFile("cmd2", []byte("#!/bin/sh"), 0755)
+	mustWriteFile("cmd3", []byte(""), 0755)
+	mustCreateEmpty("file")
+
+	os.Setenv("PATH", "/argle:"+tmpHome+":/bargle")
+	want_cmds := []string{"cmd1", "cmd2", "cmd3"}
```
xiaq, line 27: Don't use underscores in variable names

pkg/eval/external_cmd.go
```diff
@@ -28,6 +28,10 @@ type ExternalCmd struct {
 	Name string
 }
 
+// TODO: Decide what `kind-of` should report for an external command object
+// and document the rationale for the choice. It's not obvious why this
+// returns "fn" rather than "external". If this is changed
+// TestBuiltinFnExternal will need to be updated.
```
xiaq, line 34: The entire "kind" mechanism is kind of half-baked, so calling it out here is a bit weird. Better remove this and document the situation in vals.Kind instead.

pkg/eval/external_cmd_test.go
```diff
@@ -0,0 +1,41 @@
...
+	mustMkdirAll("dir")
+	mustCreateEmpty("cmdx")
+	mustWriteFile("cmd1", []byte("#!/bin/sh"), 0755)
+	mustWriteFile("cmd2", []byte("#!/bin/sh"), 0755)
+	mustWriteFile("cmd3", []byte(""), 0755)
+	mustCreateEmpty("file")
+
+	os.Setenv("PATH", tmpHome+":"+os.Getenv("PATH"))
+	Test(t,
+		That(`resolve external`).Puts("$external~"),
```
xiaq, line 21: This test case doesn't belong here.

pkg/eval/external_cmd_test.go
```diff
@@ -0,0 +1,41 @@
...
+		// TODO: Modify the ExternalCmd.Call method to wrap the Go error in a
+		// predictable Elvish error so we don't have to resort to using
+		// ThrowsAny in the following tests.
+		//
+		// The command shouldn't be found when run so we should get an
+		// exception along the lines of "executable file not found in $PATH".
+		That(`e = (external true); E:PATH=/; $e`).ThrowsAny(),
+		// The command will be found but cause an exception along the lines of
+		// "exec format error" because it is marked executable but is empty.
+		That(`e = (external cmd3); E:PATH=/; $e`).ThrowsAny(),
```
xiaq, line 39: I don't think this test case tests what is claimed; it throws an exception for the same reason the previous test case does.

pkg/eval/external_cmd_test.go
```diff
@@ -0,0 +1,41 @@
...
+	defer cleanup()
+
+	mustMkdirAll("dir")
+	mustCreateEmpty("cmdx")
+	mustWriteFile("cmd1", []byte("#!/bin/sh"), 0755)
+	mustWriteFile("cmd2", []byte("#!/bin/sh"), 0755)
+	mustWriteFile("cmd3", []byte(""), 0755)
+	mustCreateEmpty("file")
+
+	os.Setenv("PATH", tmpHome+":"+os.Getenv("PATH"))
```
xiaq, line 19: None of these setups seem to be actually used.

pkg/eval/external_cmd_test.go
```diff
@@ -0,0 +1,41 @@
...
+		That(`e = (external true); $e &option`).Throws(ErrExternalCmdOpts, "$e &option"),
+		That(`e = (external false); $e`).ThrowsCmdExit(
+			ExternalCmdExit{CmdName: "false", WaitStatus: ExitWaitStatus(1)}),
+		// TODO: Modify the ExternalCmd.Call method to wrap the Go error in a
+		// predictable Elvish error so we don't have to resort to using
+		// ThrowsAny in the following tests.
+		//
+		// The command shouldn't be found when run so we should get an
+		// exception along the lines of "executable file not found in $PATH".
+		That(`e = (external true); E:PATH=/; $e`).ThrowsAny(),
```
xiaq, line 36: Don't mutate E:PATH; use a temporary assignment.

pkg/eval/testutils.go
krader1961: Okay. I did this in the hope the more explicit panic message might help but since it should never be seen I can see your point that there is little benefit to those lines.

pkg/eval/testutils.go
krader1961: done

pkg/eval/testutils_unix.go
krader1961: done

pkg/eval/testutils_unix.go
krader1961: True but in this instance I'm using this meaning: a thing handed down by a predecessor. The intent is to make it clear we're relying on behavior established long ago but is not guaranteed to be true if we run on a UNIX-like OS.

pkg/eval/external_cmd_unix_test.go
krader1961: done

pkg/eval/external_cmd_test.go
krader1961: Agreed. Moved to *pkg/eval/builtin_fn_misc_test.go*.

pkg/eval/external_cmd.go
krader1961: Done. Comment moved (albeit rewritten) to the doc string for `func Kind`.

pkg/eval/external_cmd_test.go
krader1961: Done. I wrote it this way because I hadn't decided to un-deprecate, and test, the `$pwd` behavior when I wrote this. Since you seem to be willing to retain `$pwd`, at least for the time being, I've modified this test to use it.

pkg/eval/external_cmd_test.go
krader1961: It throws a subtly different error. Whether the two cases should be distinguishable is an open question. See the preceding "TODO". You can argue that they should throw the same exception but they don't which is why I have these two discrete tests to guard against future regressions from the current behavior. Granted, it currently doesn't do so but that's because Elvish doesn't currently provide a predictable exception for each case. Which is why these two tests use `ThrowsAny`.

pkg/eval/external_cmd_test.go
krader1961: Hmmm, I must have inadvertently deleted the tests that use those files. The "cmdx" file, for example, is meant to be used to verify that it is not a usable external command due to the lack of execute permission. But looking at it now it would be better to use `mustWriteFile` to make that explicit -- in addition to adding the test that uses that file.

krader1961: Aha! Those are all from when these tests were combined with the tests in *pkg/eval/external_cmd_unix_test.go*. I had to split the two sets of tests because of MS Windows weirdness. In the process I failed to remove these statements from this unit test. So, basically a copy/paste error.

pkg/eval/external_cmd_test.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"os"
+	"testing"
+)
+
+func TestBuiltinFnExternal(t *testing.T) {
+	tmpHome, cleanup := InTempHome()
+	defer cleanup()
+
+	os.Setenv("PATH", tmpHome+":"+os.Getenv("PATH"))
```
xiaq, line 12: Use util.WithTempEnv so that the original value can be restored.

pkg/eval/external_cmd_unix_test.go
```diff
@@ -0,0 +1,37 @@
...
+	defer cleanup()
+
+	mustMkdirAll("dir")
+	mustCreateEmpty("cmdx")
+	mustWriteFile("cmd1", []byte("#!/bin/sh"), 0755)
+	mustWriteFile("cmd2", []byte("#!/bin/sh"), 0755)
+	mustWriteFile("cmd3", []byte(""), 0755)
+	mustCreateEmpty("file")
+
+	os.Setenv("PATH", "/argle:"+tmpHome+":/bargle")
```
xiaq, line 26: Use util.WithTempEnv. Stick to more boring placeholder names (foo, bar, etc.)

xiaq (approved):
Just some minor feedback so I'll merge and do a small fixup myself.

PR #1125, opened 2020-08-25
Eliminate sys.Ioctl

krader1961:
I started by writing tests for sys.Ioctl but after I had 100% coverage
of that function it occurred to me the function could be eliminated.

Related #1062


krader1961:
Note that this is essentially identical to my earlier commit bea90271 that was broken, and had to be reverted, because it did not use `unix.IoctlSetPointerInt` in the `Tcgetpgrp` function. This version also comments out code related to flushing tty input since it can never be executed.

krader1961:
Note that we can now use `unix.IoctlSetPointerInt` on all UNIX platforms because https://github.com/golang/go/issues/40986, which I opened, was resolved.

pkg/cli/term/setup_unix.go
```diff
@@ -41,12 +39,12 @@ func setup(in, out *os.File) (func() error, error) {
...
 
 	var errFlushInput error
-	if flushInputDuringSetup {
+	/*
 		err = sys.FlushInput(fd)
 		if err != nil {
 			errFlushInput = fmt.Errorf("can't flush input: %s", err)
 		}
-	}
+	*/
```
xiaq, line 47: Don't comment out code, just remove it.

pkg/sys/termios.go
```diff
@@ -66,7 +63,10 @@ func (term *Termios) SetICRNL(v bool) {
...
 }
 
+/* See the commented out block of code in the term.setup() function.
+
 // FlushInput discards data written to a file descriptor but not read.
 func FlushInput(fd int) error {
-	return Ioctl(fd, flushIOCTL, uintptr(unix.TCIFLUSH))
+	return unix.IoctlSetPointerInt(fd, flushIOCTL, unix.TCIFLUSH)
 }
+*/
```
xiaq, line 72: Don't comment out code, just remove it.

pkg/cli/term/setup_unix.go
krader1961: Fine by me. I only made the implicit commenting out explicit because it wasn't clear why that code was never executed. As in, you didn't include a comment as to why the code was hidden behind `if false`. :smile: So I opted to keep it as comments just in case you intended to enable it at some point in the future.

PR #1129, opened 2020-08-28
go vet cleanup

krader1961:
Most of the code uses keyed fields in composite literals; i.e., struct
literals. However, running `go vet ./...` reports a few places that use
anonymous fields. This modifies those composite literals to use keyed
fields. This does make the code a bit more verbose without reducing
the likelihood of a bug. But it does make the code more consistent, use
best practices, and make it easier to notice if a potential problem is
introduced when running `go vet ./...` since that command now produces
no diagnostic output.

I considered adding a `make vet` target that explicitly ran

    go vet -composites=false ./...

I decided not to do that since consistently using keyed composite literals
is preferable to having a mix of keyed and unkeyed composite literals.

This also removes the unused `ExampleLoop` function which causes this
`go vet` warning:

    pkg/cli/loop_test.go:130:1: ExampleLoop refers to unknown identifier: Loop

### DIFF
--- a/pkg/cli/loop_test.go
+++ b/pkg/cli/loop_test.go
@@ -127,7 +127,9 @@ func quitOn(lp *loop, retTrigger event, ret string, err error) handleCb {
 	}
 }
 
-func ExampleLoop() {
+// This will be run by `go test` and compare the output to that in the comment
+// at the end of the function. See https://golang.org/pkg/testing/#pkg-examples.
+func Example_loop() {
 	buffer := ""
 	firstDrawerCall := true
 	drawer := func(flag redrawFlag) {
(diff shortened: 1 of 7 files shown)


pkg/cli/loop_test.go
```diff
@@ -126,42 +126,3 @@ func quitOn(lp *loop, retTrigger event, ret string, err error) handleCb {
 		}
 	}
 }
-
```
xiaq, line 129: Don't remove this; example functions are run as test cases by `go test`.

krader1961: Ah! Another day where I learned something new. Reinstated but with the function name mangled so `go vet` doesn't complain that it can't find a `Loop` symbol and a comment to aid future readers who might be unaware of this `go test` feature.

PR #1131, opened 2020-08-29
Var pwd tests

krader1961:


krader1961:
I wouldn't normally introduce a mock for an extremely rare error case. But in this case a) the situation isn't as rare as most people think, and b) it's a trivial mock injection which helps ensure that no future change inadvertently changes the behavior of that error condition.

pkg/eval/pwd.go
```diff
@@ -14,11 +14,13 @@ type PwdVariable struct {
 
 var _ vars.Var = PwdVariable{}
 
-// Get returns the current working directory. It returns /unknown/pwd when it
-// cannot be determined.
+// Getwd allows for unit test error injection.
+var Getwd func() (string, error) = os.Getwd
```
xiaq, line 18: Don't export this.

pkg/eval/pwd_test.go
```diff
@@ -0,0 +1,40 @@
...
+	mustMkdirAll("dir1")
+	mustMkdirAll("dir2")
+	dir1 := filepath.Join(tmpHome, "dir1")
+	dir2 := filepath.Join(tmpHome, "dir2")
+
+	Test(t,
+		That(`pwd=dir1 put $pwd; put $pwd`).Puts(dir1, tmpHome),
+		That(`cd $E:HOME/dir2; pwd=~ put $pwd; put $pwd`).Puts(tmpHome, dir2),
+		That(`cd $E:HOME/dir2; pwd=~/dir1 put $pwd; put $pwd`).Puts(dir1, dir2),
+		That(`cd $E:HOME/dir1; pwd=../dir2 put $pwd; put $pwd`).Puts(dir2, dir1),
```
xiaq, line 22: I don't see a good reason to alternate between E:HOME and ~ here; stick to the more idiomatic ~ if possible.

pkg/eval/pwd_test.go
```diff
@@ -0,0 +1,40 @@
...
+		That(`pwd=(float64 1) put $pwd`).Throws(ErrPathMustBeString, "pwd=(float64 1)"),
+	)
+}
+
+// Verify the behavior when the CWD cannot be determined.
+func TestBuiltinPwdNoCwd(t *testing.T) {
```
xiaq, line 28: `TestBuiltinPwd_GetwdError` would be a better name.

(Names of test functions are one of the few places in idiomatic Go code that allow underscores.)

pkg/eval/pwd_test.go
```diff
@@ -0,0 +1,40 @@
...
+	defer func() { Getwd = origGetwd }()
+
+	Test(t,
+		That(`put $pwd`).Puts("/unknown/pwd"),
+	)
+}
+
+func mockGetwd() (string, error) {
+	return "", errors.New("cwd unknown")
+}
```
xiaq, line 40: Either rename this to a more descriptive `mockGetwdWithError`, or inline it in the test function above.

pkg/eval/pwd.go
krader1961: done

pkg/eval/pwd_test.go
krader1961: Fair enough. Looking at it now I don't know why I did this other than due to the use of `InTempHome()`.

pkg/eval/pwd_test.go
krader1961: done

pkg/eval/pwd_test.go
krader1961: done

PR #1143, opened 2020-09-07
Sort and dedup glob expansions

### LINKED ISSUE #1019: Wildcard expansions aren't sorted -- should they be sorted?
While working on issue #1018 I added a few relevant unit tests. Which caused me to notice the tests are fragile because they assume the wildcard expansion will be in a specific order. But wildcard expansions are not sorted so the test is really depending on machine specific behavior involving the order in which objects are created and directories are traversed. It seems to me the results of a wildcard expansion should always be sorted. The cost of sorting the results will normally be negligible and will a) make the unit tests less fragile, and, more importantly, b) result in more user friendly behavior.

krader1961:
Fixes #1019

### DIFF
--- a/NEXT-RELEASE.md
+++ b/NEXT-RELEASE.md
@@ -42,6 +42,9 @@ New features in the language:
 -   Rest variables and rest arguments are no longer restricted to the last
     variable.
 
+-   Glob expansions are now always sorted and duplicates removed. See
+    https://github.com/elves/elvish/issues/1019
+
 New features in the standard library:
 
 -   A new `eval` command supports evaluating a dynamic piece of code in a
--- a/pkg/eval/compile_value.go
+++ b/pkg/eval/compile_value.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sort"
 	"strings"
 	"sync"
 
@@ -86,6 +87,7 @@ func (op compoundOp) exec(fm *Frame) ([]interface{}, error) {
 			return nil, fm.errorp(op, err)
 		}
 	}
+
 	if op.tilde {
 		newvs := make([]interface{}, len(vs))
 		for i, v := range vs {
@@ -97,6 +99,7 @@ func (op compoundOp) exec(fm *Frame) ([]interface{}, error) {
 		}
 		vs = newvs
 	}
+
 	hasGlob := false
 	for _, v := range vs {
 		if _, ok := v.(GlobPattern); ok {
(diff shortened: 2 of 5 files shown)


xiaq:
I think the deduplication is there to "fix" the behavior of multiple indices?

That's the wrong abstraction level to fix the behavior. It also makes the implementation not as efficient as it should be - a small concern in itself but a manifestation that it's just patching things up instead of fixing the issue at its root.

The root of the issue is that the behavior of multiple indices is now controlled centrally in [(*indexingOp).exec](https://github.com/elves/elvish/blob/5ec11cfe2bf91b8b385cd995527e8935ff7180c5/pkg/eval/compile_value.go#L238) - it takes each index, applies it to the indexee with `vals.Index`, and join all the results together.

To fix the issue at the root, let `vals.Index` take multiple indices, and vary the behavior of multiple indices depending on the indexee type.

krader1961:
> I think the deduplication is there to "fix" the behavior of multiple indices?

Yes.

> That's the wrong abstraction level to fix the behavior.

Agreed, but after pondering the existing code for a few hours I threw up my hands in defeat with respect to finding the optimal solution. So I opted for "fixing" the problem at the top-level. Which, while suboptimal from a performance perspective, ensures that the result does not include duplicates even if future changes to the glob mechanism introduces other ways to create duplicate pathnames.

> To fix the issue at the root, let vals.Index take multiple indices, and vary the behavior of multiple indices depending on the indexee type.

It's not at all obvious that the behavior of https://github.com/elves/elvish/blob/5ec11cfe2bf91b8b385cd995527e8935ff7180c5/pkg/eval/compile_value.go#L233 should depend on the "indexee type." Why should that eliminate duplicates when used in a glob expansion but not in non-glob context?

krader1961:
@xiaq, See also https://github.com/elves/elvish/issues/1019#issuecomment-687696406

krader1961:
Also, note that I replaced `ioutil.ReadDir` (which sorts the values it returns) with `os.ReadDir` (which does not sort the values). So my change does not sort the values twice and should be, within a couple percent, as fast as the current implementation.

krader1961:
@xiaq, See also my [comment on issue 1019](https://github.com/elves/elvish/issues/1019#issuecomment-687696406). The more I've thought about this the more convinced I am that the approach used by this PR is the correct one. Your proposal vis-a-vis changing the `vals.Index` function only makes sense if it performs deduplication regardless of the context.

krader1961:
@xiaq, I grok your last comment; however, I still don't see why it doesn't make more sense to defer the sorting, and duplicate elimination, to the glob top-level context. When filtering glob expansions there isn't any reason to first sort the glob expansion. Also, your proposal greatly complicates the code -- far more so than my change. Sometimes that makes sense as a stepping stone to related changes. But in this instance the more complex solution doesn't appear to be justified.

xiaq:
The implementation in this PR is a workaround for the real source of the problem at the wrong level, it is not a valid stepping stone.

krader1961:
@xiaq, You wrote

> The root of the issue is that the behavior of multiple indices $v[$i $j] is always equivalent to $v[$i] $v[$j]. 

That is technically true but confusing in the context of globs. In an empty directory run `touch 1 a b e2 bo`. Then compare the output of `put *?[set:aeoiu digit]*` to `put *?[set:aeoiu]* *?[digit]*`. The issue here is that the expression being indexed is `?` without regard to the surrounding `*` globs and the index expansion is forming a cartesian product with the surrounding globs. I haven't yet looked at the code so it will be interesting to see if your preferred approach to fixing this addresses the problem as efficiently, if not more, than my current approach.

xiaq:
> That is technically true but confusing in the context of globs.

That is precisely my point. The behavior of expanding `$v[$i $j]` to `$v[$i] $v[$j]` is not suitable for globs, that is the root of the problem. So fixing the problem at the root means changing the behavior of the indexing operation to *not* expand to `$v[$i] $v[$j]` when `$v` is a glob.

xiaq:
Closing this PR as the approach I prefer will likely end up with a very different implementation.

PR #1144, opened 2020-09-08
Treat SIGPIPE termination as a successful exit

### LINKED ISSUE #952: SIGPIPE should not be considered a failure for data-producers in a pipeline
This is debatable perhaps, and one could argue that it's a bug in most Unix commands that currently exist:  But I think SIGPIPE should not be considered an error.

The rationale is this: It's not uncommon to create pipelines that will terminate one or more of their commands with SIGPIPE - and typically this does not mean that anything has failed, but rather that one of the processes in the pipeline has stopped consuming values, and the program that was producing values for that process terminated as a result. I think it's fair to argue that these producers should not return an error status in this case, but the usual (default, in fact!) behavior is to terminate on an unhandled SIGPIPE, or catch the signal and terminate with the equivalent error code.

So for instance, this produces a SIGPIPE:
`e:sort --random-sort /usr/share/dict/american-english | e:head -n 10`  (Produce a list of random words)

"head" reads the first 10 lines of inputs and then terminates, breaking the pipe. "sort" then receives SIGPIPE (or possibly, equivalent information via another means, depending on implementation) next time it writes to its output. Being unable to write out more data, it terminates. This is not an "exception"al condition, but rather a fundamental part of how stream processing works in shell pipelines.

On the flip side of this argument: a "graceful" shutdown of a pipeline is not the only condition that can produce a SIGPIPE. A program could fail with SIGPIPE due to a purely internal error, or due to a connection loss, etc. This is why I consider my argument "debatable" and say "it could be considered a bug in the programs called by the shell" - If the SIGPIPE occurs in a scenario where it should not be treated as an error, then arguably "e:sort" and so on should not terminate with a non-zero exit code. I think it's a fair argument that people should simply recognize this and capture errors or use try/catch when running a pipeline that could reasonably be expected to SIGPIPE. But it's very typical for SIGPIPE to simply indicate that a pipeline has shut down.  I don't think there is a set of criteria that can be applied to reliably distinguish between a "pipeline SIGPIPE" and an "internal error SIGPIPE" - the sequence in which processes terminate isn't a reliable indicator because SIGPIPE is triggered by the consumer closing its input, which could happen before termination - and if we said "SIGPIPEs aren't exceptions if they're generated by producers in a pipeline" there's always the chance that we're suppressing some true internal failure.

(I think treating process exit codes as exceptions is a good idea, though a challenging one to resolve against a tradition in which exit codes mostly don't matter...)

krader1961:
Resolves #952

### DIFF
--- a/NEXT-RELEASE.md
+++ b/NEXT-RELEASE.md
@@ -42,6 +42,10 @@ New features in the language:
 -   Rest variables and rest arguments are no longer restricted to the last
     variable.
 
+-   Command termination due to SIGPIPE no longer causes an exception and is
+    treated as a successful exit. See
+    https://github.com/elves/elvish/issues/952.
+
 New features in the standard library:
 
 -   A new `eval` command supports evaluating a dynamic piece of code in a
--- a/pkg/eval/exception_unix_test.go
+++ b/pkg/eval/exception_unix_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	. "github.com/elves/elvish/pkg/eval"
+	. "github.com/elves/elvish/pkg/eval/evaltest"
 
 	"github.com/elves/elvish/pkg/tt"
 )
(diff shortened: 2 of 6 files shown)


xiaq:
I don't think the argument for treating SIGPIPE as success or failure has changed meaningfully, so I am rejecting this PR. I'll update #952 with more of my thoughts on the topic.

PR #1145, opened 2020-09-09
Produce actual diffs when running make checkstyle

### LINKED ISSUE #1117: The `make check-style` done by the Cirrus CI environment could be more helpful
Yet again I pushed a change that resulted in this CI failure:

```
make checkstyle-md
Markdown files that need formatting:
  NEXT-RELEASE.md
Makefile:44: recipe for target 'checkstyle-md' failed
make: *** [checkstyle-md] Error 1

Exit status: 2
```

It would be more useful if a failure actually included the unexpected style violations limited to some reasonable maximum; e.g., 1000 lines.

P.S., The situation that caused me to open this issue involved a single, empty, line I introduced as a consequence of manually resolving a merge conflict. I ran `make style` manually then `git diff` and it reported this:

```
diff --git a/NEXT-RELEASE.md b/NEXT-RELEASE.md
index 7199dd81..a54f97bc 100644
--- a/NEXT-RELEASE.md
+++ b/NEXT-RELEASE.md
@@ -1,7 +1,6 @@
 This is the draft release notes for 0.15.0, scheduled to be released on
 2021-01-01.

-
 # Deprecated features

 The following deprecated features trigger a warning whenever the code is parsed
```

krader1961:
Resolves #1117

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,17 +33,15 @@ style:
 	find . -name '*.go' | xargs goimports -w
 	find . -name '*.md' | xargs prettier --tab-width 4 --prose-wrap always --write
 
+# Check if the style of the Go and Markdown files is correct without modifying
+# those files.
 checkstyle: checkstyle-go checkstyle-md
 
 checkstyle-go:
-	echo 'Go files that need formatting:'
-	! find . -name '*.go' | xargs goimports -l \
-		| sed 's/^/  /' | grep . && echo '  None!'
+	./tools/checkstyle-go.sh
 
 checkstyle-md:
-	echo 'Markdown files that need formatting:'
-	! find . -name '*.md' | xargs prettier --tab-width 4 --prose-wrap always -l \
-		| sed 's/^/  /' | grep . && echo '  None!'
+	./tools/checkstyle-md.sh
 
 .SILENT: checkstyle-go checkstyle-md
 .PHONY: default get generate test style checkstyle checkstyle-go checkstyle-md cover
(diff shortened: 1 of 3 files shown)


krader1961:
This was harder to implement, and verify it produces the correct results on my local VMs as well as the Cirrus CI, than I expected. The version I just forced push does the right thing on my local systems, whether or not there are any formatting errors, as well as on Cirrus CI. In particular the "check style" Cirrus CI targets now pass or fail depending on whether there are formatting errors.

xiaq:
I don't like the fact that this can change `.md` files.

krader1961:
> I don't like the fact that this can change .md files.

Fair enough. It's extremely annoying that `prettier` doesn't have a "diff" option. The simplest solution is to predicate the detailed report on running inside a CI environment. If the `.md` files are modified in the CI environment it doesn't matter. If not in a CI environment just provide a list of the names of the incorrectly formatted files.

krader1961:
At this point the `checktyle-md` target script is complicated enough that it would probably be better if it were moved into an external script. Such as a *scripts/checktyle-md.sh* file that is invoked from the *Makefile*. In general I'm not a fan of Makefiles having non-trivial embedded shell scripts. But that's a discussion, and change, for another day.

krader1961:
@xiaq, PTAL. When running on someone's workstation it should behave as before without modifying the local `.md` files. When running in a CI environment it should produce a detailed diff. I'd like to move the shell scripts embedded in the *Makefile* into a `scripts` directory but that's a change for another day.

krader1961:
I decided to move the shell scripts out of the *Makefile* and into files in the *tools* directory.

xiaq:
Yeah a markdown formatter with a diff option would be nice. Running the destructive operation only in CI is a reasonable workaround, at least it makes it possible for people without a local npm installation to see how to amend their PR,

PR #1147, opened 2020-09-11
Fix typos and clarify wording

krader1961:


xiaq:
I'd keep the comparison to `&&` to POSIX shell, but I'll merge this and add it back.

PR #1150, opened 2020-09-12
Fix hang reading value channel from a redirection

### LINKED ISSUE #600: Redirecting input of value-consuming functions to a file makes it not return
```
~> cat in.txt
abc
123
lorem
456
~> each [line]{ echo $line } < in.txt
abc
123
lorem
456
("each" does not return)
```

### LINKED ISSUE #1010: Redirecting input of value-consuming builtin from file hangs
This hangs:

```
echo "line" > file
each $echo~ < file
```

OTOH `cat file | each $echo~` is fine.

krader1961:
Fixes #600
Fixes #1010

### DIFF
--- a/pkg/eval/compile_effect.go
+++ b/pkg/eval/compile_effect.go
@@ -497,21 +497,31 @@ type invalidFD struct{ fd int }
 
 func (err invalidFD) Error() string { return fmt.Sprintf("invalid fd: %d", err.fd) }
 
+func (op *redirOp) fileChan() chan interface{} {
+	if op.mode == parse.Read {
+		// ClosedChan is used when redirecting from a named file because we
+		// don't support reading "values" (as opposed to bytes) from a file.
+		return ClosedChan
+	}
+	// TODO: Replace BlackholeChan for output with a value sink that will
+	// throw an exception if it sees any values since we don't support writing
+	// values to a file without first converting them to bytes.
+	return BlackholeChan
+}
+
 func (op *redirOp) exec(fm *Frame) error {
 	var dst int
-	if op.dstOp == nil {
-		// use default dst fd
+	if op.dstOp == nil { // the common case: use default destination file-descriptor
 		switch op.mode {
 		case parse.Read:
-			dst = 0
+			dst = 0 // stdin
 		case parse.Write, parse.ReadWrite, parse.Append:
-			dst = 1
+			dst = 1 // stdout
 		default:
 			return fm.errorpf(op, "bad RedirMode; parser bug")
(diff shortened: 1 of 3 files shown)


pkg/eval/compile_effect.go
```diff
@@ -497,21 +497,31 @@ type invalidFD struct{ fd int }
 
 func (err invalidFD) Error() string { return fmt.Sprintf("invalid fd: %d", err.fd) }
 
+func (op *redirOp) fileChan() chan interface{} {
```
xiaq, line 500: This method only depends on the `mode` field of `redirOp` and there is no other benefit of coupling it with `redirOp`. Turn it into `func fileChan(mode parse.RedirMode) chan interface{}`.

pkg/eval/compile_effect.go
```diff
@@ -497,21 +497,31 @@ type invalidFD struct{ fd int }
...
+
 func (op *redirOp) exec(fm *Frame) error {
 	var dst int
-	if op.dstOp == nil {
-		// use default dst fd
+	if op.dstOp == nil { // the common case: use default destination file-descriptor
 		switch op.mode {
 		case parse.Read:
-			dst = 0
+			dst = 0 // stdin
```
xiaq, line 517: I don't find it really useful to document that 0 is stdin and 1 is stdout; their use is too common in the codebase for such comments to be useful.

website/ref/language.md
```diff
@@ -1575,6 +1575,14 @@ Redirections may appear anywhere in the command, except at the beginning,
 although this may be restricted in future. It's usually good style to write
 redirections at the end of command forms.
 
+**Important:** Elvish only supports reading and writing bytes from/to the target
```
xiaq, line 1578: I'd use "Note" rather than "Important" for consistency with the rest of the doc.

xiaq (approved):
I'll merge this and apply the feedback in a followup commit.

pkg/eval/compile_effect.go
krader1961: The problem here is that `dst` is ambiguous. When I first read this function the magic numbers, in light of the ambiguous var name, was briefly confusing. Something like `dstFd` would be clearer. Too, while anyone who has programmed on UNIX for a non-trivial amount of time recognizes that 0==stdin and 1==stdout (and 2==stderr) in a context that takes a file descriptor, I'm not a fan of using those magic literals. You probably wouldn't use `10` as a fd literal even though in the context of your program it has a well defined meaning. You would probably give it a symbolic name.

Issue #1149, opened 2020-09-12
Writing values to an output redirection should fail


krader1961:
See #1010 and #600 for background. The short version is that `put x > f` should fail with an exception since Elvish does not allow writing values to files via redirection. Elvish silently discards the value(s) at this time which is suboptimal as it can lead to data corruption.

krader1961:
As best I can tell all existing uses of `BlackholeChan` should be replaced by a method that results in an exception if an attempt is made to write to the value channel. This includes the sole use of `DevNullPorts` in `func (ev *Evaler) Eval(op Op, cfg EvalCfg)`. If the consumer is not known to be able to handle value output any value output should cause an exception.

therealpxc:
It would be good for commands whose output is structured data to behave with an awareness of whether their output is a terminal or a pipeline. Elvish literal syntax makes sense for passing them to other Elvish functions, but for humans, visually reading a list of maps with many keys can be difficult; some kind of automatic pretty-printing for (pseudo-)maps and lists of uniformly structured (pseudo-maps) would be nice.

(For example, shells which share Elvish's emphasis on structured pipelines make generous use of pretty printers for commands that result in structured data. A sensible and widely-used convention for lists of maps which share keys is a table: a row for each map and a column for each key— something like that can make even pretty dense output friendly to human eyes.)

krader1961:
Presumably when you say "output is a terminal" you really mean "output is to anything other than an Elvish value consuming pipeline" (or words to that effect). I mention this because there are a couple of cases:

1) An interactive shell where the last command in a pipeline outputs values to the tty.

2) An interactive shell where the last command in a pipeline outputs values to a file via redirection.

3) A non-interactive shell (this is really just a variation of number two).

4) A command that writes to a pipeline whose RHS is an external command; thus incapable of receiving values.

> shells which share Elvish's emphasis on structured pipelines make generous use of pretty printers for commands that result in structured data.

I have never used Windows PowerShell or nushell. I've only glanced at their documentation. What do you think would be an appropriate method for Elvish to emulate that behavior? An implicit rewriting of values to text? An explicit command the output has to be piped through; e.g., `builtin:ls | pprint`? An option to the command; e.g., `builtin:ls &long`? What happens if the command outputs text and values?

therealpxc:
> Presumably when you say "output is a terminal" you really mean "output is to anything other than an Elvish value consuming pipeline"

[...]

PR #1154, opened 2020-09-16
Use port getter methods consistently

krader1961:
I was working on a fix for #1149 and noticed inconsistencies in how I/O
ports are referred to. This simply replaces are direct references to
fm.ports[] with the appropriate getter method. Go inlines the methods so
this doesn't incur any runtime overhead.


xiaq:
Oops, I've actually done exactly the same, but forgot to push. Thanks but I'll just close this now.

PR #1156, opened 2020-09-18
Try to resolve confusion around default bind of Alt-1 vs. Alt-, for last command mode

Cottser:
Coming here because I was going through https://elv.sh/learn/cookbook.html and wondering why `Alt + 1` wasn't doing anything for me.

Seems like the actual default config is still `Alt + ,` even though an old changelog says the contrary: https://elv.sh/blog/0.9-release-notes.html

I did also find 1ecbe33d19cf121fd602ae760d4d1f96a2b3e407 which seems to reinforce that `Alt + ,` is the expected default binding.

From a stock setup, macOS 10.15.6, elvish 0.14.1 installed via homebrew:

```
pprint $edit:insert:binding
[
 &Up=	<builtin <edit:history>start>
 &Right=	<builtin <edit>move-dot-right>
 &Left=	<builtin <edit>move-dot-left>
 &Home=	<builtin <edit>move-dot-sol>
 &Delete=	<builtin <edit>kill-rune-right>
 &End=	<builtin <edit>move-dot-eol>
 &Tab=	<builtin <edit:completion>:smart-start>
 &Enter=	<builtin <edit>smart-enter>
 &Backspace=	<builtin <edit>kill-rune-left>
 &Alt-Right=	<builtin <edit>move-dot-right-word>
 &Alt-Left=	<builtin <edit>move-dot-left-word>
 &'Alt-,'=	<builtin <edit:lastcmd>start>
 &Alt-.=	<builtin <edit>insert-last-word>
 &Alt-b=	<builtin <edit>move-dot-left-word>
 &Alt-f=	<builtin <edit>move-dot-right-word>
 &Alt-x=	<builtin <edit:minibuf>:start>
 &Ctrl-Right=	<builtin <edit>move-dot-right-word>
 &Ctrl-Left=	<builtin <edit>move-dot-left-word>
 &Ctrl-D=	<builtin <edit>return-eof>
 &Ctrl-H=	<builtin <edit>kill-rune-left>
 &Ctrl-K=	<builtin <edit>kill-line-right>
 &Ctrl-L=	<builtin <edit:location>start>
 &Ctrl-N=	<builtin <edit:navigation>start>
 &Ctrl-R=	<builtin <edit:histlist>start>
 &Ctrl-U=	<builtin <edit>kill-line-left>
 &Ctrl-V=	<builtin <edit>insert-raw>
 &Ctrl-W=	<builtin <edit>kill-word-left>
]
```

Searching the code I have to assume this default is coming from https://github.com/elves/elvish/blob/master/pkg/edit/default_bindings.go

I could also change this to update `default_bindings.go` and leave the docs alone of course, but that seems like a breaking change at least with the behaviour I'm seeing.

### DIFF
--- a/pkg/eval/mods/bundled/binding.elv.go
+++ b/pkg/eval/mods/bundled/binding.elv.go
@@ -19,7 +19,7 @@ fn install {
         &Alt-Down=      $edit:move-dot-down~
         &Alt-Enter=     $edit:insert-key~
         &Alt-.=         $edit:insert-last-word~
-        &Alt-1=         $edit:lastcmd:start~
+        &Alt-,=         $edit:lastcmd:start~
         &Alt-b=         $edit:move-dot-left-word~
         &Alt-f=         $edit:move-dot-right-word~
         &Alt-Backspace= $edit:kill-word-left~
--- a/website/learn/cookbook.md
+++ b/website/learn/cookbook.md
@@ -106,10 +106,9 @@ started quickly:
 
 -   Elvish doesn't support history expansion like `!!`. Instead, it has a "last
     command mode" offering the same functionality, triggered by <span
-    class="key">Alt-1</span> by default (resembling how you type `!` using
-    <span class="key">Shift-1</span>). In this mode, you can pick individual
+    class="key">Alt-,</span> by default. In this mode, you can pick individual
     arguments from the last command using numbers, or the entire command by
-    typing <span class="key">Alt-1</span> again.
+    typing <span class="key">Alt-,</span> again.
 
     This is showing me trying to fix a forgotten `sudo`:
 


zzamboni:
Interesting, I can confirm this bug in master, but I am positively, absolutely sure `Alt-1` was working fine when I started working with Elvish. But I haven't used the default keybinding since I wrote [bang-bang](https://github.com/zzamboni/elvish-modules/blob/master/bang-bang.org), so I hadn't noticed the change.

@xiaq looking at `git blame pkg/edit/default_bindings.go`, it seems this old keybinding was reintroduced in commit 720b687d2, as part of the new editor.

I think the bug here is in the code, not in the documentation, I would propose reverting it to the correct one.

Cottser:
@zzamboni thank you!

From a release perspective it looks like this regression was added in https://github.com/elves/elvish/releases/tag/v0.13-rc1 so it's been out in the wild for a while.

I'm happy to update this branch as needed.

Is https://github.com/elves/elvish/blob/master/pkg/eval/mods/bundled/binding.elv.go dead code at this point? I naively tried:

```
~> use binding
~> binding:install
Exception: variable $edit:insert:default~ not found
Traceback:
  [bundled binding], line 4:
            &Default=       $edit:insert:default~
  [tty 6], line 1:
    binding:install
```

xiaq:
Hmm, the bundled `binding` module is indeed dead code and should just be removed.

I actually went back and forth between `Alt-1` (for its similarity with `!`) and `Alt-,` (as it is right next to `Alt-.`). I now think `Alt-,` is the better option - bash and zsh uses Alt-number keys for repeat counts and I'd like to keep that possibility open.

PR #1167, opened 2020-10-05
Fix typos found by codespell

zhsj:


xiaq:
Thanks, I really should learn to spell `indices` correctly.

PR #1170, opened 2020-10-13
Navigation mode fix

krader1961:


pkg/edit/navigation.go
```diff
@@ -27,7 +27,21 @@ import (
...
+	fname := navigation.SelectedName(app)
+	if fname == "" {
+		// User pressed [enter] in an empty directory with nothing selected;
+		// therefore, do not insert an empty string in the buffer.
+		return
+	}
+
+	// Prepend a space to the file name only if not at the start of the line
+	// and the previous char is not a space.
+	cursor := app.CodeArea().CopyState().Buffer.Dot
```
xiaq, line 39: This operation should be a transaction inside app.CodeArea().MutateState.

pkg/edit/navigation_test.go
```diff
@@ -63,23 +66,91 @@ func TestNavigation_WidthRatio(t *testing.T) {
...
+		"###### ++++++++++++++++++ ",
+		"        e                ", Styles,
+		"       //////////////////",
+	)
+
+	// Validate that inserting an empty directory is a nop.
+	f.TTYCtrl.Inject(term.K(ui.Down))          // select "e"
+	f.TTYCtrl.Inject(term.K(ui.Right))         // move into the empty "e" directory
+	f.TTYCtrl.Inject(term.K(ui.Enter, ui.Alt)) // insert nothing since the dir is empty
+	f.TestTTY(t,
```
xiaq, line 116: This test here might pass even if pressing Enter inserts something like `''` - it could catch the state before that string is entered.

This is because the editor is running on a different goroutine, and TestTTY only tests that the expected state appears once, and doesn't guarantee that it is the steady state.

You can fix this by adding something after the Enter, for example typing a '#' and expect to see only a `#` in the code area.

pkg/edit/navigation_test.go
```diff
@@ -63,23 +66,91 @@ func TestNavigation_WidthRatio(t *testing.T) {
...
+//
+// 2) Inserting the "selection" in an empty directory inserts nothing rather
+// than the literal ` ''`  that used to be inserted.
+//
+// 3) Inserting a selection when the CLI cursor is not at the start of the
+// edit buffer but the preceding char is a space does not prepend another
+// space.
+//
+// See https://github.com/elves/elvish/issues/1169
+func TestNavigationCornerCases(t *testing.T) {
```
xiaq, line 92: Please break this test into multiple tests.

pkg/edit/navigation_test.go
```diff
@@ -63,23 +66,91 @@ func TestNavigation_WidthRatio(t *testing.T) {
...
 func setupNav() (*fixture, func()) {
 	f := setup()
 	restoreLsColors := lscolors.WithTestLsColors()
 
-	testutil.ApplyDir(testutil.Dir{"d": testutil.Dir{"a": ""}})
-	err := os.Chdir("d")
-	if err != nil {
-		panic(err)
-	}
+	// Path "d/a" is an empty file and "d/e" is an empty directory.
```
xiaq, line 150: There is no need to change this.

pkg/testutil/must.go
```diff
@@ -51,3 +51,18 @@ func MustWriteFile(filename string, data []byte, perm os.FileMode) {
...
+
+// MustChdir simply wraps os.Chdir and panics if it fails.
+func MustChdir(dir string) {
+	err := os.Chdir(dir)
+	if err != nil {
+		panic(err)
+	}
+}
+
+// Must provides a single-line solution to a common pattern in unit tests.
```
xiaq, line 63: Describe what this actually does.

pkg/testutil/must.go
```diff
@@ -51,3 +51,18 @@ func MustWriteFile(filename string, data []byte, perm os.FileMode) {
 		panic(err)
 	}
 }
+
+// MustChdir simply wraps os.Chdir and panics if it fails.
```
xiaq, line 55: Describe what a function does, not what it is:

> MustChdir calls os.Chdir and panics if it fails.

pkg/eval/chdir_test.go
```diff
@@ -73,17 +73,13 @@ func TestChdirError(t *testing.T) {
 	}
 }
 
-// Saves the current working directory, and returns a function for returning to
-// it.
+// Save the current working directory and return a function for returning to it.
```
xiaq, line 76: There is no need to change this; using third-person singular for unexported functions is the convention in the codebase.

krader1961: Yes, that is the case. Reverted. Having said that it's not obvious the passive voice implied by third-person singular is preferable. Not to mention the comma is superfluous.

pkg/testutil/must.go
krader1961: Done

pkg/testutil/must.go
krader1961: Done

pkg/edit/navigation_test.go
krader1961: I beg to differ and feel pretty strongly about this aspect of my change. The `testutil.ApplyDir` function is clever but not clear. When I made this change I couldn't see how to use it to create an empty directory. I now see that should be possible by doing `"e":  testutil.Dir{}`. Ugh! I actually want to see all uses of `testutil.ApplyDir()` replaced by their obvious counterparts. I can't see any justification for this API given it makes it harder for people to understand what a unit test is doing. Especially given that they already need to understand what functions like `testutil.MustMkdirAll` and `testutil.MustCreateEmpty` do. In that sense `testutil.ApplyDir` is nothing more than additional cognitive load with no real benefit.

krader1961: Coincidentally I watched https://www.youtube.com/watch?v=NwEuRO_w8HE, "Opening keynote: Clear is better than clever - GopherCon SG 2019" just after writing my previous comment since it popped up in my Youtube recommendations. It seems applicable to this discussion.

pkg/edit/navigation_test.go
krader1961: Done

pkg/edit/navigation_test.go
krader1961: Good point. Fixed. However, I feel like I still don't understand why what you wrote is needed. Doesn't this apply to the other tests in this module? For clarity shouldn't there be a method we can call to force syncing the state of the two goroutines?

pkg/edit/navigation.go
krader1961: Fixed

krader1961:
WTF? The appveyor (MS Windows) test failure is because the Elvish editor is treating command "az" as a valid command. That seems pretty unlikely. Also, that test passes on my Windows 10 system (i.e., "az" is an invalid command). However, I experienced a similar, unexpected, failure on macOS with an earlier version of this change when the filename was "ab" which was styled as "valid" (green). Which suggests there is a deeper problem with respect to CLI editor highlighting.

krader1961:
Changing "az" to "za" fixes the test failure on Appveyor. Apparently that environment has an `az` external command. I'm sort of tempted to modify the testing framework to provide a way to indicate we don't care what styling is used for a particular segment of the command line. However, that's hard to justify since it isn't particularly difficult to pick command names that don't have such conflicts with expectations.

krader1961:
@xiaq, PTAL. The updated commit addresses all of your points. Including "the editor is running on a different goroutine". I added an additional char, via `f.TTYCtrl.Inject`, at the end of the navigation mode to ensure `TestTTY` in that test fails if it "catch(es) the state before that string (from navigation mode) is entered." However, I feel like that is doing nothing other than ensuring that particular test doesn't have a false positive due to timing. It seems to me we should address the core synchronization problem that can result in that particular test being flakey; i.e., have false negative outcomes.

pkg/edit/navigation_test.go
xiaq: The problem here is that you are testing the absence of change of the UI; typically UI tests test against the change of the UI.

pkg/edit/navigation_test.go
xiaq: I don't see a strong objective argument against `testutil.ApplyDir`. Please keep it. If you'd like to discuss this further please raise the issue separately.

krader1961: @xiaq, I see that you introduced `testutil.Dir` and `testutil.ApplyDir` a year ago. While that API has relatively few uses at this time it seems clear you intend it to be the default, most often used, API for creating unit test directory trees. Most of my objection to this relatively new API, in the context of this change, is because the existing uses are a) rare, and b) formatted in a fashion that actually obscured their behavior. I just pushed an amendment to this P.R. that should resolve this aspect of my P.R. and illustrate what I think is a clearer expression of the `testutil.Dir` pattern.

krader1961:
Okay, I reinstated the use of `testutil.ApplyDir` with what I think is a better way to format the associated data structure to reflect its meaning.

pkg/edit/navigation.go
```diff
@@ -27,7 +27,22 @@ import (
...
+		// User pressed [enter] in an empty directory with nothing selected;
+		// therefore, do not insert an empty string in the buffer.
+		return
+	}
+
+	// Insert a space only if not at the start of the CodeArea buffer and the
+	// previous char is not a space. Then insert the selected filename.
+	app.CodeArea().MutateState(func(s *cli.CodeAreaState) {
+		cursor := s.Buffer.Dot
+		if cursor != 0 && s.Buffer.Content[cursor-1] != ' ' {
```
xiaq, line 41: A space is also not needed if the cursor is at the beginning of a new line.

pkg/edit/navigation_test.go
```diff
@@ -63,23 +66,102 @@ func TestNavigation_WidthRatio(t *testing.T) {
...
 		"################ ++++++++++++++++ ",
+		"                  e              ", Styles,
+		"                 ////////////////",
 	)
 }
 
+// Test corner case: Inserting a selection when the CLI cursor is not at the
+// start of the edit buffer, but the preceding char is a space, does not
+// insert another space.
+func TestNavigationCornerCaseRedundantSpace(t *testing.T) {
```
xiaq, line 82: I'd prefer to leave out the "corner case" moniker when describing a unit test - too many unit tests would satisfy some definition of "corner case"; instead, describe completely what the behavior is being tested. In this case, I'd name this `TestNavigation_EnterDoesNotInsertSpaceAfterSpace`.

pkg/edit/navigation_test.go
```diff
@@ -63,23 +66,102 @@ func TestNavigation_WidthRatio(t *testing.T) {
...
+	f, cleanup := setupNav()
+	defer cleanup()
+
+	f.TTYCtrl.Inject(term.K('x'))              // user presses 'x'
+	f.TTYCtrl.Inject(term.K(' '))              // user presses Space
+	f.TTYCtrl.Inject(term.K('N', ui.Ctrl))     // begin navigation mode
+	f.TTYCtrl.Inject(term.K(ui.Down))          // select "e"
+	f.TTYCtrl.Inject(term.K(ui.Enter, ui.Alt)) // insert the "e" file name
```
xiaq, line 90: I'd use Enter to make the TestTTY call below slightly simpler.

pkg/edit/navigation_test.go
```diff
@@ -63,23 +66,102 @@ func TestNavigation_WidthRatio(t *testing.T) {
...
+	f.TTYCtrl.Inject(term.K('z'))              // user presses 'z'
```
xiaq, line 91: This isn't really necessary, since the Alt-Enter press about would result in a buffer change; these kind of actions are only necessary when you are testing the *absense* of change.

pkg/testutil/must.go
```diff
@@ -51,3 +51,24 @@ func MustWriteFile(filename string, data []byte, perm os.FileMode) {
...
+func MustChdir(dir string) {
+	err := os.Chdir(dir)
+	if err != nil {
+		panic(err)
+	}
+}
+
+// Must panics if the error value is not nil. It is typically used like this:
+//
+//   testutil.Must(a_function())
```
xiaq, line 65: Go functions never use snake_case; use aFunction instead.

xiaq (approved):
LGTM other than a minor feature and some stylistic feedback. I'll merge this and apply the changes myself.

PR #1173, opened 2020-10-18
Set read-only var error should include var name

### LINKED ISSUE #255: Include variable name in error message when setting read-only variable
```
~> try failed; except false; tried
<interact>:1:20: eval error: read-only; cannot be set
try failed; except false; tried
```

krader1961:
Modifying a read-only var should include the var name in the error message.

Fixes #255

### DIFF
--- a/pkg/edit/navigation.go
+++ b/pkg/edit/navigation.go
@@ -89,7 +89,7 @@ func initNavigation(ed *Editor, ev *eval.Evaler) {
 	binding := newMapBinding(ed, ev, bindingVar)
 	widthRatioVar := newListVar(vals.MakeList(1.0, 3.0, 4.0))
 
-	selectedFileVar := vars.FromGet(func() interface{} {
+	selectedFileVar := vars.FromGet("selected-file", func() interface{} {
 		name := navigation.SelectedName(ed.app)
 		if name == "" {
 			return nil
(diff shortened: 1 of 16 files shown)


krader1961:
I considered qualifying some of the read-only var name literals with their namespace (e.g., `"builtin:pid"` rather than `"pid"`) but it's not clear eliminating any ambiguity is worth the verbosity.

xiaq:
This is is not the ideal approach to fix the issue: it doesn't get the namespace right in the error message, and it unnecessarily couples a `Var` with its name.

A better place to fix this is to fix this within [`(*assignOp).exec`](https://github.com/elves/elvish/blob/d27cee0724a0cdf3ddc95beb721853daa214dd99/pkg/eval/compile_lvalue.go#L93). This function has access to fully qualified name of the LHS, and can amend the error message to use it. For example, when this `.Set` call returns an error, the current behavior is to use that error directly:

https://github.com/elves/elvish/blob/d27cee0724a0cdf3ddc95beb721853daa214dd99/pkg/eval/compile_lvalue.go#L115-L118

Instead of using that error directly, build a new error that includes the fully qualified name of the variable being assigned.

krader1961:
> This is is not the ideal approach to fix the issue

Fair enough but since you closed this does that mean you'll be fixing this in the immediate future?

PR #1177, opened 2020-10-22
Make key binding syntax more consistent

### LINKED ISSUE #1163: Document and normalize handling of symbolic UI key names
While looking into issue #912 I noticed three aspect of symbolic UI key names that needs some discussion. Here I'm talking about a binding map assignment in a statement such as `edit:insert:binding[Alt-Left] = { echo alt-left >/dev/tty }`.

1) The modifier prefix letter-case doesn't matter. You can use `Alt-`, `alt-`, `aLt-`, etc.
*) I think this behavior is desirable and should be retained. It also needs to be documented but non-canonical letter-case should be discouraged.
2) There are undocumented modifiers. For example, you can use `a-` (or `A-`) instead of `Alt-`.
*) Whether these shortened aliases should be retained is an open question. When I proposed allowing `[d]` or `[type:d]` as shorter forms of the `[type:dir]` wildcard modifier @xiaq rejected adding those aliases. It seems like the symbolic key names should also, if we're going to be consistent, not allow single letter modifiers; i.e., `A-` should not be an alias for `Alt-`.
3) Function key names must match the letter case exactly. For example, `Left` is okay while `left` is not.
*) I think the behavior should be modified to perform case insensitive matching of those names.

krader1961:
Document the format of key names. Make function key name matching
case-insensitive for consistency with how key modifers (e.g., "Alt")
are handled.

Resolves #1163

### DIFF
--- a/NEXT-RELEASE.md
+++ b/NEXT-RELEASE.md
@@ -11,6 +11,10 @@ This is the draft release notes for 0.15.0, scheduled to be released on
     -   The `rest-arg` field now contains the index of the rest argument,
         instead of the name.
 
+-   Key modifiers are no longer case insensitive. Thus `Alt` is still recognized
+    but `alt` is not. This makes key modifier parsing consistent with key names.
+    See [#1163](https://b.elv.sh/1163).
+
 # Deprecated features
 
 The following deprecated features trigger a warning whenever the code is parsed
--- a/pkg/ui/key.go
+++ b/pkg/ui/key.go
@@ -25,7 +25,7 @@ func K(r rune, mods ...Mod) Key {
 	return Key{r, mod}
 }
 
-// Default is used in the key binding table to indicate default binding.
+// Default is used in the key binding table to indicate a default binding.
 var Default = Key{DefaultBindingRune, 0}
 
 // Mod represents a modifier key.
@@ -44,10 +44,11 @@ const (
 
 const functionKeyOffset = 1000
 
-// Special negative runes to represent function keys, used in the Rune field of
-// the Key struct.
+// Special negative runes to represent function keys, used in the Rune field
+// of the Key struct. This also has a few function names that are aliases for
+// simple runes. See keyNames below for mapping these values to strings.
 const (
-	// DefaultBindingRune is a special value to represent default binding.
+	// DefaultBindingRune is a special value to represent a default binding.
 	DefaultBindingRune rune = iota - functionKeyOffset
 
 	F1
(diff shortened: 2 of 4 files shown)


krader1961:
I'm not entirely happy with this change since it illustrates why the (sometimes) case-insensitivity greatly complicates the documentation. This also doesn't properly document the behavior of the `Shift` modifier but that's for another day. But it does at least document the current behavior and makes `Ctrl-?` consistent with `"\c?"`.

krader1961:
Also, I'll be happy to add unit tests for the changed behaviors if you're otherwise happy with simply documenting the current behavior and making function key names case insensitive to match how key modifiers are matched.

xiaq:
I'd rather make key bindings consistently case-sensitive, and provide an easy way to insert the name of a key in CLI.

krader1961:
> I'd rather make key bindings consistently case-sensitive, ...

That's my preference as well. I only implemented the case insensitive solution because you had earlier implied a preference for it, but perhaps I misinterpreted what you wrote.

> ... and provide an easy way to insert the name of a key in CLI.

Is there any reason that enhancement needs to be coupled with making the case sensitivity consistent, fixing the handling of `Ctrl-?`, and documenting these symbols?

krader1961:
FWIW, I modified my change to make the key modifier case sensitive -- rather than make function names case insensitive. That results in several unit test failures since they expect key modifiers to be case insensitive.

krader1961:
@xiaq, PTAL. I've modified the change to make key modifiers case sensitive. Obviously a backward incompatible change so I've noted that in the release notes and modified the unit tests accordingly. I also opened issue #1180 to track the desirability of adding a mechanism for interactively inserting the symbolic name for a key.

krader1961:
@xiaq Ping? This should be backward compatible with the exception of key modifiers no longer being case-insensitive. Which I have documented and is unlikely to cause problems for current Elvish users.

NEXT-RELEASE.md
```diff
@@ -11,6 +11,10 @@ This is the draft release notes for 0.15.0, scheduled to be released on
     -   The `rest-arg` field now contains the index of the rest argument,
         instead of the name.
 
+-   Key modifiers are no longer case insensitive. Thus `Alt` is still recognized
```
xiaq, line 14: > Thus `Alt` is still recognized

I'd stick to "for example" for examples

pkg/ui/key.go
```diff
@@ -122,57 +140,64 @@ func (k Key) String() string {
...
+		} else {
 			return Key{}, fmt.Errorf("bad modifier: %s", parse.Quote(modname))
 		}
-		k.Mod |= mod
-		s = s[i+1:]
 	}
 
 	if len(s) == 1 {
 		k.Rune = rune(s[0])
+		if k.Rune < ' ' {
```
xiaq, line 192: Using the numerical value of 0x20 is a bit clearer here.

pkg/ui/key.go
```diff
@@ -122,57 +140,64 @@ func (k Key) String() string {
...
+			if k.Mod&Ctrl != 0 {
+				return Key{}, fmt.Errorf("Ctrl modifier with literal control char: %q", k.Rune)
+			}
+			// Convert literal control char to the equivalent canonical form;
+			// e.g., "\e" to Ctrl-'[' and "\t" to Ctrl-I.
+			k.Mod |= Ctrl
+			k.Rune += '@'
```
xiaq, line 199: Using the numerical value of 0x40 is much clearer than `@` here.

pkg/ui/key.go
```diff
@@ -182,33 +207,29 @@ func ParseKey(s string) (Key, error) {
...
+			// Normalize Ctrl-I to Tab, Ctrl-J to Enter, and Ctrl-? to Backspace.
 			if k.Rune == 'I' {
 				k.Mod &= ^Ctrl
 				k.Rune = Tab
 			} else if k.Rune == 'J' {
 				k.Mod &= ^Ctrl
 				k.Rune = Enter
+			} else if k.Rune == '?' {
+				k.Mod &= ^Ctrl
+				k.Rune = Backspace
```
xiaq, line 219: I don't think typing `Ctrl-?` actually produces backspace in any modern terminal, so I'd rather leave this out.

pkg/ui/key_test.go
```diff
@@ -51,31 +51,44 @@ var parseKeyTests = []struct {
...
-	{s: "a-x", wantKey: Key{'x', Alt}},
-	{s: "a-X", wantKey: Key{'X', Alt}},
+	{s: "A-x", wantKey: Key{'x', Alt}},
+	{s: "A-X", wantKey: Key{'X', Alt}},
 
 	// Ctrl- keys are case-insensitive.
 	{s: "C-x", wantKey: Key{'X', Ctrl}},
 	{s: "C-X", wantKey: Key{'X', Ctrl}},
 
+	// Literal control chars are equivalent to the preferred Ctrl-? formulation.
```
xiaq, line 61: > Ctrl-?

I'd use "Ctrl-<key>" instead as this can be misunderstood as the Ctrl-? key combination.

pkg/ui/key_test.go
```diff
@@ -51,31 +51,44 @@ var parseKeyTests = []struct {
...
+	{s: "\033", wantKey: Key{'[', Ctrl}},
+	{s: "\t", wantKey: K(Tab)},
```
xiaq, line 63: This is duplicate with a test below. Also, parsing of `\t` doesn't go through the code path for parsing control characters.

pkg/ui/key_test.go
```diff
@@ -51,31 +51,44 @@ var parseKeyTests = []struct {
...
+	{s: "Alt-\t", wantKey: Key{Tab, Alt}},
+	{s: "\n", wantKey: K(Enter)},
```
xiaq, line 65: Same

xiaq (approved):
Mostly LGTM. A few small issues that I'll address after merging this.

Issue #1178, opened 2020-10-30
Could not delete word in some CodeAreas


mrlzh:
While pressing delete key in xshell，elvish get a event "Key{0x48, Ctrl}".
This makes it impossible to delete words in some CodeAreas likes "COMPLETING argumen"t or "HISTORY (dedup on) ".

Reproduction method：
1.Using the xshell or hardcode 'pkg/cli/loop.go' like this:
```
case event := <-lp.inputCh:
// here is hardcode
if _,e:=event.(term.KeyEvent);e && event.(term.KeyEvent).Mod==0 &&event.(term.KeyEvent).Rune==int32(127){
    event=term.KeyEvent{Mod:4,Rune: 0x48}
}
// Consume all events in the channel to minimize redraws.
consumeAllEvents:
```
2.Press ctrl-R to enter "HISTORY (dedup on) ", and input some words.You can see that you can not delete word anymore

PR #1179, opened 2020-10-30
Fix:https://github.com/elves/elvish/issues/1178

mrlzh:


krader1961:
Your problem statement says "impossible to delete words" using Ctrl-H in history mode. But the Backspace key deletes a single rune (i.e., character), not a word. Did you mean to say "impossible to delete characters/runes"? Or are you referring to making Alt-Ctrl-H behave the same as Alt-Backspace?

Also, your change implements a needlessly opaque, and confusing, definition of `'H'` as `0x48` and giving that rune a misleading symbolic name.

xiaq:
Hi, thanks for the PR. As @krader1961 pointed out, 0x48 is just the letter `H`, there is no need to define a constant for it.

@krader1961, I believe @mrlzh meant using Backspace to delete runes; some terminals send ^H when Backspace is pressed.

krader1961:
> some terminals send ^H when Backspace is pressed.

Yes, that is true. 😄 In fact, hardware terminals such as the Televideo 9xx series (including the model 925 and 950 that I used three decades ago) supported configuring which character that key transmitted but they defaulted to Ctrl-H. The question is whether Delete (0x7F) (mistakenly labeled "Backspace" in the source code) and Ctrl-H (0x08 aka "bs" in ASCII) should be equivalent by default. My vote is yes, the two runes should behave identically by default. See, for example, https://www.computerhope.com/issues/ch001389.htm.

xiaq:
#1178 was fixed in b11fe9d.

PR #1181, opened 2020-11-14
[cleanup] trivial golint suggested changes

krader1961:
As with commit #eb2a792 I acknowledge that `golint` recommendations are
controversial and should not automatically be acted on. Nonetheless,
this change fixes legitimate lint issues such as copy/paste cleanups,
style problems I introduced (`i += 1` versus `i++`) or method/var comments
that are not in the preferred form.

### DIFF
--- a/pkg/ui/key_test.go
+++ b/pkg/ui/key_test.go
@@ -7,7 +7,7 @@ import (
 	"github.com/xiaq/persistent/hash"
 )
 
-var kTests = []struct {
+var identityTests = []struct {
 	k1 Key
 	k2 Key
 }{
@@ -17,7 +17,7 @@ var kTests = []struct {
 }
 
 func TestK(t *testing.T) {
-	for _, test := range kTests {
+	for _, test := range identityTests {
 		if test.k1 != test.k2 {
 			t.Errorf("%v != %v", test.k1, test.k2)
 		}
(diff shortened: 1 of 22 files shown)


krader1961:
Note that `golint ./... | grep -v 'should have comment or be unexported'` still enumerates six style problems:

```
pkg/env/env.go:10:2: don't use ALL_CAPS in Go names; use CamelCase
pkg/env/env.go:12:2: don't use ALL_CAPS in Go names; use CamelCase
pkg/env/env.go:18:2: don't use ALL_CAPS in Go names; use CamelCase
pkg/eval/eval.go:67:6: type name will be used as eval.EvalCfg by other packages, and that stutters; consider calling this Cfg
pkg/parse/parse.go:37:6: func name will be used as parse.ParseWithDeprecation by other packages, and that stutters; consider calling this WithDepreca>
pkg/parse/parse.go:46:6: func name will be used as parse.ParseAs by other packages, and that stutters; consider calling this As
```

However, changing the code to eliminate those warnings is certain to be controversial. Which is why this change does not address those warnings.

pkg/ui/key_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/xiaq/persistent/hash"
 )
 
-var kTests = []struct {
+var identityTests = []struct {
```
xiaq, line 10: This is the list of test cases for the function `K`, so `kTests` is more accurate.

xiaq (approved):
Thanks for cleaning everything up; there is one place where I'd keep the code as is, but otherwise LGTM.

PR #1188, opened 2020-11-28
In prompts implicitly coerce float64 to string

### LINKED ISSUE #1186: prompts should support type float64 as well as string
Sorry, @zzamboni, but while reviewing my commit history I also noticed one of your changes:

https://github.com/elves/elvish/commit/a062be0c052332afdcd821f4a95289cf92d76bf2

To reproduce the problem that change in the documentation addresses start an interactive Elvish shell and type these statements:

```
> n = 1
> edit:prompt = { sleep 2; put $n; n = (+ $n 1); put ': ' }
```

The change in that commit "fixes" the problem with the documentation but introduces an ambiguity since it mixes byte and value output in the prompt function. The order in which byte and value streams are processed is currently undefined. Thus, replacing the first `put` with `to-string`, "fixes" the problem but there is no guarantee that the number, written to the byte stream, will appear before the `": "` string that is written to the value stream. The proper fix is to change the prompt rendition code to support implicit coercion of any data type (including float64) that has a `String` method so that prompt functions can consistently write to the value stream.

krader1961:
Including the float64 data type in the prompt value stream causes an error.
This causes those values to be implicitly converted to a string as happens
everywhere else in Elvish.

I initially intended to modify the code to do the implicit string coercion
for any type not otherwise explicitly handled by the pkg/ui `Concat`
methods. I decided against that approach because doing so doesn't make
sense for some types that might appear in the value stream; e.g., Elvish
exceptions.

Fixes #1186

### DIFF
--- a/pkg/ui/text.go
+++ b/pkg/ui/text.go
@@ -61,11 +61,15 @@ func (t Text) Index(k interface{}) (interface{}, error) {
 	}
 }
 
-// Concat implements Text+string, Text+Segment and Text+Text.
+// Concat implements Text+string, Text+Segment and Text+Text. Where "string"
+// refers to strings and data types, such as float64, where it is reasonable
+// to implicitly convert the type to a string.
 func (t Text) Concat(rhs interface{}) (interface{}, error) {
 	switch rhs := rhs.(type) {
 	case string:
 		return Concat(t, T(rhs)), nil
+	case float64:
+		return Concat(t, T(vals.ToString(rhs))), nil
 	case *Segment:
 		return Concat(t, Text{rhs}), nil
 	case Text:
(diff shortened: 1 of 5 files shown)


pkg/ui/text.go
```diff
@@ -61,11 +61,15 @@ func (t Text) Index(k interface{}) (interface{}, error) {
 	}
 }
 
-// Concat implements Text+string, Text+Segment and Text+Text.
+// Concat implements Text+string, Text+Segment and Text+Text. Where "string"
+// refers to strings and data types, such as float64, where it is reasonable
+// to implicitly convert the type to a string.
```
xiaq, line 66: I'd rather not twist the definition of "string" here, and just document that this implements Text+string and Text+float64.

If the need arises for more types, it might be more convenient to say something like "string-like value" or "value convertible to string", but now it's just two types.

xiaq (approved):

PR #1190, opened 2020-11-28
Make test coverage statistics more accurate

### LINKED ISSUE #1062: Improve unit test coverage to 85% before 0.15.0 release

### LINKED ISSUE #1187: Test coverage does not count lines executed by a test in a different package
While working on issue #1186 I was perplexed that `make coverage` showed no executions of these methods:

https://github.com/elves/elvish/blob/6f3b8a787b8b8eb98e1436316f0d4cc1d00e4604/pkg/ui/text.go#L64-L86

Despite the fact that adding debugging print statements (or `panic()` calls) showed those methods were being run during test coverage. This [article](https://www.ory.sh/golang-go-code-coverage-accurate/) explains why that incorrect coverage occurs.

It is my opinion that indirect coverage is meaningful, and the default Go toolchain behavior is incorrect (i.e. undesirable), for this project. Will there be corner cases that are incorrectly considered to be covered via an indirect test? Maybe. Nonetheless, I don't think that adding explicit unit tests of the uncovered functions is more valuable than implicit tests from other packages. It's more useful to assume that indirect execution during a unit test (i.e., from a different package) is meaningful for this project's test coverage metrics.

krader1961:
The default behavior of `go test -cover` ignores coverage by integration
tests. That is, it ignores coverage by tests in a different package. This
penalizes Elvish since Elvish has a lot of integration tests that are
meant to explicitly exercise code in other packages.

Resolves #1062
Resolves #1187

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,11 +4,13 @@ environment:
   GOPATH: c:\gopath
   ELVISH_TEST_TIME_SCALE: 10
 install:
-  # Needed since the codecov uploader depends on a UNIX "find" command.
+  # The codecov uploader depends on a POSIX shell and commands such as `find`
+  # so ensure the relevant Windows MSYS2 directory is in the external command
+  # search path.
   - set PATH=C:\msys64\usr\bin;%PATH%
 build: off
 before_test:
   - go version
 test_script:
-  - go test -coverprofile=coverage -covermode=set ./...
-  - curl -s https://codecov.io/bash -o codecov && bash codecov -f coverage || ver > nul
+  - go test -coverprofile=cover -coverpkg=./pkg/... ./pkg/...
+  - curl -s https://codecov.io/bash -o codecov && bash codecov -f cover || ver > nul
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,9 @@ test:
 # Generate a basic test coverage report. This will open the report in your
 # browser. See also https://codecov.io/gh/elves/elvish/.
 cover:
-	go test -covermode=set -coverprofile=$@ ./...
-	go tool cover -html=$@
+	go test -coverprofile=cover -coverpkg=./pkg/... ./pkg/...
+	go tool cover -html=cover
+	go tool cover -func=cover | tail -1 | awk '{ print "Overall coverage:", $$NF }'
 
 # Ensure the style of Go and Markdown source files is consistent.
 style:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -51,8 +51,8 @@ test_task:
   test_script: go test $TEST_FLAG ./...
   upload_coverage_script:
     - test -z $SKIP_UPLOAD_COVERAGE || exit 0
-    - go test -coverprofile=coverage -covermode=set ./...
-    - curl -s https://codecov.io/bash -o codecov && bash codecov -f coverage -t $CODECOV_TOKEN || true
+    - go test -coverprofile=cover -coverpkg=./pkg/... ./pkg/...
+    - curl -s https://codecov.io/bash -o codecov && bash codecov -f cover -t $CODECOV_TOKEN || true
 
 checkstyle_go_task:
   name: Check the style of .go files


krader1961:
Hmmm, codecov is reporting 82.54% coverage (an increase of 2.66%) with this change. That is considerably less than the 88.0% I am seeing on my Ubuntu system and 88.4% on my macOS server. TBD is the cause of that discrepancy. This page shows at least some of the expected improvements: https://codecov.io/gh/elves/elvish/pull/1190/changes.

krader1961:
The reason Codecov reports a much lower coverage percentage appears to be because it doesn't correctly handle `import` blocks: https://codecov.io/gh/elves/elvish/src/8470a24e98c3d8ea4ad696297506bfd4457f4a30/pkg/cli/addons/histwalk/histwalk.go

Codecov reports 93.44% coverage whereas `go tool cover -html` reports 96.5%. The two tools agree on the coverage of the rest of the file; i.e., the statements after the `import` block. I spot checked a few other files and observed the same problem.

So, not only are the coverage metrics low due to how `go test -cover` works by default they are even lower because Codecov (at least as configured by this project) doesn't consider `import` blocks to have been executed even though those statements do not even appear in the coverage data file (which means they should be treated as comments).

krader1961:
Ugh! The Codecov metrics are even worse than my previous comment implied. See https://codecov.io/gh/elves/elvish/src/8470a24e98c3d8ea4ad696297506bfd4457f4a30/pkg/cli/lscolors/lscolors.go which shows a lot of red lines and a coverage metric of 52.38%. The correct value is 62.1%. Codecov is not only include `import` blocks it is basically including any line that does not begin with `//` (i.e., a comment) in the coverage metric.

krader1961:
I spent two hours trying to figure out how to get codecov.io to correctly process Go coverage data files. I thought there might be an option to the `codecov` upload script this project uses to force the correct interpretation of the coverage data file. A lot of googling and searching the [Codecov community forums](https://community.codecov.io/) didn't yield any obvious reports that this is a known problem or how to fix/workaround the problem.

My recommendation is to use the cover percentage as output by the last line of `make cover` rather than the overall coverage percentage reported by Codecov. By the more accurate, `go tool cover -func`, metric the project already exceeds the 85% coverage goal of issue #1062.

xiaq:
I am not sure the custom script is necessary. Running `go test -coverprofile=coverage -coverpkg=./... ./...` seems to produce coverage report that includes cross-package invocations; I checked the coverage for `pkg/parse/parseutil/parseutil.go` and it reports it as 94.7%, instead of 0%. Have you tried that, and if you have, what's wrong with it?

Re codecov counting `import` statements as uncovered lines, the problem doesn't seem to exist now; for example, in https://codecov.io/gh/elves/elvish/src/master/website/cmd/elvdoc/main.go, the `import` lines are uncolored (as opposed to red), meaning that codecov understands that this is a line that should be ignored from coverage reporting. Maybe the problem you've run into has something to do with the format of the output produced by the custom script?

Also, I suggest documenting the change in a more objective way (in the PR title, commit message etc.): "more accurate" is subjective and depends on what you consider desirable; whereas something like "Include cross-package invocations in test coverage report" describes the change purely in terms of what it does. You can still document the rationale for doing this in the commit message, i.e. we have agreed that this is a more desirable way to calculate test coverage for Elvish.

krader1961:
The first thing I tried, before resorting to writing that script, is using the `-coverpkg` flag. Sadly, it doesn't work as you might expect when used with `-coverprofile`. See this article among others for why: https://www.ory.sh/golang-go-code-coverage-accurate/. That article was the inspiration for the script I wrote.

Codecov is still wrong. See, for example, the 79.95% overall coverage reported for the printf P.R. Yet `go tool cover` reports 85.5%. When I look at the link you provided the imports are colored red and Codecov reports that file has 84.47% coverage whereas `make cover` shows 89.8%.

I updated the commit message per your recommendation. I also went ahead and future-proofed the script by using `-coverpkg` only for packages in the `./pkg/...` tree.

xiaq:
Have you tried running `go test -coverprofile=coverage -coverpkg=./... ./...`?

> See this article among others for why: https://www.ory.sh/golang-go-code-coverage-accurate/.

This article claims that "Golang flag -coverprofile is incompatible with -coverpgk", but that doesn't seem to be correct? Its command is:

```
$ go test -covermode=atomic \
    -coverprofile=coverage.tmp \
    -coverpkg $(go list github.com/arekkas/accurate-test-coverage/...) \
    github.com/arekkas/accurate-test-coverage/...

cannot use test profile flag with multiple packages
```

The way it passes the value for the `-coverpkg` flag is both unnecessary and wrong...

krader1961:
> This article claims that "Golang flag -coverprofile is incompatible with -coverpgk", but that doesn't seem to be correct? 

The example is wrong, and the wording is imprecise, but the two flags are in fact incompatible when profiling every package at the same time. Run this on the master branch:

```
go test -covermode=set -coverprofile=cover ./...
go tool cover -func=cover
```

You should see something like 85.3% total coverage. Now do

```
go test -covermode=set -coverprofile=cover -coverpkg=./... ./...
go tool cover -func=cover
```

You'll see 84.8% for the total coverage. The decrease is due to the inclusion of *website/cmd/...* packages that have no coverage. More importantly is that some indirect coverage, already discussed above, is not included in the metrics. Resulting in the overall coverage being 2% less than you and I would agree is more representative (i.e., "correct") for this project.

The problem is that the coverage data isn't correctly aggregated when profiling more than one package when `-coverpkg=./...` is used. I don't know if that is a bug, feature, or simply considered not worth implementing correctly by `go test -coverprofile`. While working around that shortcoming requires a script to profile each package, one at a time, that isn't entirely a bad thing. For example, you'll notice the `excluded_modules` var near the top of the script. A script of this nature makes it easier, for example, to tweak the results to account for quirks of the project.

krader1961:
@xiaq, I'm curious what you think about writing a script like *tools/test-coverage.sh* in Elvish rather than bash. That, obviously, requires building a working `elvish` binary but that is already a prerequisite for successfully running unit tests. Eat our own dog food, and all that.

xiaq:
> You'll see 84.8% for the total coverage. The decrease is due to the inclusion of website/cmd/... packages that have no coverage.

It seems that can be fixed by restricting the tests as well as `-coverpkg` to `./pkg/...`; test coverage for `website/cmd` is not that important. This is what I observe:

```
~/go/src/github.com/elves/elvish> { go test -coverprofile=coverage.unit ./pkg/...; go tool cover -func coverage.unit } | tail -n1
total:                                                                  (statements)                    85.4%
~/go/src/github.com/elves/elvish> { go test -coverprofile=coverage.full -coverpkg=./pkg/... ./pkg/...; go tool cover -func coverage.full } | tail -n1
total:                                                                  (statements)                    87.5%
```

> More importantly is that some indirect coverage, already discussed above, is not included in the metrics.

> The problem is that the coverage data isn't correctly aggregated when profiling more than one package when -coverpkg=./... is used. I don't know if that is a bug, feature, or simply considered not worth implementing correctly by go test -coverprofile. 

I am really confused by these claims. Can we talk in terms of concrete problems - for example, if you run the commands I listed above and inspect `coverage.full`, what coverage should be included but is missing?

krader1961:
> Can we talk in terms of concrete problems

Sure, see issue #1187. The impetus for this change was the explicit test I added for the fix for issue #1186 (using float64 in prompts) was not reflected in the test coverage. And that was due to the current behavior of `go test -cover`, even if `-coverpkg` is used, when emitting coverage data for more than a single package.

In addition to the problem that caused me to create this change you'll see in this [comment](https://github.com/elves/elvish/issues/1187#issuecomment-745022186) that there are other functions, such as `pkg/sys/termios.go:20:@TermiosForFd`, which are only exercised by integration tests and therefore not shown as covered in the coverage data prior to this change.

xiaq:
> > Can we talk in terms of concrete problems
> 
> Sure, see issue #1187. The impetus for this change was the explicit test I added for the fix for issue #1186 (using float64 in prompts) was not reflected in the test coverage. And that was due to the current behavior of `go test -cover`, even if `-coverpkg` is used, when emitting coverage data for more than a single package.
> 
> In addition to the problem that caused me to create this change you'll see in this [comment](https://github.com/elves/elvish/issues/1187#issuecomment-745022186) that there are other functions, such as `pkg/sys/termios.go:20:@TermiosForFd`, which are only exercised by integration tests and therefore not shown as covered in the coverage data prior to this change.

I understand the original problem. Can you explain what is wrong with the coverage produced by `go test -coverprofile=coverage.full -coverpkg=./pkg/... ./pkg/...`?

krader1961:
> Can you explain what is wrong with the coverage produced by go test -coverprofile=coverage.full -coverpkg=./pkg/... ./pkg/...?

No, because that does what I expect and, modulo not excluding test packages, does what my far more complicated script does. I could swear I tested that command earlier this year and found it did not produce the correct coverage. Perhaps it didn't work correctly in Go 1.14 or, more likely, I made a mistake when I tested that command. Regardless, I've simplified the P.R. to remove the unnecessary script.

.appveyor.yml
```diff
@@ -4,11 +4,13 @@ environment:
   GOPATH: c:\gopath
   ELVISH_TEST_TIME_SCALE: 10
 install:
-  # Needed since the codecov uploader depends on a UNIX "find" command.
+  # The coverage script and codecov uploader depend on a POSIX shell and
+  # commands such as `find` so ensure the relevant Windows MSYS2 directory is
+  # in the external command search path.
```
xiaq, line 9: This comment no longer needs to be updated, revert this chunk.

Makefile
```diff
@@ -25,8 +25,9 @@ test:
 # Generate a basic test coverage report. This will open the report in your
 # browser. See also https://codecov.io/gh/elves/elvish/.
 cover:
-	go test -covermode=set -coverprofile=$@ ./...
-	go tool cover -html=$@
+	go test -coverprofile=cover -coverpkg=./pkg/... ./pkg/...
+	go tool cover -html=cover
+	go tool cover -func=cover | tail -1 | awk '{ print "Overall coverage:", $$NF }'
```
xiaq, line 30: Why change `$@` to `cover`?

.appveyor.yml
krader1961: I would argue my comment better reflects reality. This Appveyor config doesn't just depend on a `find` command. It explicitly runs the `bash` command to execute a bash script.

Makefile
krader1961: For consistency with the other coverage generators in *.cirrus.yml* and *.appveyor.yml*. This also makes it easier to copy/paste the statements without someone having to know what `$@` means in a makefile context.

xiaq: Sounds good.

.appveyor.yml
xiaq: That makes sense, but please remove the reference to the now non-existing "coverage script" from the comment.

krader1961: Done

xiaq (approved):
Will merge after CI success.

PR #1192, opened 2020-12-10
Ensure the daemon doesn't have fds open on the tty

### LINKED ISSUE #1191: first opened elvish shell hangs on exit until all others are closed
How to reproduce:

1. Set up a keybind to launch elvish. (in my case, I have `chsh -s /usr/local/bin/elvish` + the foot terminal on super+enter, though the former requires a relogin)
2. Launch two terminals, each running an elvish as a login shell.
3. log out of the first terminal (using `^d`). Observe that it will not exit until the second one is logged out as well.

This seems to be related to the first elvish shell launching `elvish -daemon` when ran (unless one is already running).
However, when you try to exit it, it will stay open until no further clients remain.

I tried to remedy this by launching the daemon (with the same arguments) separately, but it appears that the daemon self-terminates once it has no more clients.

krader1961:
The Elvish daemon should not inherit file descriptors open on the tty.
Make the daemon's stdin open on the null device and its stdout/stderr
open on a predictable file.

Fixes #1191

### DIFF
--- a/pkg/daemon/spawn.go
+++ b/pkg/daemon/spawn.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+
+	"github.com/elves/elvish/pkg/fsutil"
 )
 
 // SpawnConfig keeps configurations for spawning the daemon.
@@ -18,17 +20,17 @@ type SpawnConfig struct {
 	// SockPath is the path to the socket on which the daemon will serve
 	// requests.
 	SockPath string
-	// LogPathPrefix is used to derive the name of the log file by adding the
-	// pid.
-	LogPathPrefix string
+	// RunDir is the directory in which to place the daemon log file.
+	RunDir string
 }
 
-// Spawn spawns a daemon process in the background by invoking BinPath, passing
-// DbPath, SockPath and LogPathPrefix as command-line arguments after resolving
-// them to absolute paths. A suitable ProcAttr is chosen depending on the OS and
-// makes sure that the daemon is detached from the current terminal (so that it
-// is not affected by I/O or signals in the current terminal), and keeps running
-// after the current process quits.
+// Spawn spawns a daemon process in the background by invoking BinPath. Passing BinPath, DbPath and
+// SockPath as command-line arguments after resolving them to absolute paths. The daemon log file is
+// created in the secure run-time dir.
+//
+// A suitable ProcAttr is chosen depending on the OS and makes sure that the daemon is detached from
+// the current terminal (so that it is not affected by I/O or signals in the current terminal), and
+// keeps running after the current process quits.
 func Spawn(cfg *SpawnConfig) error {
 	binPath := cfg.BinPath
 	// Determine binPath.
@@ -58,7 +60,7 @@ func Spawn(cfg *SpawnConfig) error {
 	binPath = abs("BinPath", binPath)
 	dbPath := abs("DbPath", cfg.DbPath)
 	sockPath := abs("SockPath", cfg.SockPath)
-	logPathPrefix := abs("LogPathPrefix", cfg.LogPathPrefix)
+	runDir := abs("RunDir", cfg.RunDir)
 	if pathError != nil {
 		return pathError
 	}
@@ -69,11 +71,17 @@ func Spawn(cfg *SpawnConfig) error {
 		"-bin", binPath,
 		"-db", dbPath,
 		"-sock", sockPath,
-		"-logprefix", logPathPrefix,
 	}
 
-	// TODO Redirect daemon stdout and stderr
+	daemonStdout, err := fsutil.ClaimFile(runDir, "daemon-*.log")
+	if err != nil {
+		return err
+	}
+	defer daemonStdout.Close()
 
-	_, err := os.StartProcess(binPath, args, procAttrForSpawn())
+	procattrs := procAttrForSpawn(daemonStdout)
+	_, err = os.StartProcess(binPath, args, procattrs)
+	procattrs.Files[0].Close() // drop our reference to the /dev/null handle passed to the daemon
+
 	return err
 }
--- a/pkg/daemon/spawn_unix.go
+++ b/pkg/daemon/spawn_unix.go
@@ -7,11 +7,15 @@ import (
 	"syscall"
 )
 
-func procAttrForSpawn() *os.ProcAttr {
+func procAttrForSpawn(stdout *os.File) *os.ProcAttr {
+	// The daemon should not inherit a reference to the stdin file descriptor
+	// of the original shell. The daemon shouldn't read anything from stdin so
+	// use the null device.
+	devnull, _ := os.OpenFile(os.DevNull, os.O_RDONLY, 0)
 	return &os.ProcAttr{
-		Dir:   "/",        // cd to /
-		Env:   []string{}, // empty environment
-		Files: []*os.File{os.Stdin, os.Stdout, os.Stderr},
+		Dir:   "/",
+		Env:   []string{},
+		Files: []*os.File{devnull, stdout, stdout},
 		Sys: &syscall.SysProcAttr{
 			Setsid: true, // detach from current terminal
 		},
--- a/pkg/daemon/spawn_windows.go
+++ b/pkg/daemon/spawn_windows.go
@@ -15,11 +15,15 @@ const (
 	daemonCreationFlags = CREATE_BREAKAWAY_FROM_JOB | CREATE_NEW_PROCESS_GROUP | DETACHED_PROCESS
 )
 
-func procAttrForSpawn() *os.ProcAttr {
+func procAttrForSpawn(stdout *os.File) *os.ProcAttr {
+	// The daemon should not inherit a reference to the stdin file descriptor
+	// of the original shell. The daemon shouldn't read anything from stdin so
+	// use the null device.
+	devnull, _ := os.OpenFile(os.DevNull, os.O_RDONLY, 0)
 	return &os.ProcAttr{
 		Dir:   `C:\`,
 		Env:   []string{"SystemRoot=" + os.Getenv("SystemRoot")}, // SystemRoot is needed for net.Listen for some reason
-		Files: []*os.File{os.Stdin, os.Stdout, os.Stderr},
+		Files: []*os.File{devnull, stdout, stdout},
 		Sys:   &syscall.SysProcAttr{CreationFlags: daemonCreationFlags},
 	}
 }
--- a/pkg/fsutil/claim.go
+++ b/pkg/fsutil/claim.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 )
@@ -53,7 +54,7 @@ func ClaimFile(dir, pattern string) (*os.File, error) {
 	}
 
 	for i := max + 1; ; i++ {
-		name := prefix + strconv.Itoa(i) + suffix
+		name := filepath.Join(dir, prefix+strconv.Itoa(i)+suffix)
 		f, err := os.OpenFile(name, os.O_RDWR|os.O_CREATE, 0666)
 		if err == nil {
 			return f, nil
--- a/pkg/prog/prog.go
+++ b/pkg/prog/prog.go
@@ -12,7 +12,6 @@ import (
 	"io"
 	"os"
 	"runtime/pprof"
-	"strconv"
 
 	"github.com/elves/elvish/pkg/logutil"
 )
@@ -34,7 +33,7 @@ func SetShowDeprecations(b bool) func() {
 
 // Flags keeps command-line flags.
 type Flags struct {
-	Log, LogPrefix, CPUProfile string
+	Log, CPUProfile string
 
 	Help, Version, BuildInfo, JSON bool
 
@@ -54,8 +53,7 @@ func newFlagSet(stderr io.Writer, f *Flags) *flag.FlagSet {
 	fs.SetOutput(stderr)
 	fs.Usage = func() { usage(stderr, fs) }
 
-	fs.StringVar(&f.Log, "log", "", "a file to write debug log to")
-	fs.StringVar(&f.LogPrefix, "logprefix", "", "the prefix for the daemon log file")
+	fs.StringVar(&f.Log, "log", "", "a file to write debug log to except for the daemon")
 	fs.StringVar(&f.CPUProfile, "cpuprofile", "", "write cpu profile to file")
 
 	fs.BoolVar(&f.Help, "help", false, "show usage help and quit")
@@ -110,13 +108,15 @@ func Run(fds [3]*os.File, args []string, programs ...Program) int {
 		}
 	}
 
-	if f.Log != "" {
+	if f.Daemon {
+		// We expect our stdout file handle is open on a unique log file for the daemon to write its
+		// log messages. See daemon.Spawn() in pkg/daemon.
+		logutil.SetOutput(fds[1])
+	} else if f.Log != "" {
 		err = logutil.SetOutputFile(f.Log)
-	} else if f.LogPrefix != "" {
-		err = logutil.SetOutputFile(f.LogPrefix + strconv.Itoa(os.Getpid()))
-	}
-	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
+		if err != nil {
+			fmt.Fprintln(fds[2], err)
+		}
 	}
 
 	if f.Help {
--- a/pkg/shell/paths.go
+++ b/pkg/shell/paths.go
@@ -11,9 +11,8 @@ import (
 
 // Paths keeps all paths required for the Elvish runtime.
 type Paths struct {
-	RunDir          string
-	Sock            string
-	DaemonLogPrefix string
+	RunDir string
+	Sock   string
 
 	DataDir string
 	Db      string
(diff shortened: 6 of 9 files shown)


pkg/daemon/spawn.go
```diff
@@ -58,7 +59,7 @@ func Spawn(cfg *SpawnConfig) error {
 	binPath = abs("BinPath", binPath)
 	dbPath := abs("DbPath", cfg.DbPath)
 	sockPath := abs("SockPath", cfg.SockPath)
-	logPathPrefix := abs("LogPathPrefix", cfg.LogPathPrefix)
+	logPath := abs("LogPathPrefix", cfg.LogPathPrefix+strconv.Itoa(os.Getpid()))
```
xiaq, line 62: Hmm, IIUC this PR moves the opening of the log file from the daemon process to its parent. Since the filename is now determined in the parent, it no longer reflects the PID of the daemon process itself, but that of the parent.

Why is this necessary, and is it related to the main purpose of this PR (ensure the daemon doesn't have fds open on the tty)?

krader1961: Yes, this does what you described. The reason is we don't want to risk losing any messages written to stdout/err before the daemon setup code can open a log file and redirect its os.Stdout/err to that log file. This is also marginally simpler than doing it in the daemon setup code. I'm also convinced that sooner, rather than later, the daemon will be eliminated. Too, this allows eliminating the `-logprefix` flag which shouldn't, and isn't mean to be, used by a person. So the fix employed by this P.R. makes more sense.

That the PID on the file doesn't match that of the daemon isn't really important. Note that the daemon explicitly records its PID in the first log message it writes. Regardless of which process names the file the PID could still match a prior execution and thus a log file that already exists.

xiaq: > The reason is we don't want to risk losing any messages written to stdout/err before the daemon setup code can open a log file and redirect its os.Stdout/err to that log file.

That is not correct. The stdout and stderr of the daemon are not redirected to the log file specified by `-log` or `-logprefix`; those flags affect the destination of the "loggers" defined in https://github.com/elves/elvish/tree/master/pkg/logutil. There is actually a TODO saying that the stdout and stderr of the daemon probably should be redirected, though :): https://github.com/elves/elvish/blob/3615988408b7105c92096a784d915ca2568a83cb/pkg/daemon/spawn.go#L75

> That the PID on the file doesn't match that of the daemon isn't really important.

I agree and disagree. I agree that it's not important to have the PID as part of the filename since the first message is reporting the PID of the daemon. However, having another mysterious number that looks like a PID but is **not** the PID of the daemon can be very confusing.

It's better to just use a serial number; so instead of `daemon.log-23178` or `daemon.log-12865` you get `daemon.log-1`, `daemon.log-2`, etc.. I've actually implemented a function to support this idea, but so far the function is unused:

https://github.com/elves/elvish/blob/3615988408b7105c92096a784d915ca2568a83cb/pkg/fsutil/claim.go#L15

For the sake of this PR, I'd prefer that you leave this part out, since it's not necessary for the more important change of ensuring the daemon doesn't have FDs open on the TTY.

krader1961: > That is not correct. The stdout and stderr of the daemon are not redirected to the log file specified by -log or -logprefix

It is true with this change since I modified the code to pass os.Stdout to the logger. See the block at line 111 in *pkg/prog/prog.go*. I overlooked the TODO. That comment should also be removed.

krader1961: Hold on, I did remove the TODO about redirecting stdout/stderr for the daemon.

xiaq: Please address this feedback:

> That the PID on the file doesn't match that of the daemon isn't really important.

I agree and disagree. I agree that it's not important to have the PID as part of the filename since the first message is reporting the PID of the daemon. However, having another mysterious number that looks like a PID but is **not** the PID of the daemon can be very confusing.

It's better to just use a serial number; so instead of `daemon.log-23178` or `daemon.log-12865` you get `daemon.log-1`, `daemon.log-2`, etc.. I've actually implemented a function to support this idea, but so far the function is unused:

https://github.com/elves/elvish/blob/3615988408b7105c92096a784d915ca2568a83cb/pkg/fsutil/claim.go#L15

krader1961: Okay, but you wrote in your previous comment immediately after talking about possibly using `ClaimFile`:

> For the sake of this PR, I'd prefer that you leave this part out, since it's not necessary for the more important change of ensuring the daemon doesn't have FDs open on the TTY.

Also, these are *temporary* files that are removed on each reboot and it should be rare for users to have more than one of these daemon log files. But since `ClaimFile` already exists I have no objection to using it. Updated PR in a few minutes....

krader1961: FYI, I found a bug in `ClaimFile`. It fails to create the file in the directory that it scanned for existing files that match the pattern. Instead it creates the file in the CWD. I've fixed that but didn't bother to update its unit test.

pkg/fsutil/claim.go
```diff
@@ -53,7 +54,7 @@ func ClaimFile(dir, pattern string) (*os.File, error) {
 	}
 
 	for i := max + 1; ; i++ {
-		name := prefix + strconv.Itoa(i) + suffix
+		name := filepath.Join(dir, prefix+strconv.Itoa(i)+suffix)
```
xiaq, line 57: Thanks for fixing the bug, but please do add a unit test.

pkg/prog/prog.go
```diff
@@ -55,7 +54,6 @@ func newFlagSet(stderr io.Writer, f *Flags) *flag.FlagSet {
 	fs.Usage = func() { usage(stderr, fs) }
 
 	fs.StringVar(&f.Log, "log", "", "a file to write debug log to")
```
xiaq, line 56: Update the description for `-log` that it is ignored when `-daemon` is specified, since the daemon always writes logs to stdout.

pkg/prog/prog.go
```diff
@@ -110,13 +108,15 @@ func Run(fds [3]*os.File, args []string, programs ...Program) int {
 		}
 	}
 
-	if f.Log != "" {
+	if f.Daemon {
+		// See daemon.Spawn(). We expect our stdout file handle is open on a
+		// unique log file for the daemon to write its log messages.
```
xiaq, line 113: I find it more readable to put references at the end; it'd also be better to include the package name too. So I'd write this as:

> We expect our stdout file handle is open on a unique log file for the daemon to write its log messages. See daemon.Spawn() in pkg/daemon.

pkg/shell/paths.go
```diff
@@ -28,8 +27,8 @@ func MakePaths(stderr io.Writer, overrides Paths) Paths {
 	p := overrides
 	setDir(&p.RunDir, "secure run directory", getSecureRunDir, stderr)
 	if p.RunDir != "" {
+		setChild(&p.RunDir, p.RunDir, "")
```
xiaq, line 30: Hmm?

xiaq (commented):
Overall LGTM now, some feedback for stylistic matters.

pkg/daemon/spawn.go
xiaq: > Okay, but you wrote in your previous comment immediately after talking about possibly using ClaimFile:

Oh, I originally thought that this PR contains two independent changes, (1) making sure that the daemon doesn't have FDs on the TTY and (2) redirecting the stdout and stderr of the daemon, so I was asking you to make this PR only do (1) but not (2) - by "that" I meant part (2). But later I realized that those two things are the same thing.

pkg/daemon/spawn.go
```diff
@@ -18,13 +20,12 @@ type SpawnConfig struct {
...
-	// LogPathPrefix is used to derive the name of the log file by adding the
-	// pid.
-	LogPathPrefix string
+	// RunDir is the directory in which to place the daemon log file.
+	RunDir string
 }
 
 // Spawn spawns a daemon process in the background by invoking BinPath, passing
-// DbPath, SockPath and LogPathPrefix as command-line arguments after resolving
+// DbPath, SockPath and LogPath as command-line arguments after resolving
```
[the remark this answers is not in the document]
krader1961: Done

pkg/daemon/spawn_unix.go
```diff
@@ -7,11 +7,15 @@ import (
 	"syscall"
 )
 
-func procAttrForSpawn() *os.ProcAttr {
+func procAttrForSpawn(stdout *os.File) *os.ProcAttr {
+	// The daemon should not inherit a reference to the stdin file descriptor
+	// of the original shell. The daemon shouldn't read anything from stdin so
+	// use the null device.
+	devnull, _ := os.OpenFile("/dev/null", os.O_RDONLY, 0)
```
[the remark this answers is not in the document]
krader1961: **No. Don't ever do that for stdin/stdout/stderr.** I've debugged two vexing problems in the past due to programs that launched other programs with one or more of those file descriptors closed. UNIX programs read from, or write to, those file descriptors not just in the "main" program but also imported packages (libraries). Every program assumes they are open on something. If they aren't the first `open()` call the program executes will allocate one of those file descriptors. Oops! You now have other parts of the program that use the stdin/out/err file descriptors reading/writing a file not meant to be used as stdin/out/err.

It is extremely unlikely a different path will ever be needed for a Go module predicated on the meta GOOS value "unix".

pkg/daemon/spawn_unix.go
```diff
@@ -7,11 +7,15 @@ import (
...
 	return &os.ProcAttr{
-		Dir:   "/",        // cd to /
+		Dir:   "/",        // make the CWD when the daemon starts the root directory
```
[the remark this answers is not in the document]
krader1961: Fair enough, I have removed it; although I think it has some, slight, value to a reader of the code not intimately familiar with the semantics of `os.ProcAttr` and `os.StartProcess`.  Your observation also applies to the comment of the `Env` struct member so I've removed it as well.

pkg/daemon/spawn_windows.go
```diff
@@ -15,11 +15,18 @@ const (
...
+	// use the null device.
+	devnull, _ := os.OpenFile("NUL:", os.O_RDONLY, 0)
+	env := []string{
+		"SystemRoot=" + os.Getenv("SystemRoot"), // this is needed for net.Listen for some reason
+	}
 	return &os.ProcAttr{
-		Dir:   `C:\`,
-		Env:   []string{"SystemRoot=" + os.Getenv("SystemRoot")}, // SystemRoot is needed for net.Listen for some reason
-		Files: []*os.File{os.Stdin, os.Stdout, os.Stderr},
+		Dir:   `C:\`, // make the CWD when the daemon starts the root directory
```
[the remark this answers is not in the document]
krader1961: Done

pkg/fsutil/claim.go
krader1961: Isn't that your responsibility since I don't think modifying this change to use `ClaimFile` is even necessary or particularly useful? Note that I wouldn't have even noticed this problem if I didn't perform due diligence to see what happens if I launch an Elvish shell built with this change without an already accessible daemon.

pkg/daemon/spawn_windows.go
```diff
@@ -15,11 +15,18 @@ const (
...
 
-func procAttrForSpawn() *os.ProcAttr {
+func procAttrForSpawn(stdout *os.File) *os.ProcAttr {
+	// The daemon should not inherit a reference to the stdin file descriptor
+	// of the original shell. The daemon shouldn't read anything from stdin so
+	// use the null device.
+	devnull, _ := os.OpenFile("NUL:", os.O_RDONLY, 0)
+	env := []string{
+		"SystemRoot=" + os.Getenv("SystemRoot"), // this is needed for net.Listen for some reason
+	}
```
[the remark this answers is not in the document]
krader1961: I made that change because the literal is non-trivial (unlike in *spawn_unix.go*) and I was already touching the adjacent lines. But I don't feel strongly enough to argue for keeping this aspect of the P.R. so I've reverted it.

pkg/prog/prog.go
krader1961: Done. However, I think this is a silly nitpick since manually launching the daemon is dangerous, not officially supported, and hopefully irrelevant in the future when the daemon is eliminated.

pkg/prog/prog.go
krader1961: Okay. Done. However, I think this is nitpicking.

pkg/prog/prog.go
xiaq: > I think this is a silly nitpick since manually launching the daemon is dangerous, not officially supported

Your premise is true from a user's perspective, but not true from a developer's perspective. You may not like the daemon, but as long as it exists as a developer I prefer making it easy to debug it and launching it manually is a valuable technique to do that.

pkg/shell/paths.go
xiaq: Did you miss this comment? This line of code doesn't do anything useful and is confusing, although it doesn't seem to do any harm so I can fix this after merging this PR.

pkg/daemon/spawn_unix.go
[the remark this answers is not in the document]
xiaq: That's a good point, but it seems that it applies more to stdout and stderr, which libraries are likely to write to. I'd be surprised if a library reads from stdin. Nonetheless, it sounds like good practice to avoid closing any of the lowest 3 FDs.

pkg/daemon/spawn.go
```diff
@@ -69,11 +70,17 @@ func Spawn(cfg *SpawnConfig) error {
...
-	// TODO Redirect daemon stdout and stderr
+	daemonStdout, err := fsutil.ClaimFile(runDir, "daemon-*.log")
+	if err != nil {
+		return err
+	}
+	defer daemonStdout.Close()
+
+	procattrs := procAttrForSpawn(daemonStdout)
+	_, err = os.StartProcess(binPath, args, procattrs)
+	procattrs.Files[0].Close() // drop our reference to the /dev/null handle passed to the daemon
```
xiaq, line 83: Instead of letting `procAttrForSpawn` open `/dev/null` or `NUL` and relying on the implicit fact that `procattrs.Files[0]` is that file, it's better to open that file here and pass it as an argument to `procAttrForSpawn`; `os.DevNull` can be used as the filename.

pkg/fsutil/claim.go
xiaq: Well, you could argue that everything is my responsibility and I wouldn't disagree :)

It's good practice to add a regression test when you fix a bug, but if doing that demotivates you from working on this PR I certainly won't insist.

xiaq (commented):
You sound frustrated, and I am sorry about that.

I appreciate the work you have put into PRs and I know it can be quite frustrating to wake up to a lot of comments and find that most of them are "nitpicking" ones.

But let me clarify my motivations too. Elvish is more or less my personal project, I take ownership over every line of code I merge, so I need make myself comfortable with every line of diff. I will continue point out all the places I am not comfortable with - and yes, I **am** picky when it comes to comments and documentation often in my own way.

**However**, I try to not block PRs purely on stylistic issues, since I can always easily make those changes in a followup commit. When I say *overall LGTM*, it means that only minor or stylistic issues remain, and it's OK for you to not act upon the remaining feedback, and leave it to me to carry out the changes; but I'll make comments nonetheless, as a note to myself.

I've done that proactively for some past PRs when review has dragged on for too long. For PRs where I don't do that proactively but you'd like me to, you can also explicitly say that you wish to have the PR merged and you'll leave those smaller changes to me.

Also, it seems that you have responded to most my comments but haven't pushed the code that addresses them? If you prefer to let me do the remaining stylistic changes, I can certainly do it in the PR's current form (I've already said *overall LGTM*), but if you already have a local commit it's better for you to push first.

xiaq:
In retrospect, when I said "overall LGTM", I should have explicitly offered you the option of having the PR merged as is, and made it your choice whether you want to further polish the PR or leave the polishing to me. I'll do that in future reviews.

pkg/shell/paths.go
krader1961: Yes, that particular change is wrong. It's left over from earlier changes to this code when I was figuring out how to adapt the code to use `ClaimFile`.

krader1961:
> You sound frustrated, and I am sorry about that.

I have no objection to being uncompromising (aka being a hard-ass) on proposed changes. I'm the same way when reviewing changes at $day-job. What pushed me over the edge was asking me to add (or fix) a unit test for `ClaimFile` since I

*) did not write that code,
*) spent an hour adapting my change to use `ClaimFile` despite my belief using it in the context of this change is pointless, and
*) the incorrect behavior of `ClaimFile` cost me another hour because I initially assumed my use of that function was incorrect, before I realized that function had a bug.

P.S., I have, partially, addressed your other point about using `os.DevNull` rather than hard-coded strings. I say "partially" because I have replaced the literal strings with that symbol but am not convinced moving those file opens to the caller is preferable. I want to sleep on that.

krader1961:
@xiaq, This seems like a good place to comment on the fact that you seemingly never do a squash-merge of third-party changes. That caused me to switch from pushing deltas in response to your feedback and instead squashing fixups locally and force pushing so there is a single commit to be merged. I think pushing delta commits makes resolving feedback easier but perhaps you prefer what I've been doing, which is fine, but I have wondered if my local squash commits and forced pushes is counterproductive to getting changes merged sooner, rather than later.

xiaq:
> @xiaq, This seems like a good place to comment on the fact that you seemingly never do a squash-merge of third-party changes. That caused me to switch from pushing deltas in response to your feedback and instead squashing fixups locally and force pushing so there is a single commit to be merged. I think pushing delta commits makes resolving feedback easier but perhaps you prefer what I've been doing, which is fine, but I have wondered if my local squash commits and forced pushes is counterproductive to getting changes merged sooner, rather than later.

https://github.com/elves/elvish/pull/949#issuecomment-605340410

xiaq:
I guess this can be merged now.

Issue #1065, opened 2020-07-03
Too many ways to output objects: `put`, `echo`, `print`, `repr`, `pprint`, and now `show`


krader1961:
Commit e9b40358 moved `exc:show` to `builtin:show` as part of overhauling how exceptions are manifested. However, the `show` command only handles exception objects and currently results in a panic if handed a `$ok` object. At present the sole purpose of `builtin:show` is to pretty-print an exception object. It seems to me it would be better to augment the `pprint` builtin to do the same thing when handed an exception object rather than introduce a new builtin. Do we really need `show` in addition to `put`, `echo`, `print`, `repr` and `pprint`?

krader1961:
In PR #1063 @xiaq pointed out that `show` colorizes its output. Also, `pprint` is just a variant `repr` that uses indentation for nested values. And, of course, `print` is just a variant of `echo` that doesn't include a terminating newline.

FWIW, my preference is to eliminate `pprint` and add a `&pretty` option to `repr`. Similarly, eliminate `print` and add a `&no-newline` (or similar) option to `echo`. It seems to me neither `pprint` or `print` are used frequently enough to justify their existence independent of their counterparts and it would be cognitively simpler to replace [`put`, `echo`, `print`, `repr`, `pprint`] with [`put`, `echo`, `repr`].

Note that the `show` command actually does more than colorize its output. In the case of exception objects it includes context that is missing from the output of `echo` or `repr`:

```
> fn wtf { e:wtf }
> x = ?(wtf)
> echo $x
[&reason=<unknown exec: "wtf": executable file not found in $PATH>]
> repr $x
[&reason=<unknown exec: "wtf": executable file not found in $PATH>]
> show $x
Exception: exec: "wtf": executable file not found in $PATH
Traceback:
  [tty 7], line 1:
    fn wtf { e:wtf }
  [tty 8], line 1:
    x = ?(wtf)
```

This raises a couple of questions. Should `repr` include the traceback context? Should `repr` have a `&colorize` option in addition to a `&pretty` option? Should pretty-printing exceptions use the `show` representation rather than treating the exception object as a pseudo-map? Note the current `repr` output of an exception object cannot be used to create an exception object. Which makes the map like output for an exception object not particularly useful.

xiaq:
`repr` should include the traceback context, it's just that I haven't found a satisfying way to expose it.

Each entry of the traceback is a `diag.Context`, which contains the full code of the originating source:

[...]

PR #1200, opened 2020-12-22
Implement a `printf` builtin

### LINKED ISSUE #1132: Add something like a `printf` builtin
I was asked to publish my `cmd-duration` module to make it easy to include a human friendly command duration in the left or right prompt. That caused me to realize my current implementation needed some improvements. One of which was to replace whole units with fractional units to one decimal point. That is, rather than "1 m" (for one minute) it might display "1.3 m" (for 1.3 minutes). That, however, requires using the external `printf` command for reliable results. Notwithstanding the generally excellent automatic conversion of float64 values to strings. I've noticed this deficiency before in my use of Elvish but generally worked around it -- often by calling the external `printf`. I've concluded this capability should be a builtin.

We almost certainly do not want to slavishly implement the [POSIX printf command](https://pubs.opengroup.org/onlinepubs/007908799/xcu/printf.html). Whether such a builtin should be named `builtin:printf`, `builtin:sprint`, or `str:printf` is an open question. Too, whether we want to simply expose the Go stdlib `fmt.Sprint` function or provide a more constrained wrapper around it is TBD.

krader1961:
There is more work to be done but this should address 99.9999% of the
expected uses of a `printf` builtin. Some corner cases which don't have
a POSIX analog, such as support for Elvish bool types will be implemented
in future changes.

Resolves #1132

### DIFF
--- a/pkg/eval/builtin_fn_io.go
+++ b/pkg/eval/builtin_fn_io.go
@@ -25,6 +25,7 @@ func init() {
 
 		// Bytes output
 		"print":  print,
+		"printf": printf,
 		"echo":   echo,
 		"pprint": pprint,
 		"repr":   repr,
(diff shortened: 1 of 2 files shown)


krader1961:
FWIW, I recognize the irony of my introducing yet another command to output data in light of issue #1065 that I opened five months ago. Nonetheless, since we already have numerous commands for outputting variables (even if implicit via command capture), and for easy discoverability, I opted for a new builtin command. I admit my position in issue #1065 has softened. In some of these cases (e.g., `echo`, `printf`) using a distinct builtin, rather than folding their functionality into a common builtin like `print`, is useful to make it harder for the user to inadvertently invoke an external command of the same name but with different behavior.

krader1961:
The force push above was solely to make the documentation of this change consistent with the existing documentation. Specifically, using `\\ **Note**:` rather than the form in my original change for this pull-request.

pkg/eval/builtin_fn_io.go
```diff
@@ -194,6 +195,180 @@ func print(fm *Frame, opts printOpts, args ...interface{}) {
...
+//
+// See Go's [`fmt`](https://golang.org/pkg/fmt/#hdr-Printing) package for
+// details about the formatting verbs and the various flags that modify the
+// default behavior; e.g., left versus right justification. We explicitly
+// handle only strings and float64 data types. Printing other types is likely
+// to result in unexpected output (but not an exception). You can use the
+// numeric formatting verbs (e.g., `%d`, `%f`) with either a float64 or a
+// string that can be converted to a number. If you use a float64 value to
+// the `%s` verb it will first be converted to a string using the rules for
+// [numbers](language.html#number]).
```
xiaq, line 218: A more systematic way to think of the approach here is that all the verbs first tries to convert the corresponding argument to the relevant type. Namely:

* `%s` converts the argument to a string
* `%d` and other integer verbs convert the argument to an int, if possible
* `%f` and other float verbs convert the argument to a float, if possible

pkg/eval/builtin_fn_io.go
```diff
@@ -194,6 +195,180 @@ func print(fm *Frame, opts printOpts, args ...interface{}) {
...
+// the POSIX specification. The two have many similarities but are not
+// identical. The second is that the number of values must match the number of
+// formatting verbs. Excess values do not result in the format being evaluated
+// additional times until all values are consumed as happens with the POSIX
+// command of the same name. You must explicitly split the value list and
+// invoke this builtin for each block of values. The third is that the Elvish
+// `printf` does not recognize backslash sequences such as `\n`. You must use
+// [double-quoted Elvish strings](language.html#double-quoted-string) if you
+// want such backslash sequences to be recognized -- just as you would with a
+// string passed to any other Elvish builtin.
```
xiaq, line 260: Formatting the differences as a bullet list would be clearer.

pkg/eval/builtin_fn_io.go
```diff
@@ -194,6 +195,180 @@ func print(fm *Frame, opts printOpts, args ...interface{}) {
...
+//
+// @cf print echo pprint repr
+
+type formatterString struct {
+	str string
+}
+type formatterFloat64 struct {
+	num float64
+}
```
xiaq, line 269: Once you think of the verbs as descriptions of how to *convert* the argument, it should become clear that it's not necessary to define different formatter for different types.

The branch for `%s` and `%v` can both use `vals.ToString`, the branch for `%d`, `%f` and other numerical verbs can all use `vals.ScanToGo` with a `int` or `float64` destination.

pkg/eval/builtin_fn_io.go
```diff
@@ -194,6 +195,180 @@ func print(fm *Frame, opts printOpts, args ...interface{}) {
...
+
+func printf(fm *Frame, template string, args ...interface{}) {
+	out := fm.OutputFile()
+	aliasedArgs := make([]interface{}, len(args))
```
xiaq, line 273: This slice stores wrappers, not aliases; `wrappedArgs` would be more accurate.

pkg/eval/builtin_fn_io.go
```diff
@@ -194,6 +195,180 @@ func print(fm *Frame, opts printOpts, args ...interface{}) {
...
+		}
+	}
+	s := fmt.Sprintf(template, aliasedArgs...)
+	out.WriteString(s)
+}
+
+// Convert a formatting verb state to a string that is logically equivalent to
+// the original formatting verb.
+func buildFmtVerb(state fmt.State, v rune) string {
+	var flags = []rune{'%'}
```
xiaq, line 300: This function uses a variety of ways to build up the string (`append`, `fmt.Sprintf` and `+`); use a [strings.Builder](https://godoc.org/strings#Builder) instead.

xiaq (approved):
LGTM overall. Guess I'll err on the side of merging early and just go ahead and merge this.

PR #1201, opened 2020-12-24
Trivial documentation consistency change

krader1961:
I noticed that most places in the documentation that emit a "note"
about Elvish behavior use `**Note**:`. There were two places that use
all uppercase. Make those consistent with the other uses. Also clarify
the note associated with the `not` builtin.

### DIFF
--- a/pkg/eval/builtin_fn_pred.go
+++ b/pkg/eval/builtin_fn_pred.go
@@ -62,7 +62,9 @@ func init() {
 // ▶ $true
 // ```
 //
-// **NOTE**: `and` and `or` are implemented as special commands.
+// **Note**: `not` is a regular command, and thus can be overriden by a
+// function definition (though you shouldn't do so), while `and` and `or` are
+// implemented as [special commands](language.html#special-commands).
 //
 // @cf bool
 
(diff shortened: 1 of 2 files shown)


pkg/eval/builtin_fn_pred.go
```diff
@@ -62,7 +62,9 @@ func init() {
 // ▶ $true
 // ```
 //
-// **NOTE**: `and` and `or` are implemented as special commands.
+// **Note**: `not` is a regular command, and thus can be overriden by a
+// function definition (though you shouldn't do so), while `and` and `or` are
+// implemented as [special commands](language.html#special-commands).
```
xiaq, line 67: > and thus can be overriden by a function definition

While this is true and an interesting piece of information, the rationale to mention that `not` is a normal command while `and` and `or` are special commands is certainly **not** to remind the reader that they could technically override `not`. Rather, the rationale is to remind the reader that `and` and `or` actually have special evaluation rules to support short-circuiting, while `not` does not need special evaluation rules.

xiaq (approved):
I'll merge this and change the comment for `not` in a followup commit.

PR #1210, opened 2021-01-07
Fix handling of unexpected errors

### LINKED ISSUE #1208: Panic on call stack (maybe?) overflow
As a result of #1206, I was playing with a test factorial function, and it seems that Elvish panics when a math overflow is reached:
```
[~]|> fn fact [n]{ if (== $n 0) { put 1 } else { put (* $n (fact (- $n 1))) } }
[~]|> fact 50
▶ (float64 3.0414093201713376e+64)
[~]|> fact 60
panic: interface conversion: error is *os.SyscallError, not *eval.Exception

goroutine 24161 [running]:
github.com/elves/elvish/pkg/eval.(*pipelineOp).exec.func1(0x152a5a0, 0xc000dca210, 0xc000a44e70, 0xc00057b540, 0xc001030590, 0x203000)
	github.com/elves/elvish/pkg/eval/compile_effect.go:153 +0x152
created by github.com/elves/elvish/pkg/eval.(*pipelineOp).exec
	github.com/elves/elvish/pkg/eval/compile_effect.go:149 +0x225
Exception: elvish exited with 2
```

krader1961:
Doing something like the following is likely to result in too many open
files (assuming `ulimit -n` == 256) resulting in a panic:

    > fn fact [n]{ if (== $n 0) { put 1 } else { put (* $n (fact (- $n 1))) } }
    > fact 60
    panic: interface conversion: error is *os.SyscallError, not
    *eval.Exception

    goroutine 24161 [running]:
    github.com/elves/elvish/pkg/eval.(*pipelineOp).exec.func1(0x152a5a0,
    0xc000dca210, 0xc000a44e70, 0xc00057b540, 0xc001030590, 0x203000)
        github.com/elves/elvish/pkg/eval/compile_effect.go:153 +0x152
    created by github.com/elves/elvish/pkg/eval.(*pipelineOp).exec
        github.com/elves/elvish/pkg/eval/compile_effect.go:149 +0x225
    Exception: elvish exited with 2

Fixes #1208

### DIFF
--- a/pkg/eval/compile_effect.go
+++ b/pkg/eval/compile_effect.go
@@ -139,7 +139,12 @@ func (op *pipelineOp) exec(fm *Frame) error {
 			err := thisOp.exec(newFm)
 			newFm.Close()
 			if err != nil {
-				*thisError = err.(*Exception)
+				switch err := err.(type) {
+				case *Exception:
+					*thisError = err
+				default:
+					*thisError = &Exception{Reason: err, StackTrace: nil}
+				}
 			}
 			wg.Done()
 			if hasChanInput {


pkg/eval/compile_effect.go
```diff
@@ -139,7 +139,12 @@ func (op *pipelineOp) exec(fm *Frame) error {
 			err := thisOp.exec(newFm)
 			newFm.Close()
 			if err != nil {
-				*thisError = err.(*Exception)
+				switch err := err.(type) {
+				case *Exception:
+					*thisError = err
+				default:
+					*thisError = &Exception{Reason: err, StackTrace: nil}
```
xiaq, line 146: This should use `StackTrace: fm.traceback`

xiaq (approved):
The deeper problem here is that the `*Exception` type assertion shouldn't have been necessary in the first place. The `error` returned by `effectOp.exec` is almost always `*Exception`, but because it is not typed as such there are a few case that have fallen through the crack. A fix of the root cause is to change the return type of `effectOp.exec` to `*Exception` and fix the few places that are not returning `*Exception` to do so.

Nonetheless I'll merge this PR as it at least fix the crash.

Issue #1220, opened 2021-01-14
"slice bounds out of range" in glob.matchElement


mrlzh:
Reappearance：
Decoding tCsrMDMyM9X4pf7r5w== with Base64 as filename and creating a file.

Executing the 'ls *.a' command causes a crash
```
panic: runtime error: slice bounds out of range [14:13]

goroutine 170 [running]:
github.com/elves/elvish/pkg/glob.matchElement(0xc000393960, 0x2, 0x2, 0xc0003bf598, 0xd, 0x0)
        /Users/testingliu/Desktop/elvish-0.14.1/pkg/glob/glob.go:238 +0x716
```

xiaq:
The issue is now fixed, but I thought I'd share some interesting facts I found when trying to test this:
* macOS will not create a file whose name has invalid UTF-8 sequences.
* Windows pretends to, but substitutes the invalid sequence with U+FFFD.
* Linux seems to be the only OS that would happily create a file whose name contains anything but NUL.

PR #1221, opened 2021-01-14
fix https://github.com/elves/elvish/issues/1220

mrlzh:
```
if j>len(name){
	break
}
```
Check if j is greater than name's length

### DIFF
--- a/pkg/glob/glob.go
+++ b/pkg/glob/glob.go
@@ -228,6 +228,9 @@ segs:
 				if !startingStar.Match(r) {
 					break
 				}
+				if j > len(name) {
+					break
+				}
 				ok, rest := matchFixedLength(chunk, name[j:])
 				if ok && (rest == "" || len(segs) > 0) {
 					name = rest


pkg/glob/glob.go
```diff
@@ -228,6 +228,9 @@ segs:
 				if !startingStar.Match(r) {
 					break
 				}
+				if j > len(name) {
```
krader1961, line 231: Either this test should be moved above the prior `if !startingStar...` block or the `j := i + len(string(r))` statement should be moved to immediately prior to this line. It is unclear if the calculation of `j` should affect whether or not `startingStar.Match(r)` is called. I don't doubt this change fixes the panic but in its current form it introduces more ambiguity about the logic.

xiaq:
This PR patches up the panic but it doesn't address the underlying correctness issue.

The underlying issue is that the code here assumes that

```go
for i, r := range s { ... }
```

Is always a lossless way to split a string, and all the `r`s, when concatenated together, will form `s`. This is only true if `s` is a valid UTF-8 string; if there is anything that can't be decoded as UTF-8, the iteration will advance by one byte with `r = 0xFFFD` (whose length is not 1 when converted to a string, which explains the panic). OS paths are not guaranteed to be valid UTF-8 strings, and the reproduction in #1220 is essentially constructing a non-UTF-8 string.

The fix should be changing the `glob` package to treat strings as byte sequences instead of rune sequences.

krader1961:
@mrlzh, See also the description of the [two ways to iterate over a string](https://yourbasic.org/golang/for-loop-range-array-slice-map-channel/#string-iteration-runes-or-bytes). @xiaq is correct that Elvish cannot assume path names are valid UTF-8 strings. But since the glob/wildcard characters are single byte ASCII characters that doesn't matter. It is sufficient to process the path name one byte at at time. If the byte is a wildcard (e.g., `*`) it is treated specially; otherwise, the byte is a literal that has to match the corresponding byte in any expansion.

xiaq:
The bug is now fixed in e061b657cbd1621f420d5f85d6626d4bea7864bb, closing.

PR #1224, opened 2021-01-17
Fix typo in description of path:is-regular

krader1961:


krader1961:
The fix for the typo should be obvious. I decided to add a note about symlink handling by `is-dir` and `is-regular` because I think that is a subtlety that warrants being documented. I'm not happy about the duplicated verbiage. When additional `is-...` commands are introduced we'll probably want to refactor that text to avoid the duplication. Perhaps by introducing a section near the top of the module documentation. But for now I think the duplication is acceptable.

xiaq:
PR LGTM, but seems that I did break coverage reports for PRs...

PR #1229, opened 2021-02-02
Debugging support (e.g., like `bash -x`) and at the Go level

krader1961:
At this stage I want people to focus on the general approach and the output -- not the details of the implementation. If there aren't strong objections to the general approach I'll create individual PRs for each stage of this evolution in functionality to make the code review process more efficient.

If you've ever wanted the Elvish equivalent of `bash -x` you'll want to try this pull-request. Pick one of your Elvish scripts and run `elvish -trace=cmd /path/to/script.elv`. You can redirect the output to a file or another terminal; e.g., `elvish -trace=cmd,file=/dev/ttys006 /path/to/script.elv`. If you want to trace all the Elvish code your interactive shell runs to produce your prompt and implement other behaviors (e.g., the line editor) run the same command without a `/path/to/script.elv` argument. You'll definitely want to redirect the trace output to a different terminal in that case.

See the updated *README.md* at https://github.com/krader1961/elvish/tree/debug#running-and-debugging-elvish for more details about the new `-trace` CLI option.


krader1961:
The forced push updates fixed unit test failures due to differences in the build environment on my system versus the CI environments.

xiaq:
Thank you. I've tried it locally. The overall approach and the output formats are quite reasonable; in particular I like how detailed the output is. So this PR is a good first step, but as you said it should be broken into several smaller ones.

I'll give some high-level feedback for the implementation too, although before that I'd discuss the design a bit.

---

The most flexible design for a tracing mechanism would be to implement this as a builtin `trace` function, which takes a lambda containing the code to trace, and an optional callback that acts on each trace event:

```
trace &on-trace $lambda
```

When the `on-trace` callback is not given, the default behaviour is just to write out each trace event, like how `-trace` now behaves.

This function is more flexible than a CLI flag. The latter is inherently global, and only supports tracing a whole file, or not at all. With a function, users can trace code at the interactive REPL, trace part of the code, and so on. If the user wishes to trace a whole file, they can simply do

```
elvish -c 'trace { eval (slurp < foo.elv) }'
```

Of course, a `-trace` or `-x` CLI flag can still be provided for the most common use case of tracing a whole file. But it would be a thin wrapper over the `trace` function.

To be clear, I am not requiring you to rewrite the implementation to a `trace` function; implementing the trace mechanism as a global CLI flag is a good stepping stone.

But still, it's useful to keep in mind how the implementation should evolve in future, since it has implications for how to best implement things now. In particular, with a `trace` function, the state of tracing should not be global, either; it should be a local state within each `Frame`.

---

Now now high-level feedback for the implementation.

* Having a DSL in the argument of `-trace` feels a bit over-engineered. The same amount of customisability can be achieved simply by defining a few more flags. For example, the format of the timestamp can be another flag, `-trace-timestamp`.

* I appreciate that you're trying to make the `logutil` functionality consistent with the new tracing mechanism, but the consistency here is somewhat forced. The existing `logutil` is for Elvish developers debugging Elvish itself; the new tracing functionality is for Elvish users debugging Elvish code. So the use cases are distinct.

  The implementation also diverges. The fact that you defined a separate `fmPrintln` function for tracing is also telling of the underlying inconsistency. Tracing needs to show traceback in Elvish sources. Internal logging has no associated Elvish traceback; it seems you are actually printing traceback in the Go source, but that's not actually as important as the Elvish traceback in tracing. Usually the content of the log message is enough to identify the place that logs it.

  In future, when tracing is implemented as a `trace` function, the implementation will diverge more. As I described above, a `trace` function requires moving the tracing configuration from global states to `Frame`. On the other hand, internal logging will remain a global mechanism since there is no `Frame` tied to those internal logs.

  It's more productive to keep internal logging (`logutil`) and tracing separate.

Please let me know what you think.

xiaq:
Some more feedback:

I appreciate that you are adding documentation on the Elvish command itself, but there is a better place for it to live: in the output of `elvish -help`. In fact, with this PR, most of the flags have duplicate documentation in both README and in https://github.com/elves/elvish/blob/master/pkg/prog/prog.go. It's better to put all the documentation in the code; the output of `elvish -help` can be exported to a web page on https://elv.sh (like how elvdoc in code are exported to web pages). This is how Go's command documentation works too (https://golang.org/cmd/go/ has the same content as `go help`).

krader1961:
I disagree that this should be implemented in terms of a `trace` builtin. It would only be useful for the REPL case and would be cumbersome to use for the other use cases. It also begs the question of how you would use such a command to trace all the hidden activity that happens in an interactive shell; e.g., to render the prompt. Having said that, I would like there to be a way to trace just the commands run as a result of the user typing a statement at a REPL prompt; i.e., exclude exclude activity such as rendering the prompt.

> Having a DSL in the argument of -trace feels a bit over-engineered.

Not really. Adding additional flags increases ambiguity. Not least because you end up with tens of flags. All of which implicitly enable tracing. It's far clearer to 

> Usually the content of the log message is enough to identify the place that logs it.

Yes, but for someone not intimately familiar with the code (e.g., me) it can be extremely useful to understand the sequence of function calls that lead to a particular point in the code. I've already used the backtraces a large number of times, by adding `trace.Printf(trace.Adhoc, -1, ...)`, calls understand the context for a specific function such as those in *pkg/cli/term/setup_unix.go*. Too, sometimes similarly worded messages occur in multiple places. I actually think all of the existing log messages should include the current stack frame for context. That is, the new log calls should be `trace.Printf(trace.SOME_CLASS, 1, ...)` rather than passing zero for the call context argument.

Note too that `eval.Println()` is implemented in terms of `trace.Printf()`. I did that for several reasons. Not least of which is consistency in the format of the output. In the future I expect `trace.Printf()` to honor the `-json` flag and emit JSON rather than the current human friendly format. Too, in the process of writing this code I've already modified `eval.Println()` a couple of times to change the call of `trace.Printf()` to pass -1 rather than zero for the backtrace argument to better understand the context for something like the `(*delElemOp).exec()` function. I haven't decided if that is likely to be used often enough to make it configurable at run time.

> It's better to put all the documentation in the code; the output of elvish -help can be exported to a web page on https://elv.sh

The problem is that produces voluminous output that 99.9% of elvish users wanting to know about the basic set of CLI flags won't care about. Too, the usage info for each flag written by `flag.PrintDefaults()` function is pre-formatted plain text. Which makes it impossible to style the text (or include embedded links). It also means the output isn't automatically resized to fit the screen.

> This is how Go's command documentation works too (https://golang.org/cmd/go/ has the same content as go help).

Not exactly. The command documentation in those web pages appears to depend on an internal structure, `base.Command`, not the flag package. See, for example, https://golang.org/src/cmd/go/internal/work/build.go. But my earlier point remains. Notice how that usage information is plain text and while the web page shows that as a `<pre>` formatted block the web page also includes a lot of additional information regarding the command line flags.

krader1961:
I do intend to implement a `trace` module to expose most of the functionality.  In particular, a `trace:printf &class=user "some message"` command that can be embedded in Elvish scripts. That command will be a low cost no-op unless the user does the equivalent of `-trace=user`. Whether there should be a single, or multiple, user message class(es) is TBD. Specifying the message class via a command option makes it possible to future proof the command if initially decide to support a single user message class and subsequently decide to support multiple classes.

Also TBD is how to handle different levels of detail in a particular message class. Lots of legacy UNIX programs have a verbose (`-v`) flag that can be specified multiple times to increase the level of detail. Including `tmux` that I still use every day and couldn't live without. I have never liked the verbose flag approach. That's part of what informed the current design of the trace package this P.R. introduces. My expectation is that as we gain more experience with how people use tracing we'll revise the trace API to support coarse and fine grain logging in some fashion. Whether that is by adding distinct message classes (e.g., `terminal` and `terminal-output`) or some other mechanism is TBD.

krader1961:
> I appreciate that you are adding documentation on the Elvish command itself, but there is a better place for it to live: in the output of elvish -help.

Perhaps some of that documentation should be moved to the flag definitions that are emitted by `flag.PrintDefaults()`; i.e., `elvish -help`. Nonetheless, there is still a need for a web page like https://golang.org/cmd/go/ that includes more information not reasonable to include in the output of `elvish -help`. I'm now thinking most of that information should be in a document, possibly named *website/ref/_running.md*, that is linked to from the top-level *README.md* and possibly other locations.

krader1961:
Also, I'm wondering whether it would be better to replace, or augment, the `-trace` CLI flag with something like a `ELVISH_TRACE` env var. There is a rich history of using environment variables for this type of behavior. Including the Go language tools. Which is another reason to keep all the tracing configuration a DSL of the `-trace` flag. Of course doing so raises the question of what to do if an Elvish script modifies that env var. Specifically, should it affect the tracing behavior of the running elvish program?

xiaq:
Summary of my current thinking:

* I still believe a `trace` function is eventually what we want. Again, I am not asking you to implement it as part of your PR.

* I still believe that there should be separate flags for tracing options, instead of a single `-trace` flag.

* I still believe that tracing and internal logging should be separate, and that the entire "logger" (how the code used to call it) or "class" (how your PR calls it) system in internal logging is over-engineered and should be just removed.

* I am convinced that printing backtraces in internal logging is useful.

* I think that the output of `elvish -help` doesn't have to be 100% the same as a full documentation for command-line usage, but duplication should be avoided.

Please find replies to your comments below.

---

> I disagree that this should be implemented in terms of a trace builtin. It would only be useful for the REPL case and would be cumbersome to use for the other use cases. It also begs the question of how you would use such a command to trace all the hidden activity that happens in an interactive shell; e.g., to render the prompt. Having said that, I would like there to be a way to trace just the commands run as a result of the user typing a statement at a REPL prompt; i.e., exclude exclude activity such as rendering the prompt.

You seem to contradict yourself: implementing tracing as a `trace` builtin will exactly enable the kind of granular tracing that can exclude activities such as rendering the prompt. Just run `trace { code to trace }`.

I'd encourage you to try to understand the other potential use cases I outlined earlier before dismissing the idea, but let me also elaborate a bit more: with a dynamic language like Elvish, making debug facilities accessible as a normal language facility instead of a facility **outside** the language is quite powerful. One simple case would be to enable tracing only conditionally; for example, if you have a line of code that calls `f`, and it only seems to misbehave when calling `f` with `x`, you can very easily express it like this:

```sh
# assume $a is set above
if (eq $a x) {
  trace { f $a } # call f with tracing enabled
} else {
  f $a # just call f
}
```

This is a technique similar to conditional breakpoints, and it can be a very efficient debug technique when you already have some vague idea of where things are going wrong.

Another significant upside of implementing `trace` as a function is that it makes it very easy to customize the output option, simply by supplying a callback function; supporting rich output options is much clumsier with command-line options. The callback could also do things like saving the traces in a variable, doing some filtering and transformation, and so on.

Yet another small upside is that it will make testing the tracing function much easier.

> Adding additional flags increases ambiguity. Not least because you end up with tens of flags. All of which implicitly enable tracing.

I don't find that to be a problem; FWIW, `go test` accepts several coverage-related flags: `-cover`, `-covermode`, `-coverpkg` and `-coverprofile`. It doesn't seem to be a problem for `go test`, at least not for me.

I'd also encourage you to not exaggerate the potential proliferation of flags - hyperbole makes your argument *weaker*, not stronger. Are there really "tens of flags"? I count 3: `-trace-classes` (which will not be needed, if you separate internal logging with traces), `-trace-file`, `-trace-time` (which can be `relative`, `utc` or `local`).

> Yes, but for someone not intimately familiar with the code (e.g., me) it can be extremely useful to understand the sequence of function calls that lead to a particular point in the code.

You convinced me on this, let's keep the stack traces for internal logs.

> Perhaps some of that documentation should be moved to the flag definitions that are emitted by flag.PrintDefaults(); i.e., elvish -help. Nonetheless, there is still a need for a web page like https://golang.org/cmd/go/ that includes more information not reasonable to include in the output of elvish -help. I'm now thinking most of that information should be in a document, possibly named website/ref/_running.md, that is linked to from the top-level README.md and possibly other locations.

Well, I agree with you that the output of `elvish -help` does not need to be exactly the same as the full documentation for command-line usage. But obviously they have some overlap, and I don't want the overlapping bits to be duplicated across different places. TBH I am not sure what the best way to handle would be.

> Note too that eval.Println() is implemented in terms of trace.Printf(). I did that for several reasons. Not least of which is consistency in the format of the output. ...

Reusing some code between internal logging and tracing is totally fine, and can be done without merging these two systems entirely together.

> Also, I'm wondering whether it would be better to replace, or augment, the -trace CLI flag with something like a ELVISH_TRACE env var. There is a rich history of using environment variables for this type of behavior. Including the Go language tools. Which is another reason to keep all the tracing configuration a DSL of the -trace flag. Of course doing so raises the question of what to do if an Elvish script modifies that env var. Specifically, should it affect the tracing behavior of the running elvish program?

I am not sure about `ELVISH_TRACE`; if you could cite some more examples of other programs using environment variables for similar behaviors, I might be convinced.

However, supporting an environment variable does *not* mean that it has to correspond to a single CLI flag; we could generalize the environment variable mechanism to all flags; for example `ELVISH_FLAGS` can contain `-trace -trace-file=...`.

Issue #1235, opened 2021-02-11
Add ability to define read-only variables


krader1961:
The introduction of the `var` builtin opens some interesting possibilities. The first one that comes to mind which might be useful is the ability to define readonly variables; e.g., `var &readonly myvar = value`.

krader1961:
When I first read the previous comment I interpreted `val` as `set`. That is, `val` could be used to assign a value to a declared var or, if the var had not been declared with `var`, it would define a readonly var. It took a couple of minutes before I realized `val` was meant to be an analogue of `var`: the former declaring a readonly var, the latter declaring a modifiable var. But that's not my primary objection to introducing a `val` special builtin.

When introducing compile time attributes for variables wouldn't it be better to use an extensible mechanism that doesn't require an open ended number of special builtins that can't be composed? Obviously, this presupposes that, in addition to the readonly attribute, we might want to introduce something like `var &num my-var = 42` to disallow assigning non-num values to `my-var`; at compile and run time both. It's not obvious introducing such attributes adds more value than problems (e.g., more runtime exceptions) but it seems like something that should be discussed before introducing yet another special builtin.

krader1961:
> The rationale is that I'll be gradually shifting Elvish towards a more FP style, where the majority of variables will be read-only local variables.

It seems far too late to turn Elvish into a FP style language where most variables are immutable, IMHO. Doing so would break most of the extant Elvish code and I don't see how even an unreasonably intelligent migration tool would mitigate the pain. At this stage it makes more sense, again IMHO, to provide a mechanism each snippet can use to detect unexpected mutations. Also, as I've said earlier, read-only is only one of many possible attributes it might be useful to enforce on a variable. For example, it might be useful to mark a variable as holding only a `num` datatype as a way to keep unexpected non-num assignments to the var from happening.

MirKml:
Hello, still no support for readonly variables?
This is useful feature even when whole style isn't mostly FP. Bash support it very long time and it helps a lot.
In bash it's possible to set readonly variable later with _readonly_ keyword also. From this moment, nobody can assign another value. But shadowing doesn't work for readonly vars in bash - which is little bit impractical.

But it's very useful for "configuration" part of command line script after arg parse. I use it a lot.
When I finish arg parse phase, all variables as result of this phase are set or re-set as readonly - the configuration - and are accessible as readonly global for rest of script - (because of troubles with non functional shadowing :-))

xiaq:
It's still in my plan, just haven't got around to it.

PR #1239, opened 2021-02-12
Improve error message when changing a read-only var

### LINKED ISSUE #255: Include variable name in error message when setting read-only variable

krader1961:
Modifying a read-only var should include the var name in the error message.

Fixes #255

### DIFF
--- a/pkg/edit/navigation.go
+++ b/pkg/edit/navigation.go
@@ -108,7 +108,7 @@ func initNavigation(ed *Editor, ev *eval.Evaler, nb eval.NsBuilder) {
 	binding := newMapBinding(ed, ev, bindingVar)
 	widthRatioVar := newListVar(vals.MakeList(1.0, 3.0, 4.0))
 
-	selectedFileVar := vars.FromGet(func() interface{} {
+	selectedFileVar := vars.FromGet("edit:selected-file", func() interface{} {
 		name := navigation.SelectedName(ed.app)
 		if name == "" {
 			return nil
--- a/pkg/edit/ns_helper_test.go
+++ b/pkg/edit/ns_helper_test.go
@@ -9,7 +9,7 @@ import (
 	"src.elv.sh/pkg/eval/vars"
 )
 
-var testVar = vars.NewReadOnly("")
+var testVar = vars.NewReadOnly("testVar", "")
 
 var eachVariableInTopTests = []struct {
 	builtin   *eval.Ns
@@ -31,14 +31,14 @@ var eachVariableInTopTests = []struct {
 	},
 	{
 		builtin: eval.NsBuilder{
-			"mod:": vars.NewReadOnly(eval.NsBuilder{"a": testVar, "b": testVar}.Ns()),
+			"mod:": vars.NewReadOnly("mod:", eval.NsBuilder{"a": testVar, "b": testVar}.Ns()),
 		}.Ns(),
 		ns:        "mod:",
 		wantNames: []string{"a", "b"},
 	},
 	{
 		global: eval.NsBuilder{
-			"mod:": vars.NewReadOnly(eval.NsBuilder{"a": testVar, "b": testVar}.Ns()),
+			"mod:": vars.NewReadOnly("mod:", eval.NsBuilder{"a": testVar, "b": testVar}.Ns()),
 		}.Ns(),
 		ns:        "mod:",
 		wantNames: []string{"a", "b"},
--- a/pkg/edit/testutils_test.go
+++ b/pkg/edit/testutils_test.go
@@ -37,7 +37,7 @@ func rc(codes ...string) func(*fixture) {
 func assign(name string, val interface{}) func(*fixture) {
 	return func(f *fixture) {
 		f.Evaler.AddGlobal(eval.CombineNs(f.Evaler.Global(),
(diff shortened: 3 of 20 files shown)


krader1961:
I like the more accurate error messages produced by this change but am not happy with how intrusive it is. Here is an example of the more detailed error message:

```
> edit:selected-file = y
Exception: read-only variable edit:selected-file cannot be set
[tty 1], line 1: edit:selected-file = x
```

krader1961:
On the other hand notice how much more the error from doing `true = x` is. Without this change:

```
Exception: read-only variable; cannot be set
[tty 140], line 1: true = x
```

With this change:

```
Exception: read-only variable builtin:true cannot be set
[tty 7], line 1: true = x
```

xiaq:
> I like the more accurate error messages produced by this change but am not happy with how intrusive it is.

Indeed. Seeing how invasive the implementation needs to be to make the names of variables available to the error message, I feel this approach is not worth it.

It would be much less invasive to make the stack trace more accurate: for example, when running `set true = 1`, underline the `true` portion. This is slightly less obvious but at least the information is there.

Once we have that, another thing worth exploring is to somehow extract the variable name from the stack trace, and mix into the error message. Not sure how invasive that would be though.

krader1961:
I'm pretty sure that doing this magically, rather than the straightforward way this P.R. does it, is going to be at least as intrusive and far harder for future readers of the code to understand. Also, consider the following examples for why something like this change is helpful. Yes, the `eval` example is contrived but perhaps there are similar examples more likely to be seen in the wild.

```
> a true b = 1 2 3
Exception: read-only variable builtin:true cannot be set
[tty 1], line 1: a true b = 1 2 3
> x = nil
> eval $x' = true'
Exception: read-only variable builtin:nil cannot be set
Traceback:
  [eval 1], line 1:
    nil = true
  [tty 3], line 1:
    eval $x' = true'
```

Also, consider if issue #1235 is implemented; allowing users to define read-only vars.

xiaq:
> I'm pretty sure that doing this magically, rather than the straightforward way this P.R. does it, is going to be at least as intrusive and far harder for future readers of the code to understand.

I don't see how making the stack trace more accurate and underline the variable being assigned is "magical".

krader1961:
My use of "magical" was in reference to constructing the warning message -- not the highlighting of the backtrace. However, the two are roughly equivalent. Note that if you're going to highlight the variable in the read-only assignment you should also highlight the value portion of the statement. I am basically saying that I think this straightforward solution is preferable to deriving the var name from the AST. Despite the fact it touches a lot of files; albeit in trivial ways.

krader1961:
> ... I feel this approach is not worth it.

That might be true. But it does solve the problem in a straightforward manner that is easy to understand. And, perhaps more importantly, establishes a precedent for including adjunct information about each variable that is useful in many ways.

xiaq:
I saw you closed the PR, but it's still worthwhile to clarify my opinion:

First, this is my main proposal. The outcome is slightly less ideal than having the variable name in the message, but the implementation is much more straightforward:

> It would be much less invasive to make the stack trace more accurate: for example, when running `set true = 1`, underline the `true` portion. This is slightly less obvious but at least the information is there.

Regarding your comment:

>  And, perhaps more importantly, establishes a precedent for including adjunct information about each variable that is useful in many ways.

I'm not sure whether such a precedent is worth establishing; so far it only solves this single problem. You seem to have some other problems in mind where such information is useful, but it's not obvious to me. If it does turn out to be quite useful, I'm certainly open to revisiting this decision.

krader1961:
> I'm not sure whether such a precedent is worth establishing; so far it only solves this single problem. You seem to have some other problems in mind where such information is useful, but it's not obvious to me. If it does turn out to be quite useful, I'm certainly open to revisiting this decision.

I should have written "adjunct information..." that "could be useful" rather than "is useful in many ways". Which is to say, I can't help but think that the recently introduced `var` command opens up opportunities for new behavior. Such as a `&readonly` option as I [proposed here](https://github.com/elves/elvish/issues/1235). Yes, that doesn't require this change to produce improved error messages. However, I can't help but think other constraints (or attributes) might be better represented by var structure members than distinct var types.

Despite the above discussion I still think this straightforward solution is preferable to a hypothetical solution involving the parser that seems unlikely to be implemented in my lifetime. In no small part because that solution is greatly complicated by the reasonable expectation it highlight both the variable and value portion of the statement.

krader1961:
I also closed this pull-request because I don't really care if assignment to read-only var error messages is improved. I do, however, care about improved logging and thus want to focus my resources on P.R. #1229.

xiaq:
> Despite the above discussion I still think this straightforward solution is preferable to a hypothetical solution involving the parser that seems unlikely to be implemented in my lifetime.

The "hypothetical solution involving the parser that seems unlikely to be implemented in my lifetime" was implemented in a33ecb2da4ce02954fedad47ed40f01903f5ab4d; the change only takes 4 lines - the opcode structure actually already contains the range information of each LHS variable, they were just not being passed to `fm.errorpf`. If your goal was to provoke me into implementing this you have succeeded. :)

PR #1252, opened 2021-02-26
Fix when cursor is shown by `edit:clear`

### LINKED ISSUE #1238: Key bound clear command hide prompts
Hello,
following the example in https://elv.sh/ref/edit.html#keybindings I tried `edit:insert:binding[Ctrl-L] = { clear > /dev/tty }`.

Unlike running the `clear` command, which leave me with an empty terminal with the prompt (expected behavior), using `Ctrl-L` empties completely the screen, deleting right and left prompts but still putting the cursor in the right place (where the left prompt should end).

I tried to fix by using `edit:insert:binding[Ctrl-L] = { clear > /dev/tty ; edit:redraw &full=$true }` but this makes the cursor "jump" and that gets pretty annoying.

Is seems like a bug to me, but I may be missing something.

Elvish 0.15.0
OS: Ubuntu 20.04
Terminal: Windows Terminal

Thank you.

krader1961:
Showing the cursor should only happen after the prompt is redrawn.

Fixes #1238

### DIFF
--- a/pkg/cli/clitest/fake_tty.go
+++ b/pkg/cli/clitest/fake_tty.go
@@ -126,6 +126,12 @@ func (t *fakeTTY) UpdateBuffer(bufNotes, buf *term.Buffer, _ bool) error {
 	return nil
 }
 
+func (t *fakeTTY) HideCursor() {
+}
+
+func (t *fakeTTY) ShowCursor() {
+}
+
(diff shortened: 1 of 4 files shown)


xiaq:
Thanks. This reminds me of some cleanup I still wanted to do with the TTY interface, which I'll do after merging this.

PR #1253, opened 2021-02-26
Deprecated builtins

krader1961:


pkg/eval/var_ref.go
```diff
@@ -42,15 +42,20 @@ const (
...
 	searchLocal(k string) int
 	searchCapture(k string) int
 	searchBuiltin(k string, r diag.Ranger) int
+	checkIfDeprecated(k string, r diag.Ranger)
 }
 
 // Resolves a qname into a varRef.
 func resolveVarRef(s scopeSearcher, qname string, r diag.Ranger) *varRef {
 	qname = strings.TrimPrefix(qname, ":")
+	s.checkIfDeprecated(qname, r)
```
xiaq, line 58: The idea of using a single map to track deprecated variables is misguided, and this place is where things are wrong: how to refer to the namespace a variable lives in depends on the scope context, so attempting to match `qname` with static keys like `builtin:eval-symlink~` is not correct.

For example, if the `qname` here is `builtin:eval-symlink~`, it does not necessarily refer to `eval-symlink~` in the builtin namespace because `$builtin:` could be shadowed and refer to a different namespace (e.g. by running `use path builtin`).

As an opposite example, if the `qname` here is `foo:eval-symlink~`, it could still refer to `eval-symlink~` in the builtin namespace because `$foo:` could be an alias for the builtin namespace (e.g. by running `use builtin foo`).

These may sound like pedantic examples because people probably don't twist the builtin namespace in such ways. However, other builtin namespaces are quite likely to be aliased; for example, if the generalized deprecation table is used to deprecate builtins in `path:`, it won't catch any usages if the user just does `use path p`.

For the generalization to work, deprecation information needs to be attached to `Ns` objects; this PR is simplistic and doesn't work.

krader1961: > for example, if the generalized deprecation table is used to deprecate builtins in path:, it won't catch any usages if the user just does `use path p`.

That is true, and this PR is simplistic, but it does address some of the problems with the current implementation which doesn't handle any of your scenarios correctly. I'm going to leave it to you to create a proper fix since I lack sufficient motivation to create a solution you're likely to accept.

Issue #1240, opened 2021-02-13
Broken symlink causes wildcard matching to fail


muesli:
I have a directory with two files in it:

```
file1
file2
```

`file1` is a symlink pointing to a file that doesn't exist anymore.
`file2` is a regular file.

`ls file*` now fails: `Exception: wildcard has no match`

Fixing the symlink of `file`1 resolves the issue.

hanche:
As @krader1961 has mentioned in chat, this issue and #1236 are probably related.

krader1961:
I just tried an experiment with several files and directories, including a dead symlink, in a test dir. Running `put *` **does not throw an error**. It simply stops enumerating files when it reaches the dead symlink. Note that `put d*` does throw an exception. Which suggests there might be a second problem with wildcard expansion; or at least the need to unify some code for consistent behavior.

clonejo:
One example where this bug hits hard is a git-annex repo where not all files are locally available. Those files are then represented by dangling symlinks. Normally you could do move operations on those symlinks just fine, but not having wildcards available makes it a hassle.

PR #1254, opened 2021-02-27
Handling symlinks with an unresolvable target

### LINKED ISSUE #1236: Should path:eval-symlinks behavior be changed for invalid paths?
**TL;DR:**

The `path:eval-symlinks` command should just return the input path when it does not exist rather than throwing a "no such file or directory" exception. To be determined is what to do about other errors returned by `os.Lstat`. Should all errors result in returning the original path or should only the "not exists" error be treated specially? It's not clear what other error conditions `os.Lstat` can return or how to distinguish between the various possibilities.

Note that the external `realpath` command, that `path:eval-symlinks` is intended to mimic, exits with a zero status when the path is invalid and simply regurgitates the input path.

**The long version:**

I noticed @xiaq fixed a bug in the `path:is-dir` and `path:is-regular` commands I introduced. Specifically, the documentation stated that a symlink (in the final component) would always result in those commands returning false. If you want symlinks to be ignored you first need to apply `path:eval-symlinks`. My implementation used `os.Stat` when it should have used `os.Lstat`. However, @xiaq's fix caused me to notice something quite annoying about `path:eval-symlinks`.

My *~/.elvish/rc.elv* script used to contain this statement:

```
if (not (path:is-dir $p)) {
```

That `path:is-dir` outputs `$false` if the path does not refer to a directory; even if it is a symlink to a directory. I made the naive update:

```
if (not (path:is-dir (path:resolve-symlinks $p)) {
```

That results in `path:resolve` throwing an exception if the path does not exist. I instead have to write this:

```
if (not (path:is-dir (try { path:eval-symlinks $p } except { put $p }))) {
```

That seems suboptimal. Note that the full context is this block of code I now have near the bottom of my *~/.elvish/rc.elv* script:

```
# Warn about any non-existent directories in $E:PATH.
each [p]{
  if (not (path:is-dir (try { path:eval-symlinks $p } except { put $p }))) {
    echo (styled "Warning: directory "$p" in $paths does not exist." red)
  }
} $paths
```

Note that the external `realpath` command, that `path:eval-symlinks` mimics, exits with a zero status when the path is invalid and simply regurgitates the input path.

krader1961:
Symlinks in a pathname that cannot be resolved because the target is
not valid represent a special case. Specifically, if resolving a symlink
results in a syscall.ENOENT error it is friendlier, and generally more
useful, to treat it as a non-error that simply outputs the original path.
This makes the command consistent with the behavior of the external
`realpath` it mimics.

Fixes #1236

### DIFF
--- a/0.16.0-release-notes.md
+++ b/0.16.0-release-notes.md
@@ -11,6 +11,10 @@ This is the draft release notes for 0.16.0. It is scheduled to be released on
     Instead, the history listing mode now applies smart-case matching by
     default.
 
+-   The `path:eval-symlinks` command no longer returns an error when the
+    argument is an invalid path; instead it now returns the original path.
+    ([#1236](https://b.elv.sh/1236)).
+
 # Notable bugfixes
 
 -   The `path:is-dir` and `path:is-regular` commands no longer follow symlinks,
--- a/pkg/eval/mods/path/path.go
+++ b/pkg/eval/mods/path/path.go
@@ -2,8 +2,10 @@
 package path
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
+	"syscall"
 
 	"src.elv.sh/pkg/eval"
 )
@@ -17,7 +19,7 @@ var fns = map[string]interface{}{
 	"clean":         filepath.Clean,
 	"dir":           filepath.Dir,
 	"ext":           filepath.Ext,
-	"eval-symlinks": filepath.EvalSymlinks,
+	"eval-symlinks": evalSymlinks,
 	"is-abs":        filepath.IsAbs,
 	"is-dir":        isDir,
 	"is-regular":    isRegular,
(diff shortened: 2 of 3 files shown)


krader1961:
If this update is merged we'll want to make a similar change to the wildcard/glob expansion logic. See issue #1240.

xiaq:
With the current default behavior (throws exception for broken symlink), it is very easy to manually implement the proposed fallback behavior:

```
fn eval-symlink-tolerant [p]{
  try {
    path:eval-symlink $p
  } except {
    put $p
  }
}
```

But if the default behavior is changed to fall back to the original path for broken symlink, it's much harder to get the original default behavior back.

I am also not sure when the fallback behavior is actually useful. I can't think of any use cases, but you probably have some in mind. If it is genuinely useful, it could probably be an option to `eval-symlink`.

krader1961:
> it is very easy to manually implement the proposed fallback behavior... *using an Elvish function*

No, it isn't. Precisely because the specific error varies by platform and is otherwise hard to detect using Elvish even if you limit the problem to UNIX. See my comment inside the `evalSymlinks` function.

> I can't think of any use cases, but you probably have some in mind. 

See my problem description in issue #1236. More generally, this is how the external `realpath` command, that this feature mimics, has behaved forever. Yes, Elvish explicitly deviates from the POSIX shell standard so that is a weak argument. Nonetheless, this seems to be a case where deviating from existing practice will cause problems for no benefit. It certainly did for me. Just because the Go stdlib `filepath.EvalSymlinks` function returns an error for this case does not mean that propagating the error to Elvish is appropriate. The tradeoffs for the two languages are different.

> If it is genuinely useful, it could probably be an option to `eval-symlink`.

I'm happy to make the behavior optional. Updated PR forthcoming.

krader1961:
Closing this PR since I realized my current approach, even if made optional, is not exactly what we want. Specifically, this solution does not constrain the ENOENT handling to the final component of the path due to limitations of the Go stdlib `filepath.EvalSymlinks` function. It is not clear that we can mimic the behavior of the legacy `realpath` command without significant effort.

PR #1260, opened 2021-03-10
Add equivalent of `mktemp -d` as a builtin

### LINKED ISSUE #1255: The `path` module should have something like a `mktemp` command
I was looking at my `external_edit_command` function and noticed I had hardcoded a */tmp/...* path to avoid the pain of using the external `mktemp` command. The pain is due to the SysV, BSD, and Gnu implementations having different behavior. Not to mention creating a temp file on Windows. A `path:mktemp` command would be very useful. We don't necessarily want, or need, to mimic the existing external command but obviously do want options to create a unique directory versus a regular file and some sort of pattern mechanism to control the naming of the final component of the path.

krader1961:
Resolves #1255

### DIFF
--- a/0.16.0-release-notes.md
+++ b/0.16.0-release-notes.md
@@ -49,6 +49,10 @@ New features in the standard library:
 
 -   A new `file:` module contains utilities for manipulating files.
 
+-   Commands for creating, what are usually meant to be temporary, unique
+    directories, path:temp-dir, and regular files, path:temp-file
+    ([#1255](https://b.elv.sh/1255)).
+
 New features in the interactive editor:
 
 -   The editor now supports setting global bindings via `$edit:global-binding`.
--- a/pkg/eval/evaltest/evaltest.go
+++ b/pkg/eval/evaltest/evaltest.go
@@ -43,6 +43,10 @@ type Result struct {
 	Exception        error
 }
 
+// MatchingRegexp is used in a `Puts()` call when the value being matched should be a string
+// matching a regexp rather than a literal string.
+type MatchingRegexp struct{ Pattern string }
+
 // The following functions and methods are used to build Test structs. They are
 // supposed to read like English, so a test that "put x" should put "x" reads:
 //
@@ -232,8 +236,15 @@ func matchOut(want, got []interface{}) bool {
 				return false
 			}
 		default:
-			if !vals.Equal(got[i], want[i]) {
-				return false
+			switch w := want[i].(type) {
+			case MatchingRegexp:
+				if !vals.MatchesRegexp(got[i], w.Pattern) {
+					return false
+				}
+			default:
+				if !vals.Equal(got[i], want[i]) {
+					return false
+				}
 			}
 		}
 	}
--- a/pkg/eval/mods/path/path.go
+++ b/pkg/eval/mods/path/path.go
@@ -2,10 +2,12 @@
 package path
 
 import (
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
 	"src.elv.sh/pkg/eval"
+	"src.elv.sh/pkg/eval/errs"
 )
 
 // Ns is the namespace for the re: module.
@@ -21,6 +23,8 @@ var fns = map[string]interface{}{
 	"is-abs":        filepath.IsAbs,
 	"is-dir":        isDir,
 	"is-regular":    isRegular,
+	"temp-dir":      tempDir,
+	"temp-file":     tempFile,
 }
 
 //elvdoc:fn abs
@@ -175,4 +179,100 @@ func isDir(path string) bool {
 func isRegular(path string) bool {
 	fi, err := os.Lstat(path)
 	return err == nil && fi.Mode().IsRegular()
+}
+
+//elvdoc:fn temp-dir
+//
+// ```elvish
+// temp-dir &dir=$dir $pattern?
+// ```
+//
+// Create a unique directory and output its name. The &dir option determines where the directory
+// will be created, and its default value is appropriate for your system. The `$pattern` value is
+// optional. If omitted it defaults to `elvish-*`. The last star in the pattern is replaced by a
+// random string. It is your responsibility to remove the (presumably) temporary directory.
+//
+// ```elvish-transcript
+// ~> path:temp-dir
+// ▶ /tmp/elvish-RANDOMSTR
+// ~> path:temp-dir x-
+// ▶ /tmp/x-RANDOMSTR
+// ~> path:temp-dir 'x-*.y'
+// ▶ /tmp/x-RANDOMSTR.y
+// ~> path:temp-dir &dir=.
+// ▶ elvish-RANDOMSTR
+// ~> path:temp-dir &dir=/some/dir
+// ▶ /some/dir/elvish-RANDOMSTR
+// ```
+
+type mktempOpt struct{ Dir string }
+
+func (o *mktempOpt) SetDefaultOptions() {}
+
+func tempDir(opts mktempOpt, args ...string) (string, error) {
+	var pattern string
+	switch len(args) {
+	case 0:
+		pattern = "elvish-*"
+	case 1:
+		pattern = args[0]
+	default:
+		return "", errs.ArityMismatch{
+			What:     "arguments here",
+			ValidLow: 0, ValidHigh: 1, Actual: len(args)}
+	}
+
+	return ioutil.TempDir(opts.Dir, pattern)
+}
+
+//elvdoc:fn temp-file
+//
+// ```elvish
+// temp-file [&dir=$dir] [$pattern]
+// ```
+//
+// Create a unique file and output a [file](language.html#file) object opened for reading and
+// writing. The &dir option determines where the directory will be created, and its default value is
+// appropriate for your system. The `$pattern` value is optional. If omitted it defaults to
+// `elvish-*`. The last star in the pattern is replaced by a random string. It is your
+// responsibility to remove the (presumably) temporary file.
+//
+// You can use [`fclose`](builtin.html#fclose) to close the file. You can use `$f[name]` to extract
+// the name of the file so it can be used as an argument for another command; e.g., `rm`.
+//
+// ```elvish-transcript
+// ~> f = path:temp-file
+// ~> put $f[name]
+// ▶ /tmp/elvish-RANDOMSTR
+// ~> echo hello > $f
+// ~> cat $f[name]
+// hello
+// ~> f = path:temp-file x-
+// ~> put $f[name]
+// ▶ /tmp/x-RANDOMSTR
+// ~> f = path:temp-file 'x-*.y'
+// ~> put $f[name]
+// ▶ /tmp/x-RANDOMSTR.y
+// ~> f = path:temp-file &dir=.
+// ~> put $f[name]
+// ▶ elvish-RANDOMSTR
+// ~> f = path:temp-file &dir=/some/dir
+// ~> put $f[name]
+// ▶ /some/dir/elvish-RANDOMSTR
+// ```
+
+func tempFile(opts mktempOpt, args ...string) (*os.File, error) {
+	var pattern string
+	switch len(args) {
+	case 0:
+		pattern = "elvish-*"
+	case 1:
+		pattern = args[0]
+	default:
+		return nil, errs.ArityMismatch{
+			What:     "arguments here",
+			ValidLow: 0, ValidHigh: 1, Actual: len(args)}
+	}
+
+	return ioutil.TempFile(opts.Dir, pattern)
 }
--- a/pkg/eval/mods/path/path_test.go
+++ b/pkg/eval/mods/path/path_test.go
@@ -2,9 +2,11 @@ package path
 
 import (
 	"path/filepath"
+	"regexp"
 	"testing"
 
 	"src.elv.sh/pkg/eval"
+	"src.elv.sh/pkg/eval/errs"
 	. "src.elv.sh/pkg/eval/evaltest"
 	"src.elv.sh/pkg/testutil"
 )
@@ -30,6 +32,10 @@ func TestPath(t *testing.T) {
 	if err != nil {
 		panic("unable to convert a/b/c.png to an absolute path")
 	}
+
+	// This is needed for path tests that use a regexp for validating a path since Windows uses a
+	// backslash as the path separator and a backslash is special in a regexp.
+	sep := regexp.QuoteMeta(string(filepath.Separator))
 
 	setup := func(ev *eval.Evaler) {
 		ev.AddGlobal(eval.NsBuilder{}.AddNs("path", Ns).Ns())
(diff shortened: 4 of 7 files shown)


krader1961:
> this is nice, but why only emulate mktemp -d and not mktemp (i.e. also create temp files)?

That will be in a separate commit. Primarily because it's a little more complicated due to wanting to preserve the semantics of the `ioutil.TempFile` function that returns an open file, not just a file name.

krader1961:
@zzamboni, I've added a `path:temp-file` command. If you can, and are willing, to build an elvish program from this pull-request I would like feedback regarding the dichotomy of the return values of the `path:temp-dir` and `path:temp-file` commands.

@xiaq, This P.R. has caused me to notice that there are unrelated problems with the use of Elvish `file` objects. Such as not consistently recognizing when they are closed. But I think those observations are outside the scope of this enhancement.

zzamboni:
@krader1961 I have tested your branch, it looks good to me. I locally updated my `alias` module to use `path:temp-file` instead of `mktemp` and it works fine.

I think it can be a source of confusion that `path:temp-file` returns a file object instead of just the path, particularly with `path:temp-dir` returns the path. Are there any advantages of returning the file object instead of the path?

In any case, thanks for this! I think it's very welcome functionality.

zzamboni:
For reference, I updated some of my modules to use `path:temp-file` instead of `mktemp`, the changes were trivial: https://github.com/zzamboni/elvish-modules/commit/9edc73cb639f36b503e7b2ef24b7be3804bb1cef

krader1961:
> Are there any advantages of returning the file object instead of the path?

The primary advantage is that it helps avoid certain classes of security bugs involving temp files. However, we can't fully take advantage of that until there is additional support for file objects such as a `file:seek` (or at least a `file:rewind`) command. Of course, there is also the pragmatic reason that the Go stdlib function we're using returns a file object rather than just a path name.

0.16.0-release-notes.md
```diff
@@ -21,6 +21,9 @@ This is the draft release notes for 0.16.0. It is scheduled to be released on
 
 # Notable new features
 
+-   Command for creating temporary directories: path:temp-dir
+    ([#1255](https://b.elv.sh/1255)).
+
```
xiaq, line 26: Document path:temp-file too

pkg/eval/evaltest/evaltest.go
```diff
@@ -35,6 +35,7 @@ type TestCase struct {
 }
 
 type Result struct {
+	ValueRe   bool // are ValueOut values treated as a regexp matching the corresponding value
```
xiaq, line 38: Support for regexp matching is better implemented using a special type (say `MatchingRegexp`), like how Approximately works. This will allow mixing outputs that should be regexp-matched and those that don't:

```
  That("some code").Puts("foo", MatchingRegexp{Pattern: `.*bar`})
```

pkg/eval/mods/path/path.go
```diff
@@ -176,3 +181,100 @@ func isRegular(path string) bool {
...
 }
+
+//elvdoc:fn temp-dir
+//
+// ```elvish
+// temp-dir [&dir=$dir] [$pattern]
+// ```
+//
+// Create a temporary, unique, directory and output its name. The default for the `&dir` option is a
+// value appropriate for your system. You can use `&dir=/some/path` to specify a particular
```
xiaq, line 192: > The default for the `&dir` option is a value appropriate for your system.

Better to first say what the `&dir` option **is** before saying what it defaults to. Something like:

> The `&dir` option determines where the directory will be created, and its default value ...

pkg/eval/mods/path/path.go
```diff
@@ -176,3 +181,100 @@ func isRegular(path string) bool {
...
+	switch len(args) {
+	case 0:
+		pattern = "elvish-*"
+	case 1:
+		pattern = args[0]
+	default:
+		return "", errs.ArityMismatch{
+			What:     "arguments here",
+			ValidLow: 0, ValidHigh: 1, Actual: len(args)}
+	}
```
xiaq, line 225: This is worth extracting to a function like:

```go
func getOptionalStringArg(args []string, default string) (string, error) { ... }
```

pkg/eval/mods/path/path.go
```diff
@@ -176,3 +181,100 @@ func isRegular(path string) bool {
 	fi, err := os.Lstat(path)
 	return err == nil && fi.Mode().IsRegular()
 }
+
+//elvdoc:fn temp-dir
+//
+// ```elvish
+// temp-dir [&dir=$dir] [$pattern]
```
xiaq, line 188: Please follow the convention for the signature of othe commands. So this will be:

```
temp-dir &dir=$dir $pattern?
```

pkg/eval/mods/path/path.go
```diff
@@ -176,3 +181,100 @@ func isRegular(path string) bool {
...
+			ValidLow: 0, ValidHigh: 1, Actual: len(args)}
+	}
+
+	return ioutil.TempDir(opts.Dir, pattern)
+}
+
+//elvdoc:fn temp-file
+//
+// ```elvish
+// temp-file [&dir=$dir] [$pattern]
```
xiaq, line 233: Same as above

pkg/eval/mods/path/path.go
```diff
@@ -176,3 +181,100 @@ func isRegular(path string) bool {
...
 	return err == nil && fi.Mode().IsRegular()
 }
+
+//elvdoc:fn temp-dir
+//
+// ```elvish
+// temp-dir [&dir=$dir] [$pattern]
+// ```
+//
+// Create a temporary, unique, directory and output its name. The default for the `&dir` option is a
```
xiaq, line 191: > Create a temporary, unique, directory

The last comma is bogus

pkg/eval/vals/aliased_types.go
```diff
@@ -7,8 +7,37 @@ import (
...
 )
 
-// File is an alias for *os.File.
-type File = *os.File
+// File wraps *os.File so that Elvish can treat the object as a read-only psuedo-map.
+type File struct {
+	OsFile *os.File `elvish:"hidden"`
+	// IoMode causes `$f[io-mode]` to return "ro" if the file is read-only, "rw" if it is
+	// read-write, and "wo" if it is write-only.
+	IoMode string
```
xiaq, line 15: In general, given an `os.File`, it is not possible to know the mode it was opened with; this makes the `File` type here only suitable for wraping `os.File` created in the Elvish runtime and limits interoperability. Better to just leave out this field.

pkg/eval/vals/aliased_types.go
```diff
@@ -7,8 +7,37 @@ import (
...
+}
+
+// IsStructMap tags File as a psuedo-map.
+func (File) IsStructMap() {}
```
xiaq, line 19: What is publically documented as pseudo-map in fact corresponds to two different internal concepts:

* A structmap is a pure data type - it is a struct that has a one-to-one correspondance between its struct fields and the fields exposed to Elvish

* A psuedo-structmap is any type that may be introspected like a map, but may contain other internal information.

The File type here is a pseudo-structmap and should be implemented as such (see https://github.com/elves/elvish/blob/f5aaba06911ac9d47d40adcb1f2fe615c77c741e/pkg/eval/vals/struct_map.go#L38)

pkg/eval/vals/struct_map.go
```diff
@@ -7,29 +7,32 @@ import (
 	"src.elv.sh/pkg/strutil"
 )
 
-// StructMap may be implemented by a struct to mark the struct as a "struct
-// map", which causes Elvish to treat it like a read-only map. Each exported,
-// named field and getter method (a method taking no argument and returning one
-// value) becomes a field of the map, with the name mapped to dash-case.
+// The StructMap interface may be implemented by a struct to mark the struct as a psuedo-map. This
```
xiaq, line 10: Don't change this file; see my comment above about structmap vs pseudo-structmap.

pkg/eval/vals/aliased_types.go
```diff
@@ -7,8 +7,37 @@ import (
...
+// Fd causes `$f[fd]` to return the numeric file descriptor of the embedded os.File object.
+//
+// We coerce it to a float64 since that is currently the only Elvish numeric type and we don't want
+// the value coerced to a string.
+func (f File) Fd() float64 {
+	fdPtr := f.OsFile.Fd()
+	if fdPtr == ^uintptr(0) {
+		return -1
+	}
+	return float64(fdPtr)
```
xiaq, line 36: The default for exporting integer values to Elvish is to convert it to a string (https://github.com/elves/elvish/blob/f5aaba06911ac9d47d40adcb1f2fe615c77c741e/pkg/eval/vals/conversion.go#L110). Before Elvish gets an integer type, better to keep the consistency.

xiaq (changes requested):

pkg/eval/evaltest/evaltest.go
krader1961: Good idea. This change doesn't require the generalization but it could be useful in the future and is a better API. Done.

0.16.0-release-notes.md
krader1961: Done in the third commit of this P.R. which actually introduces that command.

pkg/eval/mods/path/path.go
krader1961: Done

pkg/eval/mods/path/path.go
krader1961: Changed. Old habits die hard :-)

pkg/eval/mods/path/path.go
krader1961: Not only that, the first comma should be replaced by "and": `Create a temporary and unique directory...`. In fact, I'll go further and state that the "temporary" qualifier should be omitted since the directory is not temporary if the program doesn't explicitly delete it. Which is noted by the last sentence of the documentation.

pkg/eval/mods/path/path.go
krader1961: Done

pkg/eval/vals/aliased_types.go
krader1961: True, but at the moment we don't wrap any `os.File` object that we didn't create. We might want to wrap the equivalent of `os.Stdin`, et.al. in the future where we do not know the I/O mode. So rather than remove the field, since it is potentially useful for diagnostic purposes, I'll document that it could be `"??"` if Elvish doesn't know the I/O mode. Too, if we do ever expose the `os.Std*` objects as "file" objects that can be directly manipulated we can reasonably default the `IoMode` to "ro" if we're wrapping `os.Stdin`, otherwise "wo".

pkg/eval/mods/path/path.go
krader1961: FWIW, grepping the source for `len(args)` yielded only one other use that could use a `getOptionalStringArg` function: the [`cd` command](https://github.com/elves/elvish/blob/29034c20e9a30e9826c9f9aeac309b4d519db08e/pkg/eval/builtin_fn_fs.go#L69-L80). While my case has a zero cost default value the `cd` case does not. Which will require using a lambda, rather than string, arg for the default value. And even that is complicated by the need to handle an error return from the lambda that returns a default string.

I don't think a `getOptionalStringArg ` function is worth it at this time. This should be revisited if the ["here-values" syntax](https://github.com/elves/elvish/issues/1261) is implemented.

krader1961: I also looked at every file matched by `grep -h ' \.\.\.string)'` and, again, only found my two uses and the `cd` case. So I am still inclined to think the proposed abstraction is premature given that a default value cannot be counted on to be a static string.

krader1961: Also, it just occurred to me the default value might not even be a string. Which further complicates the implementation. Yes, the default value is a string for the one existing case, and the two cases I introduced, but that is by no means guaranteed to be true for future cases.

@xiaq, Feel free to comment further, and un-resolve this conversation, but as you can tell I feel pretty strongly that this is not the time to introduce that abstraction. Certainly not in the context of this pull-request which is already non-trivial.

pkg/eval/vals/aliased_types.go
krader1961: Done.

pkg/eval/vals/aliased_types.go
krader1961: @xiaq: I'm still struggling, after several days of looking at the code, to see the point of the `PseudoStructMap` interface being a distinct type from `StructMap`. What I opted to do was add an `elvish:"hidden"` struct tag to flag that an otherwise public struct member should not be exposed by the Elvish pseudo-map mechanism. That avoids the extra layer of indirection incurred by the `PseudoStructMap` interface. It is also, IMHO, considerably simpler when compared to the `PseudoStructMap` interface.

Granted, I did that because I didn't understand the `PseudoStructMap` interface. I simply looked at how the `StructMap` interface worked and built on that by introducing support for the `elvish:"hidden"` struct tag. Nonetheless, I think my approach is preferable. I am confident that I can rewrite the existing uses of `PseudoStructMap` to instead use just the `StructMap` interface and greatly simplify the existing code.

krader1961:
I force-pushed an update of this pull-request to reflect a rebase on the master branch and addressing all criticisms other than whether the new `File` type is a `StructMap` or `PseudoStructMap`.

pkg/eval/vals/aliased_types.go
xiaq: Struct maps are transparent; they are fully described by their fields - like maps, just with an immutable key set and type-constrained values. Pseudo-struct maps are opaque, but just happen to support introspection by the indexing operation (hence the "pseudo", since they are not really maps). This is why struct maps should not support hiding some of their fields.

The distinction between transparent and opaque types is not consequential for Elvish right now, but it will be in future when Elvish's type system get developed more. Many languages in the Lisp tradition have always made such a distinction - Racket has [opaque and transparent structs](https://docs.racket-lang.org/guide/define-struct.html#%28part._trans-struct%29) and [classes](https://docs.racket-lang.org/guide/classes.html), Clojure has [deftype and defrecord](https://clojure.org/reference/datatypes). The [Why have both deftype and defrecord?](https://clojure.org/reference/datatypes#_why_have_both_deftype_and_defrecord) section from the Clojure doc is especially insightful.

Even Java is catching up - although Java's OO style has always been geared towards opaque types, there's been a [recent proposal](https://cr.openjdk.java.net/~briangoetz/amber/datum.html) on introducing "data classes", something Google implemented as [@AutoValue](https://github.com/google/auto/blob/master/value/userguide/index.md) long ago.

pkg/eval/vals/aliased_types.go
xiaq: That can certainly work, but after thinking more about this matter I think it's best to not wrap the `*os.File` type at all. Instead, to support introspection on `*os.File` objects, add it to the type switch in the `Index` function. My experience over working on the value system of Elvish suggests that it's best to stick to the types offered by Go as much as possible.

For the purpose of this PR, I recommend that you just leave out all the introspection-related changes (i.e. all files in `pkg/eval/vals`). I will do a followup commit to implement introspection in my preferred fashion.

krader1961: Without introspection the output of `path:temp-file` isn't really useful since there wouldn't be any way to delete the temporary file. I'll implement indexing support so `$file[fd]` and `$file[name]` yield useful values. I'll leave it to you to decide whether, and how, a `*os.File` is exposed as a pseudo-map so commands like `keys` work on such objects. I'll also squash the commits and rebase on the current master branch.

krader1961:
@xiaq: Please take another look. I replaced my incorrect struct map implementation with simpler support for indexing so the `fd` and `name` keys on a "file" object yield the expected values. I am not happy with the need to introduce expressions such as `vals.Num(big.NewInt(-1))`, but that is a consequence of the new Elvish number type which does not yet have established patterns for its use.

xiaq:
Please rebase on the latest commit, the diff is now showing all my changes to the number system.

I'm not sure why you need to use `vals.Num(big.NewInt(-1))`, you should be able to just write `-1`; Go's `int`s are automatically interoperable in Elvish.

xiaq:
> ... but that is a consequence of the new Elvish number type which does not yet have established patterns for its use

The pattern for non-numeric commands dealing with integers is simple: just use `int`.

If you are going to change some of the numeric commands, you should be able to follow the existing patterns.

krader1961:
Not sure what happened with this pull-request. I did `git pull upstream master` then `git rebase master` and had to make some changes to accommodate your commits before doing a `git push -f`. Borg standard workflow.

As for the `vals.Num(big.NewInt(-1))`, that's because in the process of figuring out how to handle the uintptr return value from *File.Fd() I initially coerced it to a int64 which is not interoperable with the new Elvish number system. Which resulted in my convoluted solution. I now realize that coercing the uintptr to an int does the right thing even on 32-bit platforms and for negative and positive values.

xiaq (approved):
There are a few minor changes I'd like to make, which I'll do after merging this PR.

PR #1264, opened 2021-03-15
Some golint cleanups

krader1961:


pkg/cli/mode/mode.go
```diff
@@ -4,18 +4,22 @@ package mode
...
 // mode widget.
-func ModePrompt(content string, space bool) func() ui.Text {
-	p := ModeLine(content, space)
+func modePrompt(content string, space bool) func() ui.Text {
+	p := modeLine(content, space)
 	return func() ui.Text { return p }
 }
+
+// Prompt is solely for the benefit of pkg/edit/minibuf.
+// TODO: Figure out how to make that a CLI widget so we don't have to publicly export modePrompt.
```
xiaq, line 24: The prompt can't be easily turned into a CLI widget with the current widget abstraction - even if it could, it seems overkill. So it is intentional that this is exported, it's not something that needs to be fixed.

What you're doing here - de-exporting the function and name the exported alias without the `Mode` suffix - is fine though. Personally I'm not too bothered by golint's warning, but I'm also not against fixing them. Just remove the TODO here, and move the docstring of `modePrompt` here so that it shows up nicely in godoc.

pkg/cli/mode/navigation.go
```diff
@@ -1,6 +1,6 @@
-// Package navigation provides the functionality of navigating the filesystem.
 package mode
 
+// This module provides the navigation mode widget for navigating the filesystem.
```
xiaq, line 3: This is some leftover when the nav mode was in its own module. Just remove this.

pkg/testutil/testdir_test.go
```diff
@@ -124,21 +124,21 @@ func TestApplyDir_Symlinks(t *testing.T) {
 		{"sf1", filepath.Join("d1", "f")},
 	}
 	for _, fromTo := range symlinkTargets {
-		symlink, expected_target := fromTo[0], fromTo[1]
+		symlink, expectedTarget := fromTo[0], fromTo[1]
```
xiaq, line 127: Follow the naming convention elsewhere and call this `wantTarget`.

pkg/testutil/testdir_test.go
```diff
@@ -124,21 +124,21 @@ func TestApplyDir_Symlinks(t *testing.T) {
...
 		fi, err := os.Lstat(symlink)
 		if err != nil || fi.Mode()&os.ModeSymlink == 0 {
 			t.Errorf("File %v isn't a symlink: err %v\n"+
 				"os.ModeSymlink %032b\n"+
 				"fi.Mode        %032b",
 				symlink, err, os.ModeSymlink, fi.Mode())
 		}
-		actual_target, err := filepath.EvalSymlinks(symlink)
-		if err != nil || expected_target != actual_target {
+		actualTarget, err := filepath.EvalSymlinks(symlink)
```
xiaq, line 135: Follow the naming convention elsewhere and call this `gotTarget`.

pkg/ui/key_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"src.elv.sh/pkg/eval/vals"
 )
 
-var kTests = []struct {
+var keyTests = []struct {
```
xiaq, line 10: The `golint` warning on this is hilarious, it thinks that the leading `k` in `kTests` is the private member marker used by some C++ and C# code...

But renaming this to `keyTests` is wrong because these are test cases of the `K` function. A lesser evil is to rename this to `KTests`.

pkg/cli/mode/mode.go
```diff
@@ -4,18 +4,22 @@ package mode
 
 import "src.elv.sh/pkg/ui"
 
-// ModeLine returns a text styled as a modeline.
-func ModeLine(content string, space bool) ui.Text {
+// modeLine returns a text styled as a modeline.
```
xiaq, line 7: godoc for unexported symbols should just elide the symbol:

> // Returns a ...

pkg/cli/mode/navigation.go
krader1961: Done

pkg/cli/mode/mode.go
krader1961: Done.

I too am not particularly bothered by some of golint's warnings. But the stuttering warning this change addresses does suggest an API problem; hence the "TODO" that I've removed in response to your feedback.

pkg/testutil/testdir_test.go
krader1961: Done, but I think consistently using `exp` (short for "expected") and `act` (short for "actual") unit test var name prefixes is preferable. Especially since, in this case, there is no existing use of `wantTarget` in the source code. 😸

pkg/testutil/testdir_test.go
krader1961: Done

pkg/ui/key_test.go
krader1961: Done.

pkg/cli/mode/mode.go
krader1961: I didn't know that. Done.

krader1961:
Sorry about the force-push but it was needed to deal with the merge conflicts created by other commits made to the master branch since I created this PR.

pkg/testutil/testdir_test.go
xiaq: The `want` and `got` prefixes are used consistently in all the other test files.

pkg/cli/mode/mode.go
xiaq: There isn't a strong consensus in the Go community, some people prefer writing out the symbol (in lower case), but I've settled on this pattern, since this avoids starting an English sentences with a lower-case letter. I'm probably influenced by Google's [C++ style guides](https://google.github.io/styleguide/cppguide.html#Function_Comments) on this matter.

pkg/cli/mode/mode.go
```diff
@@ -4,18 +4,20 @@ package mode
...
-// mode widget.
-func ModePrompt(content string, space bool) func() ui.Text {
-	p := ModeLine(content, space)
+func modePrompt(content string, space bool) func() ui.Text {
+	p := modeLine(content, space)
 	return func() ui.Text { return p }
 }
+
+// Prompt returns a callback suitable as the prompt in the codearea of a
+// mode widget. This var is solely for the benefit of pkg/edit/minibuf.
```
xiaq, line 22: > This var is solely for the benefit of pkg/edit/minibuf.

I would remove this because this function is generally useful for implementing modes outside this package.

PR #1266, opened 2021-03-16
Fix missing package in Docker build

FallenWarrior2k:
It appears that in Alpine 3.12, the man package was renamed to mandoc.
To prevent similar issues from re-occurring, I also added version constraints to the images referenced in the Dockerfile.


xiaq:
Thanks!

Issue #971, opened 2020-04-17
Document the limited vi editing mode


krader1961:
Someone asked on IRC how if Elvish supported vi mode editing. I couldn't find any mention in the documentation and greping the source didn't return any promising results. It turns out that doing

```
edit:insert:binding[Ctrl-'['] = $edit:command:start~
```

will enable a very limited vi editing mode. See also #730 for why the features are currently limited. The currently support vi commands can be found here: https://github.com/elves/elvish/blob/f30a2b6837a792727aa665a96229e1c0a1e71d91/edit/insert.go#L99

What's interesting about the current implementation is that it still leaves most (all?) of the emacs bindings in effect; e.g., [ctrl-A] moves to the start of the line in insert mode.

prologic:
It would be great if we could improve the vi-mode support for editing in Elvish 👌

Petruchio:
> It would be great if we could improve the vi-mode support for editing in Elvish 👌

I just wanted to "vote" for this, as a potential user.  I would very much like to leave Zsh, but this is integral to the way I continuously use the shell.

Issue #410, opened 2017-07-08
Bring back job control


xiaq:

krader1961:
I am a thumbs down on this proposal. I started programming in 1977 and eagerly embraced UNIX in the 1980's when interacting with a computer running UNIX was almost always over a serial connection and you rarely had more than a single connection (i.e., terminal session) to the system. Shell job control was essential in that world. Especially if you were interacting with a system via dial-up connection. Like I did as a Sequent Computer Systems support technician when interacting with its customer's systems in the 1980's and 1990's.

In 2020 the need for POSIX style job control is limited to situations like those mentioned above. And those situations are exceedingly rare today. It is exceedingly rare for an Elvish user to actually need POSIX shell job control and that will only decrease in the future. Also, given the quirks of multi-threaded programs (whether written in Go or C using POSIX threads) on UNIX like systems it is impossible to do this correctly 100% of the time.

zzamboni:
I agree with @krader1961 on this. Normally if I need a new session, I just open a new tab in my terminal. Even remotely (which, admittedly, I don't do much these days), I'd rather use `screen` or `tmux` to handle multiple sessions, rather than rely on job control.

While I think it would be cool to have, I think the effort to get it right would not be worth it. Like *cough* https://github.com/elves/elvish/issues/1039 *cough* ;)

andrewbaxter:
I use job control a lot for watching logs while running commands, I'd like this feature.

IcedQuinn:
I mostly use job control for

1. A GUI process i nontheless have to start in a shell and then disconnect and throw away, or,
2. Pausing a text editor (vim, kak) to run a command with the intent to return to the editor briefly after

I use more tabs/terminal windows for multiple ongoing things. But the quick flip between editor and running a command is what crashes me in to the wall when i looked at elvish.

PR #1270, opened 2021-03-21
Cmd duration

krader1961:
This PR implements both an interactive command duration variable and a callback for capturing that data. It does not include any unit tests. Once we've decided whether only one solution, or both, is worth merging I'll add appropriate unit tests. Personally I can't see much point in the callback mechanism and would use the $edit:command:duration var in my prompt rather than registering a callback to capture that data. Nonetheless, since I've implemented the callback it might be worth merging simply because there might be a use case I haven't thought of that justifies the added complexity.

### DIFF
--- a/0.16.0-release-notes.md
+++ b/0.16.0-release-notes.md
@@ -69,3 +69,9 @@ New features in the interactive editor:
 
 -   The editor now uses a DSL for filtering items in completion, history
     listing, location and navigation modes.
+
+-   A new `edit:after-command` hook that is invoked after each interactive
+    command line is run.
+
+-   A new `edit:command-duration` variable that is the number of seconds to
+    execute the most recent interactive command line.
--- a/pkg/edit/editor.go
+++ b/pkg/edit/editor.go
@@ -24,6 +24,10 @@ type Editor struct {
 
 	excMutex sync.RWMutex
 	excList  vals.List
+
+	// Maybe move this to another type that represents the REPL cycle as a whole, not just the
+	// read/edit portion represented by the Editor type.
+	AfterCommand []func(src parse.Source, duration float64, err error)
 }
 
 // An interface that wraps notifyf and notifyError. It is only implemented by
@@ -68,6 +72,7 @@ func NewEditor(tty cli.TTY, ev *eval.Evaler, st store.Store) *Editor {
 	initInstant(ed, ev, nb)
 	initMinibuf(ed, ev, nb)
 
+	initRepl(ed, ev, nb)
 	initBufferBuiltins(ed.app, nb)
 	initTTYBuiltins(ed.app, tty, nb)
 	initMiscBuiltins(ed.app, nb)
@@ -75,7 +80,7 @@ func NewEditor(tty cli.TTY, ev *eval.Evaler, st store.Store) *Editor {
 	initStoreAPI(ed.app, nb, hs)
 
 	ed.ns = nb.Ns()
-	evalDefaultBinding(ev, ed.ns)
+	initElvishState(ev, ed.ns)
 
 	return ed
 }
@@ -89,8 +94,9 @@ func initExceptionsAPI(ed *Editor, nb eval.NsBuilder) {
 	nb.Add("exceptions", vars.FromPtrWithMutex(&ed.excList, &ed.excMutex))
 }
 
-func evalDefaultBinding(ev *eval.Evaler, ns *eval.Ns) {
-	src := parse.Source{Name: "[default bindings]", Code: defaultBindingsElv}
+// Initialize the `edit` module by executing the pre-defined Elvish code for the module.
+func initElvishState(ev *eval.Evaler, ns *eval.Ns) {
+	src := parse.Source{Name: "[RC file]", Code: elvInit}
 	err := ev.Eval(src, eval.EvalCfg{Global: ns})
 	if err != nil {
 		panic(err)
@@ -100,6 +106,13 @@ func evalDefaultBinding(ev *eval.Evaler, ns *eval.Ns) {
 // ReadCode reads input from the user.
 func (ed *Editor) ReadCode() (string, error) {
 	return ed.app.ReadCode()
+}
+
+// RunAfterCommandHooks runs callbacks involving the interactive completion of a command line.
+func (ed *Editor) RunAfterCommandHooks(src parse.Source, duration float64, err error) {
+	for _, f := range ed.AfterCommand {
+		f(src, duration, err)
+	}
 }
 
 // Ns returns a namespace for manipulating the editor from Elvish code.
--- a/pkg/edit/repl.go
+++ b/pkg/edit/repl.go
@@ -1,0 +1,60 @@
+package edit
+
+// This file encapsulates functionality related to a complete REPL cycle. Such as capturing
+// information about the most recently executed interactive command.
+
+import (
+	"src.elv.sh/pkg/eval"
+	"src.elv.sh/pkg/eval/vals"
+	"src.elv.sh/pkg/eval/vars"
+	"src.elv.sh/pkg/parse"
+)
+
+//elvdoc:var after-command
+//
+// A list of functions to call after each interactive command completes. There is one pre-defined
+// function used to populate the [`$edit:command-duration`](./edit.html#editcommand-duration)
+// variable. Each function is called with a single [map](https://elv.sh/ref/language.html#map)
+// argument containing the following keys:
+//
+// * `code`: A string containing the code that was executed. If `is-file` is true this is the
+// content of the file that was sourced.
+//
+// * `duration`: A [floating-point number](https://elv.sh/ref/language.html#number) representing the
+// command execution duration in seconds.
+//
+// * `error`: An [exception](../ref/language.html#exception) object if the command terminated with
+// an exception, else [`$nil`](../ref/language.html#nil).
+//
+// * `name`: A string describing where the code originated; e.g., `[tty 1]` for the first
+// interactive REPL cycle. If `is-file` is true this is the path of the file that was sourced.
+//
+// * `is-file`: False if the code was entered interactively. At the moment this is only true when
+// this hook is run after sourcing the RC file.
+//
+// @cf edit:command-duration
+
+//elvdoc:var command-duration
+//
+// Duration, in seconds, of the most recent interactive command. This can be useful in your prompt
+// to provide feedback on how long a command took to run. The initial value of this variable is the
+// time to evaluate your *~/.elvish/rc.elv* script before printing the first prompt.
+//
+// @cf edit:after-command
+
+var commandDuration float64
+
+func initRepl(ed *Editor, ev *eval.Evaler, nb eval.NsBuilder) {
+	// TODO: Change this to a read-only var, possibly by introducing a vars.FromPtrReadonly
+	// function, to guard against attempts to modify the value from Elvish code.
+	nb.Add("command-duration", vars.FromPtr(&commandDuration))
+
+	hook := newListVar(vals.EmptyList)
+	nb["after-command"] = hook
+	ed.AfterCommand = append(ed.AfterCommand,
+		func(src parse.Source, duration float64, err error) {
+			m := vals.MakeMap("name", src.Name, "code", src.Code, "is-file", src.IsFile,
+				"duration", duration, "error", err)
+			callHooks(ev, "$<edit>:after-command", hook.Get().(vals.List), m)
+		})
+}
--- a/pkg/eval/compile_effect.go
+++ b/pkg/eval/compile_effect.go
@@ -75,11 +75,6 @@ func (op *pipelineOp) exec(fm *Frame) Exception {
 		fm.intCh = nil
 		fm.background = true
 		fm.Evaler.addNumBgJobs(1)
-
-		if fm.Evaler.Editor() != nil {
-			// TODO: Redirect output in interactive mode so that the line
-			// editor does not get messed up.
-		}
 	}
 
 	nforms := len(op.subops)
@@ -142,12 +137,7 @@ func (op *pipelineOp) exec(fm *Frame) Exception {
 				msg += ", errors = " + err.Error()
 			}
 			if fm.Evaler.getNotifyBgJobSuccess() || err != nil {
-				editor := fm.Evaler.Editor()
-				if editor != nil {
-					editor.Notify("%s", msg)
-				} else {
-					fm.ErrorFile().WriteString(msg + "\n")
-				}
+				fm.ErrorFile().WriteString(msg + "\n")
 			}
 		}()
 		return nil
--- a/pkg/shell/shell.go
+++ b/pkg/shell/shell.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/signal"
 	"strconv"
+	"time"
 
 	"src.elv.sh/pkg/cli/term"
 	"src.elv.sh/pkg/env"
@@ -62,11 +63,14 @@ func setupShell(fds [3]*os.File, p Paths, spawn bool) (*eval.Evaler, func()) {
 	}
 }
 
-func evalInTTY(ev *eval.Evaler, fds [3]*os.File, src parse.Source) error {
+func evalInTTY(ev *eval.Evaler, fds [3]*os.File, src parse.Source) (float64, error) {
+	start := time.Now()
 	ports, cleanup := eval.PortsFromFiles(fds, ev.ValuePrefix())
 	defer cleanup()
-	return ev.Eval(src, eval.EvalCfg{
+	err := ev.Eval(src, eval.EvalCfg{
 		Ports: ports, Interrupt: eval.ListenInterrupts, PutInFg: true})
+	end := time.Now()
+	return end.Sub(start).Seconds(), err
 }
 
 func incSHLVL() func() {
--- a/website/ref/edit.md
+++ b/website/ref/edit.md
@@ -431,28 +431,48 @@ hence that candidates for all completion types are matched by prefix.
 
 ## Hooks
 
-Hooks are functions that are executed at certain points in time. In Elvish, this
-functionality is provided by lists of functions.
+Hooks are functions that are executed at certain points in time. In Elvish this
+functionality is provided by variables that are a list of functions. Some hooks
+are populated with one or more functions when Elvish starts. In general you
+should append to a hook variable rather than assign a list of functions to it.
+That is, rather than doing `hook-var = [ []{ put 'I ran' } ]` you should do
+`hook-var = [ $@hook-var []{ put 'I ran' } ]`.
 
-There are current two hooks:
+These are the editor/REPL hooks:
 
--   `$edit:before-readline`, whose elements are called before the editor reads
-    code, with no arguments.
+-   [`$edit:before-readline`](https://elv.sh/ref/edit.html#editbefore-readline):
+    The functions are called before the editor runs. Each function is called
+    with no arguments.
 
--   `$edit:after-readline`, whose elements are called, after the editor reads
-    code, with a sole element -- the line just read.
+-   [`$edit:after-readline`](https://elv.sh/ref/edit.html#editafter-readline):
+    The functions are called after the editor accepts a command for execution.
+    Each function is called with a sole argument: the line just read.
+
+-   [`$edit:after-command`](https://elv.sh/ref/edit.html#editafter-command): The
+    functions are called after the shell executes the command you entered
+    (typically by pressing the `Enter` key). Each function is called with a sole
+    argument: a map that provides information about the executed command. This
+    hook is also called after your interactive RC file is executed and before
+    the first prompt is output.
 
 Example usage:
 
 ```elvish
 edit:before-readline = [{ echo 'going to read' }]
 edit:after-readline = [[line]{ echo 'just read '$line }]
+edit:after-command = [[m]{ echo 'command took '$m[duration]' seconds' }]
 ```
 
-Then every time you accept a chunk of code (and thus leaving the editor),
-`just read` followed by the code is printed; and at the very beginning of an
-Elvish session, or after a chunk of code is executed, `going to read` is
-printed.
+Given the above hooks...
+
+1. Every time you accept a chunk of code (normally by pressing Enter)
+   `just read` is printed.
+
+1. At the very beginning of an Elvish session, or after a chunk of code is
+   handled, `going to read` is printed.
+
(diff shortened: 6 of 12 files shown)


zzamboni:
The variable is useful for inserting the information in the prompt, but the callback might be more useful for producing notifications, like my [long-running-notifications](https://github.com/zzamboni/elvish-modules/blob/master/long-running-notifications.org) module does.

However, I don't much see the difference between the `edit:after-command` and the `edit:before-readline` callbacks. Wouldn't they run one after the other in any case?

krader1961:
> However, I don't much see the difference between the edit:after-command and the edit:before-readline callbacks. Wouldn't they run one after the other in any case?

Note that `edit:before-readline` currently has no arguments. That could be changed but conceptually doing something before a prompt is written is different from doing something after a command completes. Too, simply pressing Enter will result in `edit:before-readline` being run. Since no command was run you need additional complexity in the interpretation of its arguments to distinguish between command completion and no command being run. Your long running notification mechanism is a useful example of how the `edit:after-command` callback is useful. Finally, as a practical matter the way the code is currently structured passing the command completion data to the `edit:before-readline` callback is difficult. Which isn't really an argument for not doing it that way but suggests the existing callback isn't the appropriate way to expose the event.

krader1961:
Also, the temporal relationships of the callbacks to each other and other actions (e.g., writing the prompt) is something that could be better documented.

krader1961:
Since @zzamboni provided an example of how an "edit:after-command" callback is useful, and I still maintain that a command duration variable is useful even though technically redundant, I think both mechanisms should be implemented. Thus I'm going to go ahead and remove the leading hyphen from the var name (so it is no longer marked experimental) and add unit tests for both features.

zzamboni:
I'd say the variable is still useful for quick access to the information, e.g. to put it in the prompt.

xiaq:
I prefer to keep just the hook; the variable can be easily implemented in Elvish code with the hook mechanism.

pkg/cli/app.go
```diff
@@ -59,6 +61,7 @@ type app struct {
 	RPromptPersistent func() bool
 	BeforeReadline    []func()
 	AfterReadline     []func(string)
+	AfterCommand      []func(string, float64, error)
```
xiaq, line 64: The after-command hook does not belong in the `App` type; the `App` type is only responsible for the "R" (read) part in a REPL, while an after-command hook is only of concern for the whole REPL lifecycle.

In fact, the after-command hook doesn't belong in the `Editor` type either, because the `Editor` type is also only concerned about the R part in REPL. Eventually there should be a `REPL` type that takes care of REPL lifecycle stuff and that would be an appropriate home for the after-command hook.

But before that is done, the `Editor` type is a better temporary home than `App`, so move it there.

pkg/edit/command_api.go
```diff
@@ -3,18 +3,34 @@ package edit
...
-// Starts the command mode.
+// Start command mode.
+
+//elvdoc:var command:-duration
+//
+// Duration, in seconds, of the most recent interactive command. This can be useful in your prompt
+// (left or right) to provide feedback on how long a command took to run. The initial value of this
+// variable is the time to evaluate your *~/.elvish/rc.elv* script before issuing the first prompt.
+
+var CommandDuration float64
```
xiaq, line 19: The `edit:command:` namespace is actually for the vi-like command-mode, putting the duration variable there would be a mistake.

I also commented that I prefer to just keep the hook; the variable can be easily emulated in Elvish code.

pkg/edit/config_api.go
```diff
@@ -60,6 +61,20 @@ func initAfterReadline(appSpec *cli.AppSpec, ev *eval.Evaler, nb eval.NsBuilder)
 	})
 }
 
+//elvdoc:var after-command
+//
+// A list of functions to call after each interactive command completes. Each function is
+// called with a single string argument containing the code that has been read.
```
xiaq, line 67: > // Each function is
> // called with a single string argument containing the code that has been read.

Update

pkg/edit/config_api.go
```diff
@@ -60,6 +61,20 @@ func initAfterReadline(appSpec *cli.AppSpec, ev *eval.Evaler, nb eval.NsBuilder)
...
+
+func initAfterCommand(appSpec *cli.AppSpec, ev *eval.Evaler, nb eval.NsBuilder) {
+	hook := newListVar(vals.EmptyList)
+	nb["after-command"] = hook
+	appSpec.AfterCommand = append(appSpec.AfterCommand,
+		func(line string, duration float64, err error) {
+		callHooks(ev, "$<edit>:after-command", hook.Get().(vals.List), line, duration, err)
```
xiaq, line 74: Instead of calling the after-command hooks with 3 arguments, better to call it with one map argument [&executed-code=... &duration=... &error=...]

pkg/edit/editor.go
```diff
@@ -102,6 +102,11 @@ func (ed *Editor) ReadCode() (string, error) {
 	return ed.app.ReadCode()
 }
 
+// CaptureResult runs callbacks involving the interactive completion of a command line.
+func (ed *Editor) CaptureResult(line string, duration float64, err error) {
```
xiaq, line 106: `RunAfterCommandHooks` sounds much more straightforward.

pkg/shell/shell.go
```diff
@@ -62,11 +63,14 @@ func setupShell(fds [3]*os.File, p Paths, spawn bool) (*eval.Evaler, func()) {
 	}
 }
 
-func evalInTTY(ev *eval.Evaler, fds [3]*os.File, src parse.Source) error {
+func evalInTTY(ev *eval.Evaler, fds [3]*os.File, src parse.Source) (float64, error) {
```
xiaq, line 66: I don't see a good reason that `evalInTTY` should return the duration; this can easily be done by the caller:

```go
start := time.Now()
err := evalInTTY(...)
end := time.Now()
duration := end.Sub(start)
```

krader1961:
> I prefer to keep just the hook; the variable can be easily implemented in Elvish code with the hook mechanism.

As I noted in an earlier comment the truth value of that assertion depends on how you define "easily". I think the vast majority of users who simply want to include the last command duration in their prompt will not find the hook an "easy" way to get that information.

pkg/edit/command_api.go
krader1961: Ahhh, yes, I simply latched onto the "command" namespace name. Part of the problem is this namespace has not documentation. See issue #971. I'll move it to the `edit` namespace pending further discussion about whether it is redundant given the callback.

pkg/edit/config_api.go
krader1961: Good idea since it future proofs the API against additional attributes. Will change. However, the cost of this mechanism is another reason to retain a `$edit:command-duration` variable as a close to zero cost, both edit and run time, for users that simply want to display the duration in their prompt.

pkg/edit/editor.go
krader1961: Agreed

pkg/shell/shell.go
krader1961: That's reasonable, but note that the reason I did it this way is there are three callers of `evalInTTY` and two of them care about the duration. The third, in *pkg/shell/script.go*, might want to expose the duration in the future. It seems redundant, and risky, to have those callers all compute the duration in potentially incompatible fashion. Do you still want me to move the duration calculation to each call site?

pkg/edit/config_api.go
krader1961: Done

pkg/cli/app.go
krader1961: Done. And I'll note that moving the implementation also reduced the footprint of this change since it no longer touches *pkg/cli/app.go*. But that raises the question why this belongs in the Editor implementation but the before and after readline callbacks do not.

krader1961:
> I prefer to keep just the hook; the variable can be easily implemented in Elvish code with the hook mechanism.

Hmmm, I just noticed that this could be implemented as default behavior by adding appropriate statements to the Elvish code in *pkg/edit/default_bindings.go*. I want to sleep on this but I'm still not convinced the near zero cost of the explicit module variable should be substituted with Elvish code that adds significant cost to the most common use case.

Also, I haven't pushed an update to this P.R. because adding unit tests is proving more difficult than I expected. The `shell.Interact()` function isn't run by the editor unit tests. So the code I added to that function to set `$edit:cmd-duration` and run the `$edit:after-command` hooks isn't exercised by the edit unit tests.

xiaq:
> > I prefer to keep just the hook; the variable can be easily implemented in Elvish code with the hook mechanism.
> 
> Hmmm, I just noticed that this could be implemented as default behavior by adding appropriate statements to the Elvish code in _pkg/edit/default_bindings.go_. I want to sleep on this but I'm still not convinced the near zero cost of the explicit module variable should be substituted with Elvish code that adds significant cost to the most common use case.
> 

Implementing such a variable with Elvish code is a good idea. If you do that, default_bindings.go should also be renamed to something like edit_init.go.

By "significant cost" do you mean the overhead of executing Elvish code versus executing Go code? Elvish is much less performant than Go, but (1) users perhaps already have some code in their `rc.elv`, and (2) eventually the editor implementation should be moved more and more from Go into Elvish anyway. If the performance of Elvish becomes a bottleneck for (2), the solution is to speed up Elvish, not keeping things in Go.

> Also, I haven't pushed an update to this P.R. because adding unit tests is proving more difficult than I expected. The `shell.Interact()` function isn't run by the editor unit tests. So the code I added to that function to set `$edit:cmd-duration` and run the `$edit:after-command` hooks isn't exercised by the edit unit tests.

That's exactly why there needs to be an additional REPL abstraction (I had another comment about this) which will be correct place to test this.

krader1961:
> Implementing such a variable with Elvish code is a good idea.

That's trivial and, I agree, a good idea in general. Spefically, it is better to implement as much functionality in Elvish code as is possible; as opposed to implementing the same logic in Go. However, note that we still need to define the underlying var in the Go source. Which splits the implementation.

krader1961:
I just realized that `$edit:command-duration` cannot be implemented by a predefined `$edit:after-command` hook if we want to capture the cost of running the rc script (i.e., the shell.sourceRC function). That capability alone, IMHO, justifies the two lines added to the Go source to support that Elvish namespace variable. I really like seeing the time taken to run my *~/.elvish/rc.elv* script in the first prompt.

krader1961:
> That's exactly why there needs to be an additional REPL abstraction (I had another comment about this) which will be correct place to test this.

I'm not convinced that is needed or justified. At present the `shell.Interact()` function is effectively the REPL abstraction. That could be refactored to an explicit REPL abstraction but it's not clear that effort is justified unless we intend to introduce something like a `repl:` namespace rather than keep `after-command` and `command-duration` vars in the `edit:` namespace. It doesn't seem to me to make much difference where the updates of those vars occurs.

Too, we can add unit tests of these two vars to `TestInteract_SingleCommand()` by changing the `progtest.Setup()` function to use a real tty rather than a pipe. Thus causing the `shell.Interact()` function to construct a normal editor rather than a minimal editor.

xiaq:
> I just realized that `$edit:command-duration` cannot be implemented by a predefined `$edit:after-command` hook if we want to capture the cost of running the rc script (i.e., the shell.sourceRC function). That capability alone, IMHO, justifies the two lines added to the Go source to support that Elvish namespace variable. I really like seeing the time taken to run my _~/.elvish/rc.elv_ script in the first prompt.

My preferred solution here is to make the after-command hook run for rc.elv too.

> > That's exactly why there needs to be an additional REPL abstraction (I had another comment about this) which will be correct place to test this.
> 
> I'm not convinced that is needed or justified. At present the `shell.Interact()` function is effectively the REPL abstraction. That could be refactored to an explicit REPL abstraction but it's not clear that effort is justified unless we intend to introduce something like a `repl:` namespace rather than keep `after-command` and `command-duration` vars in the `edit:` namespace. It doesn't seem to me to make much difference where the updates of those vars occurs.
> 
> Too, we can add unit tests of these two vars to `TestInteract_SingleCommand()` by changing the `progtest.Setup()` function to use a real tty rather than a pipe. Thus causing the `shell.Interact()` function to construct a normal editor rather than a minimal editor.

We don't actually disagree; if you make the necessary changes to make `shell.Interact` easy to test, then it is a proper REPL abstraction. I use the word "abstraction" pretty loosely.

krader1961:
> My preferred solution here is to make the after-command hook run for rc.elv too.

Okay, I can do that. TBD is whether that is a net increase in complexity.

> We don't actually disagree; if you make the necessary changes to make shell.Interact easy to test....

Given that we already depend on `github.com/creack/pty` it should be a straightforward enhancement to the interactive shell test framework. Of course, the fact the current framework uses pipes means "dragons be here" is operative. Which might mean more work than expected to change from pipes to ptys. Still, there is no reason that should not be possible and a net improvement for test coverage.

krader1961:
@xiaq: PTAL. I have more refactoring in mind. For example, the description of hooks should be moved from the edit page to the language page and augmented to include mention of the other hooks (e.g., `builtin:after-chdir`. I'll also start working on changing `progtest.Setup()` to use a pty rather than pipes so that a `edit:after-command` unit test can be written.

pkg/edit/repl.go
```diff
@@ -0,0 +1,51 @@
...
+
+//elvdoc:var command-duration
+//
+// Duration, in seconds, of the most recent interactive command. This can be useful in your prompt
+// to provide feedback on how long a command took to run. The initial value of this variable is the
+// time to evaluate your *~/.elvish/rc.elv* script before printing the first prompt.
+//
+// @cf edit:after-command
+
+var CommandDuration float64
```
xiaq, line 37: This should not be a global variable.

pkg/edit/repl.go
```diff
@@ -0,0 +1,51 @@
...
+	"src.elv.sh/pkg/eval"
+	"src.elv.sh/pkg/eval/vals"
+	"src.elv.sh/pkg/eval/vars"
+)
+
+//elvdoc:var after-command
+//
+// A list of functions to call after each interactive command completes. There is one pre-defined
+// function used to populate the [`$edit:command-duration`](./edit.html#editcommand-duration)
+// variable. Each function is called with a single pseudo-map argument containing the following
```
xiaq, line 16: > a single pseudo-map argument

The implementation uses a real map

pkg/eval/compile_effect.go
```diff
@@ -142,12 +138,7 @@ func (op *pipelineOp) exec(fm *Frame) Exception {
 				msg += ", errors = " + err.Error()
 			}
 			if fm.Evaler.getNotifyBgJobSuccess() || err != nil {
-				editor := fm.Evaler.Editor()
-				if editor != nil {
-					editor.Notify("%s", msg)
-				} else {
-					fm.ErrorFile().WriteString(msg + "\n")
-				}
+				fm.ErrorFile().WriteString(msg + "\n")
```
xiaq, line 141: You were probably trying to clean up the evaler so that it no longer needs to contain a reference to the editor; cleanups should not be done at the expense of real features. Removing the notification logic is a regression.

pkg/eval/compile_effect.go
```diff
@@ -75,11 +75,7 @@ func (op *pipelineOp) exec(fm *Frame) Exception {
 		fm.intCh = nil
 		fm.background = true
 		fm.Evaler.addNumBgJobs(1)
-
-		if fm.Evaler.Editor() != nil {
-			// TODO: Redirect output in interactive mode so that the line
-			// editor does not get messed up.
-		}
+		// TODO: Redirect output in interactive mode so that the line editor does not get messed up.
```
xiaq, line 78: Not sure why this needs to be changed.

pkg/edit/editor.go
```diff
@@ -102,6 +108,13 @@ func (ed *Editor) ReadCode() (string, error) {
 	return ed.app.ReadCode()
 }
 
+// RunAfterCommandHooks runs callbacks involving the interactive completion of a command line.
+func (ed *Editor) RunAfterCommandHooks(command string, duration float64, err error) {
```
xiaq, line 112: Instead of passing the command as a string, pass it as a `parse.Source`, a type that is already used to represent a piece of source code to execute.

xiaq (changes requested):

0.16.0-release-notes.md
```diff
@@ -59,3 +59,9 @@ New features in the interactive editor:
 
 -   The editor now uses a DSL for filtering items in completion, history
     listing, location and navigation modes.
+
+-   A new `edit:after-command` callback that is invoked after each interactive
+    command line is run.
```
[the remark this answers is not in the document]
krader1961: Done: s/callback/hook/

pkg/edit/repl.go
krader1961: Agreed. That it is global is an artifact of a much earlier version of this change.

pkg/edit/repl.go
krader1961: Good catch. Originally I intended to use a struct map but ultimately decided a real map was simpler and potentially useful to allow the hook to modify the content.

pkg/eval/compile_effect.go
krader1961: I decided to simply remove the TODO.

pkg/eval/compile_effect.go
krader1961: There is currently no way for `fm.Evaler.Editor()` to return a non-nil value. Both in actual use and unit tests. The presence of this unused interface method greatly complicated the implementation of this change for zero benefit since the added functions were simply stubs that did nothing.

pkg/edit/editor.go
krader1961: /facepalm and Homer Simpson "Doh!" Fixed. Which resulted in my deciding to also expose the "name" and "is-file" attributes as well as "code" (formerly "command").

pkg/edit/repl.go
krader1961: I just realized making this a real map isn't particularly useful given the [semantics of mutating maps](https://elv.sh/ref/language.html#set) and [assigning to containers](https://elv.sh/learn/unique-semantics.html#assignment-semantics). I was thinking that making it a real map would allow an earlier hook to affect the data seen by a later hook. It's arguably a good thing that isn't possible from a security perspective. I'm going to leave it a real map for the moment. It is easy enough to make it a lighter weight struct map.

pkg/eval/compile_effect.go
krader1961: Also, it's not at all obvious that we should, or event want to, automatically redirect/capture the output of background jobs. The POSIX standard handles this via tty job control; e.g., `SIGTTIN` and `SIGTTOU` sent to a background job that is not in control of the tty. There is discussion in issue #410 about whether Elvish should fully implement POSIX job control. Regardless, I think there is no good argument for retaining the existing, unused, notification mechanism since doing so increases the complexity of this change for zero benefit today, and probably tomorrow. This background job notification mechanism can always be reintroduced if, and when, there is consensus it is needed and how it should work.

krader1961:
@xiaq, this version of the P.R. addresses all your feedback other than the removal of the `*Evaler.Editor` method. As I explained earlier that never returns a non-nil value and keeping that API greatly complicates this change for no good reason. Not to mention these lines in the `type Evaler` definition which implies the existing uses should be replaced with a different mechanism:

```
// TODO: Remove these dependency by providing more general extension points.
daemonClient daemon.Client
editor       Editor
```

We might want to add "TODO" comments where `fm.Evaler.Editor()` used to be used but it's far from obvious that would be useful. It seems to me that those two cases should be addressed by open issues explaining how the current behavior is sub-optimal and what might be better behavior.

pkg/eval/compile_effect.go
xiaq: Huh, OK, I didn't realise this functionality has been broken. While I'd prefer you keep it since I do plan to fix the feature, that does mean removing it is harmless now.

website/ref/edit.md
```diff
@@ -431,28 +431,48 @@ hence that candidates for all completion types are matched by prefix.
...
-There are current two hooks:
-
--   `$edit:before-readline`, whose elements are called before the editor reads
-    code, with no arguments.
-
--   `$edit:after-readline`, whose elements are called, after the editor reads
-    code, with a sole element -- the line just read.
+Hooks are functions that are executed at certain points in time. In Elvish this
+functionality is provided by variables that are a list of functions. Some hooks
+are populated with one or more functions when Elvish starts. In general you
```
xiaq, line 436: > In general...

Put non-normative text after a "NOTE".

pkg/edit/repl.go
```diff
@@ -0,0 +1,60 @@
...
+
+//elvdoc:var after-command
+//
+// A list of functions to call after each interactive command completes. There is one pre-defined
+// function used to populate the [`$edit:command-duration`](./edit.html#editcommand-duration)
+// variable. Each function is called with a single [map](https://elv.sh/ref/language.html#map)
+// argument containing the following keys:
+//
+// * `code`: A string containing the code that was executed. If `is-file` is true this is the
+// content of the file that was sourced.
```
xiaq, line 21: Better just expose the `Source` struct, which is itself a pseudo-map and can be introspected (https://elv.sh/ref/builtin.html#src)

pkg/edit/repl.go
```diff
@@ -0,0 +1,60 @@
...
+
+//elvdoc:var command-duration
+//
+// Duration, in seconds, of the most recent interactive command. This can be useful in your prompt
+// to provide feedback on how long a command took to run. The initial value of this variable is the
+// time to evaluate your *~/.elvish/rc.elv* script before printing the first prompt.
+//
+// @cf edit:after-command
+
+var commandDuration float64
```
xiaq, line 45: You de-exported this, but it's still a global. This should be allocated locally within `initRepl` for safety - this is a pattern used in many other `init*` functions.

pkg/edit/repl.go
```diff
@@ -0,0 +1,60 @@
...
+
+func initRepl(ed *Editor, ev *eval.Evaler, nb eval.NsBuilder) {
+	// TODO: Change this to a read-only var, possibly by introducing a vars.FromPtrReadonly
+	// function, to guard against attempts to modify the value from Elvish code.
+	nb.Add("command-duration", vars.FromPtr(&commandDuration))
+
+	hook := newListVar(vals.EmptyList)
```
xiaq, line 52: Even though there is only one hook for now, naming the `hook` suggests that this is "the" hook for the REPL, which is not true; `afterCommandsHook` is better.

xiaq (approved):
A few minor stuff I'll fix after merging this.

PR #1271, opened 2021-03-22
Remove invalid font-strecth CSS attribute

krader1961:
The `font-strecth` CSS attribute is not valid. It should be spelled
`font-stretch`. It is also not needed since the default is "normal". So
just remove that CSS attribute.


xiaq:
Thanks for finding the typo.

> It is also not needed since the default is "normal".

No, in a `@font-face` declaration the default of `font-stretch` is that specified by the underlying font data. It's more robust to specify this explicitly in the CSS, especially considering that all the other [basic properties](https://drafts.csswg.org/css-fonts-3/#basic-font-props) are specified in the CSS.

xiaq:
Fixed in 4e89c374d94561077f8d4173dca59f1997e917a6.

PR #1273, opened 2021-03-27
Include name in error msg setting a readonly var

krader1961:
When attempting to update a read-only var return an error struct rather
than a simple string (i.e., a Go `error` type).  This makes it possible
to include the var name in the error message.  This builds on commit
a33ecb2d that highlights the offending var name in the stack trace but
does not include the var name in the error message. With this change the
error message includes the offending var name.

Related #255


pkg/eval/ns_test.go
```diff
@@ -26,6 +26,6 @@ func TestNs(t *testing.T) {
 
 func TestBuiltinFunctionsReadOnly(t *testing.T) {
 	Test(t,
-		That("return~ = { }").Throws(vars.ErrSetReadOnlyVar, "return~"),
+		That("return~ = { }").Throws(ErrorWithType(errs.SetReadOnlyVar{}), "return~"),
```
xiaq, line 29: This can test for the value of the error `errs.SetReadOnlyVar{VarName: "return~"}`

pkg/eval/frame.go
```diff
@@ -210,11 +211,13 @@ func (fm *Frame) errorp(r diag.Ranger, e error) Exception {
 		return nil
 	case Exception:
 		return e
+	case errs.SetReadOnlyVar:
+		ctx := diag.NewContext(fm.srcMeta.Name, fm.srcMeta.Code, r)
+		e.VarName = ctx.RelevantString()
+		return &exception{e, &StackTrace{Head: ctx, Next: fm.traceback}}
```
xiaq, line 217: This can be deduplicated with the `default` branch below by introducing another type assertion.

xiaq (approved):
A few minor issues I'll will fix after merging.

PR #1274, opened 2021-03-29
Initialize a lambda var to $nop~

### LINKED ISSUE #1248: x~ cannot be set to $nil
Since initial value of `x~` is `$nil`, I would assume that `$nil` can be assigned to this variable any time.
```
[~]─> var x~
[~]─> put $x~
▶ $nil
[~]─> set x~ = $nil
Exception: wrong type: need nil, got nil
[tty 3], line 1: set x~ = $nil
[~]─>
```

krader1961:
You cannot assign $nil to a lambda var so that should not be the initial
value of `var x~`. Instead, assign $builtin:nop~.

Fixes #1248

### DIFF
--- a/0.16.0-release-notes.md
+++ b/0.16.0-release-notes.md
@@ -11,6 +11,10 @@ This is the draft release notes for 0.16.0. It is scheduled to be released on
     Instead, the history listing mode now applies smart-case matching by
     default.
 
+-   Declaring a variable with a `~` suffix, without an explicit initial value,
+    now defaults to $builtin:nop rather than $builtin:nil
+    ([#1248](https://b.elv.sh/1248)).
+
 # Notable bugfixes
 
 -   The `path:is-dir` and `path:is-regular` commands no longer follow symlinks,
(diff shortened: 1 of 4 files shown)


krader1961:
I force pushed to replace "lambda var" with "command var" in the change description.  This is an area of Elvish terminology that needs clarification.

0.16.0-release-notes.md
```diff
@@ -11,6 +11,10 @@ This is the draft release notes for 0.16.0. It is scheduled to be released on
     Instead, the history listing mode now applies smart-case matching by
     default.
 
+-   Declaring a variable with a `~` suffix, without an explicit initial value,
+    now defaults to $builtin:nop rather than $builtin:nil
```
xiaq, line 15: $builtin:nop~

Also, since the `builtin:` namespace no longer exists without an explicit `use`, it's better to just say "builtin" without using it as a variable name qualifier; same for doc and comments below.

pkg/eval/builtin_special_test.go
```diff
@@ -24,6 +24,9 @@ func TestVar(t *testing.T) {
 		That("var 'a/b'", "put $'a/b'").Puts(nil),
 		// Declaring one variable whose name ends in ":".
 		That("var a:").DoesNothing(),
+		// Declaring a variable whose name ends in "~" has default value $builtin:nop rather than
+		// $builtin:nil. See https://github.com/elves/elvish/issues/1248.
```
xiaq, line 28: No need to refer to the old behavior here, just say that it's equivalent to the `nop` function

xiaq (approved):

PR #1277, opened 2021-04-01
Add option to reinstate old path:is-* behavior

krader1961:
This introduces a `&follow-symlink` option to the two existing `path:is-`
commands. It is sometimes useful to know if a given path resolves to a
particular filesystem object type even when the final component of the
path is a symbolic link.

Related #1236

### DIFF
--- a/pkg/eval/mods/path/path_test.go
+++ b/pkg/eval/mods/path/path_test.go
@@ -13,12 +13,15 @@ import (
 
 var testDir = testutil.Dir{
 	"d1": testutil.Dir{
-		"f": testutil.Symlink{Target: filepath.Join("d2", "f")},
 		"d2": testutil.Dir{
 			"empty": "",
 			"f":     "",
 			"g":     testutil.Symlink{Target: "f"},
 		},
+		// These symlink definitions must occur after definition of their targets because on Windows
+		// the behavior of the os.Symlink function depends on the nature of the target; unlike UNIX.
+		"d": testutil.Symlink{"d2"},
+		"f": testutil.Symlink{filepath.Join("d2", "f")},
 	},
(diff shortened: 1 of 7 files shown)


pkg/eval/mods/path/path_test.go
```diff
@@ -59,9 +63,20 @@ func TestPath(t *testing.T) {
 		That(`path:is-dir a/b/s`).Puts(false),
 		That(`path:is-dir `+tmpdir).Puts(true),
 		That(`path:is-dir s1`).Puts(false),
+		That(`path:is-dir d1/d`).Puts(false),
+		// WTF: This fails on Windows 10.
+		That(`path:is-dir d1/d &follow-symlink`).Puts(true),
```
xiaq, line 68: Please leave "WTF" out of the code. If you can't figure out why this fails in Windows 10, leave this line commented out and add a TODO to fix this test on Windows.

krader1961: The whole point of a "WTF" comment is to remind me of something that needs to be resolved before anyone else sees the change. All of the tests pass on Windows. So I need to figure out if I botched a `git master rebase`, simply overlooked the need to delete that comment, or something else. Will push a corrected change soon.

krader1961: It looks like removing this "WTF" comment was an oversight. I fixed the reason for this test failure on Windows but didn't remove this comment. Lines 19 to 22 of the change to this file explains why symlinks have to be created after the referent on Windows10.

krader1961:
Obsolete "WTF" comment removed, rebased on master, and I made some wording tweaks. PTAL.

krader1961:
Another force push to resolve the conflict with other changes by @xiaq to the *0.16.0-release-notes.md* file.

PR #1280, opened 2021-04-04
Remove unnecessary strconv.Itoa calls

krader1961:


krader1961:
The CI failures on macos don't have anything to do with this change. Those failures are an artifact of the CI environment having problems independent of this change.

xiaq:
These strings are becoming ints in the `num` branch, the change here will soon be obsolete. Thanks for doing the cleanup but to avoid merge conflict I will just close this.

PR #1281, opened 2021-04-04
file module now building and has 4 function added

supreetsingh10:
file_test.go has been made, the functions in builtin_fn_io.go relating to files have been commented out and marked deprecated.

### DIFF
--- a/0.16.0-release-notes.md
+++ b/0.16.0-release-notes.md
@@ -32,6 +32,9 @@ New features in the interactive editor:
 -   The editor now uses a DSL for filtering items in completion, history
     listing, location and navigation modes.
    
--   A new `file:` builtin module to work with files. Comes with functions such
-    as file:fopen, file:close and more
+-   A new `file:` module to work with files.
 
+# Deprecated features
+
+-   The `builtin:fopen` and `builtin:fclose` commands are deprecated. Use 
+    the equivalent commands in the new `file:` module.  
--- a/0.16.0-release-notes.md
+++ b/0.16.0-release-notes.md
@@ -31,3 +31,7 @@ New features in the interactive editor:
 
 -   The editor now uses a DSL for filtering items in completion, history
     listing, location and navigation modes.
+   
+-   A new `file:` builtin module to work with files. Comes with functions such
+    as file:fopen, file:close and more
+
--- a/pkg/eval/builtin_fn_io.go
+++ b/pkg/eval/builtin_fn_io.go
@@ -770,10 +770,10 @@ func toJSON(fm *Frame, inputs Inputs) error {
 // a file.
 // ~> fclose $f
 // ```
+// This function is deprecated; use [file:open](./file.html#open) instead.
 //
 // @cf fclose
 
-// Deprecated written in file module instead
 func fopen(name string) (vals.File, error) {
 	// TODO support opening files for writing etc as well.
 	return os.Open(name)
@@ -787,9 +787,10 @@ func fopen(name string) (vals.File, error) {
 //
 // Close a file opened with `fopen`.
 //
+// This function is deprecated; use [file:close](./file.html#close) instead.
+//
 // @cf fopen
 
-// Deprecated, written in file module instead.
 func fclose(f vals.File) error {
 	return f.Close()
 }
@@ -828,7 +829,6 @@ func fclose(f vals.File) error {
 //
 // @cf prclose pwclose
 
-// Deprecated written in file.go
 func pipe() (vals.Pipe, error) {
 	r, w, err := os.Pipe()
 	return vals.NewPipe(r, w), err
@@ -844,7 +844,6 @@ func pipe() (vals.Pipe, error) {
 //
 // @cf pwclose pipe
 
-// Deprecated written in file.go
 func prclose(p vals.Pipe) error {
 	return p.ReadEnd.Close()
 }
--- a/pkg/eval/builtin_fn_io.go
+++ b/pkg/eval/builtin_fn_io.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"os"
+	//	"os"
 	"strconv"
 	"strings"
 
@@ -51,11 +51,12 @@ func init() {
 		"to-lines": toLines,
 		"to-json":  toJSON,
 
+		// Deprecated written in file.go instead.
 		// File and pipe
-		"fopen":   fopen,
-		"fclose":  fclose,
-		"pipe":    pipe,
-		"prclose": prclose,
+		// "fopen":   fopen,
+		// "fclose":  fclose,
+		// "pipe":    pipe,
+		// "prclose": prclose,
 		"pwclose": pwclose,
 	})
 }
@@ -772,10 +773,11 @@ func toJSON(fm *Frame, inputs Inputs) error {
 //
 // @cf fclose
 
-func fopen(name string) (vals.File, error) {
-	// TODO support opening files for writing etc as well.
-	return os.Open(name)
-}
+// Deprecated written in file module instead
+// func fopen(name string) (vals.File, error) {
+// TODO support opening files for writing etc as well.
+//return os.Open(name)
+// }
 
 //elvdoc:fn fclose
 //
@@ -787,9 +789,10 @@ func fopen(name string) (vals.File, error) {
 //
 // @cf fopen
 
-func fclose(f vals.File) error {
-	return f.Close()
-}
+// Deprecated, written in file module instead.
+// func fclose(f vals.File) error {
+// return f.Close()
+// }
 
 //elvdoc:fn pipe
 //
@@ -825,10 +828,11 @@ func fclose(f vals.File) error {
 //
 // @cf prclose pwclose
 
-func pipe() (vals.Pipe, error) {
-	r, w, err := os.Pipe()
-	return vals.NewPipe(r, w), err
-}
+// Deprecated written in file.go
+// func pipe() (vals.Pipe, error) {
+// r, w, err := os.Pipe()
+// return vals.NewPipe(r, w), err
+// }
 
 //elvdoc:fn prclose
 //
@@ -840,9 +844,10 @@ func pipe() (vals.Pipe, error) {
 //
 // @cf pwclose pipe
 
-func prclose(p vals.Pipe) error {
-	return p.ReadEnd.Close()
-}
+// Deprecated written in file.go
+// func prclose(p vals.Pipe) error {
+// return p.ReadEnd.Close()
+// }
 
 //elvdoc:fn pwclose
 //
--- a/pkg/eval/builtin_fn_io.go
+++ b/pkg/eval/builtin_fn_io.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	//	"os"
+	"os"
 	"strconv"
 	"strings"
 
@@ -53,10 +53,10 @@ func init() {
 
 		// Deprecated written in file.go instead.
 		// File and pipe
-		// "fopen":   fopen,
-		// "fclose":  fclose,
-		// "pipe":    pipe,
-		// "prclose": prclose,
+		"fopen":   fopen,
+		"fclose":  fclose,
+		"pipe":    pipe,
+		"prclose": prclose,
 		"pwclose": pwclose,
 	})
 }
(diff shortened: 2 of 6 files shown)


pkg/eval/builtin_fn_io.go
```diff
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"os"
+	//	"os"
```
krader1961: I know your subsequent commit reinstated this but in general you should not leave this type of comment in the source. Its presence means that either it is no longer needed and you forgot to remove this line before sending it out for review, or it is needed and the change is incomplete. What I do when introducing this type of comment is include "WTF" in the text to flag something needing more attention: `// WTF "os"`. In the four decades I've been programming I can count on the fingers of one hand the number of times "WTF" legitimately appeared in a project. I always do a "grep WTF **" to check for the presence of such comments to verify I have resolved all such comments before creating a pull-request.

P.S., I am not suggesting you handle this situation the way I do it. You should use whatever approach helps you avoid sending a change for review while it still has unresolved issues as indicated by this type of comment.

pkg/eval/builtin_fn_io.go
```diff
@@ -787,9 +789,10 @@ func fopen(name string) (vals.File, error) {
 //
 // @cf fopen
 
-func fclose(f vals.File) error {
-	return f.Close()
-}
+// Deprecated, written in file module instead.
+// func fclose(f vals.File) error {
```
krader1961: Do not comment out a unit test for a deprecated feature. The tests should continue to run until the deprecated feature is actually removed.

pkg/eval/mods/file/file.go
```diff
@@ -0,0 +1,100 @@
...
+// files are opened for reading using fopen
+// Will not accept string name of the file
+// ```elvish
+// fclose $fileptr
+// ```
+func fclose(f vals.File) error {
+	return f.Close()
+}
+
+//elvdoc:fn pipe
```
krader1961, line 42: I think you're attempting to do too much with this section of the change and would recommend omitting the `pipe`, `prclose` and `pwclose` commands from this module. They can be introduced by a subsequent change.

pkg/eval/mods/file/file.go
```diff
@@ -0,0 +1,100 @@
...
+
+package file
+
+import (
+	"os"
+	"src.elv.sh/pkg/eval"
+	"src.elv.sh/pkg/eval/vals"
+)
+
+// elvdoc:fn fopen
```
krader1961, line 12: In this module the "f" prefix is redundant. It should just be "open"; i.e., "file:open". Same for the function definition and the entry in the namespace table. Similarly for "fclose".

pkg/eval/builtin_fn_io.go
```diff
@@ -787,9 +789,10 @@ func fopen(name string) (vals.File, error) {
 //
 // @cf fopen
 
-func fclose(f vals.File) error {
-	return f.Close()
-}
+// Deprecated, written in file module instead.
```
krader1961, line 792: The deprecation should be part of the documentation. See, for example:

https://github.com/elves/elvish/blob/29034c20e9a30e9826c9f9aeac309b4d519db08e/pkg/eval/builtin_fn_str.go#L177-L183

You also need to add appropriate lines to the `checkDeprecatedBuiltin` function in *pkg/eval/compiler.go*.

krader1961 (changes requested):
In addition to the comments about specific aspects of this P.R. I would also recommend doing a squash-merge of the three commits on your system. Then force push an update (e.g., "git push -f") so that there is a single commit being reviewed.

pkg/eval/builtin_fn_io.go
supreetsingh10: This is a piece of great advice thank you for this.

supreetsingh10:
I am working on all the changes you requested. Thank you for the help.

pkg/eval/mods/file/file_test.go
```diff
@@ -13,12 +13,5 @@ func TestFile(t *testing.T) {
 	}
 	evaltest.TestWithSetup(t, setup,
 		evaltest.That(`file:fopen`).Throws(evaltest.AnyError),
-		evaltest.That(`f = (file:fopen file.go)`).Puts(),
```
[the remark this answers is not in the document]
supreetsingh10: The open and close functions were in ```buitin_fn_io.go```, at first I tried looking for their unit tests in ```builtin_fn_io_test.go``` but they were not there. I ran the ```grep  "TestFopen" -r -i  * ```, I did not get any results. So I made their tests, using the ```str_test.go```.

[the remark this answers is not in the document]
krader1961: Note that most tests of specific commands won't have a corresponding `Test...` function -- instead  related commands are aggregated into a single unit test. See, for example, *pkg/eval/mods/str/str_test.go*.

I was surprised to find only a single test of `fopen` and `fclose`, and the purpose of that test is to confirm a "file" object can be used as the source of an input redirect:

https://github.com/elves/elvish/blob/8a7073a20fb224f0a202f5c7aaf92bf7bc56d47d/pkg/eval/compile_effect_test.go#L194-L196

It doesn't even verify that the "file" object is no longer usable after the `fclose` is run.  Go ahead and copy that test into your test module and keep the other tests you wrote.

pkg/eval/builtin_fn_io.go
```diff
@@ -771,7 +771,7 @@ func toJSON(fm *Frame, inputs Inputs) error {
 // ```
 //
 // @cf fclose
-
+// This function had been deprecated, use file:open, instead
```
xiaq, line 774: Put this above the @cf line and in its own paragraph, and add a link:

```
//
// This function is deprecated; use [file:open](./file.html#open) instead.
//
// @cf fclose
```

pkg/eval/builtin_fn_io.go
```diff
@@ -786,7 +786,7 @@ func fopen(name string) (vals.File, error) {
 // Close a file opened with `fopen`.
 //
 // @cf fopen
-
+// This function had been deprecated, use file:close, instead
```
xiaq, line 789: Same

0.16.0-release-notes.md
```diff
@@ -31,3 +31,10 @@ New features in the interactive editor:
 
 -   The editor now uses a DSL for filtering items in completion, history
     listing, location and navigation modes.
+   
+-   A new `file:` module to work with files.
+
+# Deprecated features
+
+-   The `builtin:fopen` and `builtin:fclose` commands are deprecated. Use 
+    the equivalent commands in the new `file:` module.
```
xiaq, line 40: Please reformat this file with prettier, as documented in CONTRIBUTING.md. The CI error message also contains a diff, you can apply it manually if you find installing prettier too cumbersome.

pkg/eval/mods/file/file.go
```diff
@@ -0,0 +1,48 @@
...
+// File should be closed after opening using fclose
+// Example
+//  ```elvish-transcript
+// ~> cat a.txt
+// Some text in file
+// f = (fopen a.txt)
+// ~> cat < $f
+// Some text in file
+// ~> fclose $f
+// ```
```
xiaq, line 27: Elvdocs are markdown after `//`, so individual paragraphs should be separated by lines containing just `//`.

To build the reference doc for the new file module, add it to website/ref/index.toml and write a prelude Markdown file; see website/ref/math.toml for an example.

pkg/eval/mods/file/file_test.go
```diff
@@ -12,6 +12,12 @@ func TestFile(t *testing.T) {
 		ev.AddGlobal(eval.NsBuilder{}.AddNs("file", Ns).Ns())
 	}
 	evaltest.TestWithSetup(t, setup,
-		evaltest.That(`file:fopen`).Throws(evaltest.AnyError),
+		evaltest.That(`echo haha > out3`, `f = (file:open out3)`, `slurp <$f`, ` file:close $f`).Puts("haha\n"),
+		//No name is specified hence should give an error.
```
krader1961, line 16: Insert a space after the `//`.

pkg/eval/mods/file/file_test.go
```diff
@@ -12,6 +12,12 @@ func TestFile(t *testing.T) {
...
+		evaltest.That(`file:open`).Throws(evaltest.AnyError),
+		evaltest.That(`file:close`).Throws(evaltest.AnyError),
```
krader1961, line 18: That's good for your first attempt at writing an Elvish unit test but not ideal. It's better to verify the actual exception matches what we expect rather than matching any exception. In this case we want to match a `errs.ArityMismatch` error as shown in this example:

https://github.com/elves/elvish/blob/8a7073a20fb224f0a202f5c7aaf92bf7bc56d47d/pkg/eval/compile_effect_test.go#L146-L151

Although the "Actual" value should be zero for these two tests.

pkg/eval/mods/file/file_test.go
```diff
@@ -12,6 +12,12 @@ func TestFile(t *testing.T) {
...
+
+		//Should given an error when try to close an already closed file.
```
krader1961, line 20: Space after the `//`.

Also, please squash your commits into a single commit and do a force push. The four commits makes reviewing this much harder. See, for example, https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history but the gist is to execute `git rebase -i HEAD~5` and replace the word "pick" with "s" (or "squash") on every commit but the first one.

pkg/eval/mods/file/file.go
krader1961: Also, note that the text needs to be updated to reflect the new name of the command; i.e., `file:close` rather than `fopen`.

pkg/eval/mods/file/file_test.go
krader1961: FYI: Before I squash rebased your commits they added up to 197 insertions and 115 deletions (as reported by `git log --stat -4`). After squashing your commits there are 84 insertions and 2 deletions. The squashed series of commits is considerably more compact and easier to review.

pkg/eval/mods/file/file_test.go
supreetsingh10: Thank you sir. I am learning from you.

0.16.0-release-notes.md
supreetsingh10: Yes sir, I will take care of this.

pkg/eval/mods/file/file_test.go
supreetsingh10: This is the second time you are giving me the same feedback, I would like to apologize for my ignorance, but I didn't understand what did you mean by squash push at first. I did look up ```git force push``` but could not find learn and understand anything. Thank you for sending me the link. I will take care of it in future.

krader1961: @supreetsingh10: No need to apologize. I get the sense you are young and inexperienced. Much like the college interns I see at my $day-job. We want you to learn from your experience contributing to this project and be successful at getting changes you authored merged. You simply need to be willing to accept feedback. When you don't understand a suggestion like my earlier  suggestion to do a "squash-merge" you should ask for clarification. For example, "What do you mean by squash-merge?"

krader1961: @supreetsingh10: My recommendation is to squash your commits to a single commit using `git rebase -i HEAD~4` and changing the "pick" to "squash" for the last three commits. Then make the changes @xiaq and I recommended, merge those with `git commit --amend -a`, then force push to Github for review with `git push -f`.

supreetsingh10: Yes, sir, I am always looking forward to the feedback, and at the same time I would not like to burden you with my questions, so I research things myself and patch up the work that has been assigned to me. Nevertheless, since you have made this clear, I will not hesitate to ask questions in the future.

supreetsingh10:
Just did that squash push, for some reason new_realease.md got added, I wanted to format it for later. Excuse that for now, I will format it with prettier soon. I made all the changes, I hope my git skills don't make a fool of me today. Haha thanks

0.16.0-release-notes.md
supreetsingh10: Didn't mean to push this one yet, I will get back to it soon. Need to format it.

krader1961:
@supreetsingh10: You did not do a squash merge (aka squash rebase). I don't know what you did but it wasn't a squash merge since `git log` still shows four distinct commits in this pull-request. I suspect you squashed just a subset of your commits. When you run `git log` it should show a single commit by you.

supreetsingh10:
I have messed up big time. I am getting this thing, ```git @7b9061aa rebase-i 4/6 ~3```, in my elvish folder. Before that, I had made all the edits properly, read every single comment you posted here and checked if everything was up to the mark. Then I lost all of that due to some git mistake in ```git rebase -i```, I was getting some error, I tried resolving that with a lot of googling and researching. Then I did all those changes again, then I pushed this code. I had resolved that error, and it turns out it is not a squash push. I will understand what git is and what it does. I will then delete this repo of mine and fork elvish again and then push the changes. I have had no problem with anything, it is just this git thing that is causing all this trouble. I will study git today and figure out what I am doing wrong. Then push the changes.

xiaq:
Closing this since #1285 has been merged.

PR #1285, opened 2021-04-08
Links file module, marks fopen and fclose deprecation in compiler.go

supreetsingh10:
and builtin_fn_io.go

### DIFF
--- a/pkg/eval/mods/file/file.go
+++ b/pkg/eval/mods/file/file.go
@@ -1,0 +1,47 @@
+package file
+
+import (
+	"os"
+
+	"src.elv.sh/pkg/eval"
+	"src.elv.sh/pkg/eval/vals"
+)
+
+// elvdoc:fn open
+// ```elvish
+// file:open $filename
+// ```
+// Opens the file for reading.
+// File should be closed after opening using close
+// Example
+//  ```elvish-transcript
+// ~> cat a.txt
+// Some text in file
+// f = (file:open a.txt)
+// ~> cat < $f
+// Some text in file
+// ~> file:close $f
+// ```
+
+func open(name string) (vals.File, error) {
+	return os.Open(name)
+}
+
+// elvdoc: fn close
+// flcose is used to close an already opened file.
+// files are opened for reading using open
+// Will not accept string name of the file
+// ```elvish
+// ~> file:close $fileptr
+// ```
+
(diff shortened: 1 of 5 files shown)


pkg/eval/mods/file/file.go
```diff
@@ -0,0 +1,47 @@
...
+// File should be closed after opening using close
+// Example
+//  ```elvish-transcript
+// ~> cat a.txt
+// Some text in file
+// f = (file:open a.txt)
+// ~> cat < $f
+// Some text in file
+// ~> file:close $f
+// ```
```
xiaq, line 24: This doc (and the one for `close` below) need more work; you could also have just copied the existing doc from `fopen` and `fclose`. I'll follow up with a commit after merging this.

pkg/eval/mods/file/file.go
```diff
@@ -0,0 +1,47 @@
...
+func close(f vals.File) error {
+	return f.Close()
+}
+
+var Ns = eval.NsBuilder{}.AddGoFns("file:", fns).Ns()
+
+var fns = map[string]interface{}{
+	"close": close,
+	"open":  open,
+}
```
xiaq, line 47: Follow the convention in other module implementations, and put these at the beginning of the file.

pkg/eval/mods/file/file_test.go
```diff
@@ -0,0 +1,29 @@
+package file
+
+import (
+	"testing"
+
+	"src.elv.sh/pkg/eval"
+	"src.elv.sh/pkg/eval/errs"
+	"src.elv.sh/pkg/eval/evaltest"
```
xiaq, line 8: Follow the convention in other module implementations and use a dot import (see https://golang.org/ref/spec#Import_declarations) to import `evaltest`.

xiaq (approved):
Some minor feedback on the code style. I'll merge this PR and tweak the code a little bit in a followup commit.

pkg/eval/mods/file/file.go
supreetsingh10: I understand, thank you for suggesting this.

pkg/eval/mods/file/file.go
supreetsingh10: Alright, I used str.go as a reference, anyway I will take care of this in the future.

PR #1286, opened 2021-04-09
File module with pipe function. Pipe command marked deprecated in com…

supreetsingh10:
…piler.go and builtin_fn_io.go

### DIFF
--- a/pkg/eval/mods/file/file_test.go
+++ b/pkg/eval/mods/file/file_test.go
@@ -29,6 +29,25 @@ func TestFile(t *testing.T) {
 			errs.ArityMismatch{
 				What:     "arguments here",
 				ValidLow: 1, ValidHigh: 1, Actual: 0}),
+
+		That("file:prclose").Throws(
+			errs.ArityMismatch{
+				What:     "arguments here",
+				ValidLow: 1, ValidHigh: 1, Actual: 0}),
+
+		That("file:pwclose").Throws(
+			errs.ArityMismatch{
+				What:     "arguments here",
+				ValidLow: 1, ValidHigh: 1, Actual: 0}),
+
+		That(`p = (file:pipe)`, `echo haha > $p `, `pwclose $p`,
+			`slurp < $p`, `prclose $p`).Puts("haha\n"),
+
+		That(`p = (file:pipe)`, `echo Zeppelin > $p`, `file:pwclose $p`,
+			`echo Sabbath > $p`, `slurp < $p`, `file:prclose $p`).Puts("Zeppelin\n"),
+
+		That(`p = (file:pipe)`, `echo Legolas > $p`, `file:prclose $p`,
+			`slurp < $p`).Throws(AnyError),
 	)
 
 }
--- a/pkg/eval/builtin_fn_io.go
+++ b/pkg/eval/builtin_fn_io.go
@@ -827,6 +827,8 @@ func fclose(f vals.File) error {
 // # $p should be closed with prclose and pwclose afterwards
 // ```
 //
+// This function is deprecated; use [file:pipe](./file.html#pipe) instead.
+//
 // @cf prclose pwclose
 
 func pipe() (vals.Pipe, error) {
@@ -842,6 +844,8 @@ func pipe() (vals.Pipe, error) {
 //
 // Close the read end of a pipe.
 //
+// This function is deprecated; use [file:prclose](./file.html#prclose) instead.
+//
 // @cf pwclose pipe
 
 func prclose(p vals.Pipe) error {
@@ -855,6 +859,8 @@ func prclose(p vals.Pipe) error {
 // ```
 //
 // Close the write end of a pipe.
+//
+// This function is deprecated; use [file:pwclose](./file.html#pwclose) instead.
 //
 // @cf prclose pipe
 
--- a/pkg/eval/compiler.go
+++ b/pkg/eval/compiler.go
@@ -129,6 +129,12 @@ func (cp *compiler) checkDeprecatedBuiltin(name string, r diag.Ranger) {
 		msg = `the "fopen" command is deprecated; use "file:open" instead`
 	case "fclose~":
 		msg = `the "fclose" command is deprecated; use "file:close" instead`
+	case "pipe~":
+		msg = `the "pipe" command is deprecated; use "file:pipe" instead`
+	case "prclose~":
+		msg = `the "prclose" command is deprecated; use "file:prclose" instead`
+	case "pwclose":
+		msg = `the "pwclose" command is deprecated; use "file:pwclose" instead`
 	default:
 		return
 	}
--- a/pkg/eval/mods/file/file.go
+++ b/pkg/eval/mods/file/file.go
@@ -10,8 +10,11 @@ import (
 var Ns = eval.NsBuilder{}.AddGoFns("file:", fns).Ns()
 
 var fns = map[string]interface{}{
-	"close": close,
-	"open":  open,
+	"close":   close,
+	"open":    open,
+	"pipe":    pipe,
+	"prclose": prclose,
+	"pwclose": pwclose,
 }
 
 //elvdoc:fn open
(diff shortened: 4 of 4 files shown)


krader1961:
I didn't see any obvious problems with this change but I don't understand why `pipe` is being moved to `file:pipe` but the related `prclose` and `pwclose` commands are not included.

Also, the first line of a commit comment is generally expected to be a single sentence of at most 50 characters. You can include more detail about the change in subsequent paragraphs. This is not a hard and fast rule. It is just what people usually expect. Google "git commit comment standards" for some discussions of this topic. So your commit might begin with something like "Move builtins pipe, prclose, pwclose to pipe:.".

supreetsingh10:
Yes, sir, ```prclose``` and ```pwclose``` will not be included in this module. What other commands should I work on according to you? ```file:seek``` cannot be implemented just yet, because of some work remaining, this was told by Mr. Xiaq. Let me know what commands should I implement.

krader1961:
@supreetsingh10: I don't understand why you are moving the builtin `pipe` command but not `prclose` and `pwclose` to the recently introduced `file` module. The three functions are intimately related and the latter two have no use as a builtin other than with the output of the `file:pipe` command.

supreetsingh10:
Mr. @krader1961, Mr @xiaq  told me that he has other ideas for ```prclose``` and ```pwclose```, hence I did not move them here. I can work on other file module functions. I have made ```file:seek```` but I am not sure if we are "seeking" that function, haha.

krader1961:
@supreetsingh10: I am unaware of any ideas by @xiaq involving pipe objects, and the `prclose` and `pwclose` commands, that would have any impact on this P.R. The three commands are intimately related. It doesn't make any sense to move `builtin:pipe` to `file:pipe` without also moving `prclose` and `pwclose`.

Regarding a hypothetical `file:seek` command: see https://golang.org/pkg/os/#File.Seek. Implementing that command requires some discussion about the Elvish API and how it maps to the underlying Go method. That's why I suggested you not attempt to implement `file:seek` or `file:is-tty` until you had gained some experience making simpler changes to the code.

supreetsingh10:
I see the points you have made Mr. @krader1961, I will move the ```prclose``` and ```pwclose``` in the file module. As far as ```file:seek``` is concerned, I made a prototype of it just for myself. Since you had raised an issue on Github and I wanted to play around with it.

pkg/eval/mods/file/file_test.go
```diff
@@ -30,7 +30,21 @@ func TestFile(t *testing.T) {
...
+			errs.ArityMismatch{
+				What:     "arguments here",
+				ValidLow: 1, ValidHigh: 1, Actual: 0}),
+
+		That("file:pwclose").Throws(
+			errs.ArityMismatch{
+				What:     "arguments here",
+				ValidLow: 1, ValidHigh: 1, Actual: 0}),
+
+		That(`p = (file:pipe); echo haha > $p; pwclose $p
```
krader1961, line 43: Don't wrap strings by embedding a newline. You can also avoid the need for semicolons by splitting the command line into separate statements:

```
That(`p = (file:pipe)`, `echo haha > $p`, `pwclose $p`, `slurp < $p`, `prclose $p`).Puts("haha\n")
```

You can then split the long line at whichever `, ` boundaries make the most sense.

pkg/eval/mods/file/file_test.go
```diff
@@ -30,7 +30,21 @@ func TestFile(t *testing.T) {
...
+		; slurp < $p; prclose $p`).Puts("haha\n"),
+
+		That(`p = (file:pipe); echo haha > $p; file:prclose $p
+		; head -n1 < $p`).Throws(AnyError),
```
krader1961, line 47: This fails because the external `head` command on FreeBSD doesn't exit with a non-zero status when the pipe is closed. In general it is best to avoid relying on external commands in unit tests .

krader1961:
@supreetsingh10: Your change is failing on FreeBSD:

```
--- FAIL: TestFile (0.03s)
    --- FAIL: TestFile/p_=_(file:pipe);_echo_haha_>_$p;_file:prclose_$p___;_head_-n1_<_$p (0.01s)
        file_test.go:18: unexpected exception
        file_test.go:18: got: <nil>
        file_test.go:18: want: exception with reason any error
```

See my comment about the unit test that is failing.

pkg/eval/mods/file/file_test.go
supreetsingh10: So what should I do differently? If I do not wish on removing this test case, what should I do differently?

krader1961: Replace the external `head` command with an appropriate builtin that will succeed if the read end of the pipe is open and fail, by raising an error, if it is closed. In an interactive elvish shell try a few builtins that read from stdin to find one that would be useful for this test.

supreetsingh10: Thank you very much, will push the updated file today.

krader1961:
@supreetsingh10: The two most recent commits look okay. However, please squash-merge the four commits in this P.R. and do a force push (`git push -f`). You probably also need to rebase on the current `master` branch. I attempted to apply this P.R. to my local repo and it failed (something I've never seen before):

```
> hub am -3 https://github.com/elves/elvish/pull/1286
Applying: File module with pipe function. Pipe command marked deprecated in compiler.go and builtin_fn_io.go
Applying: Prclose and pwclose with Deprecation marked in specific file. Give feedback regarding the test cases.
Applying: Prclose and pwclose with Deprecation marked in specific file. Give feedback regarding the test cases.
error: sha1 information is lacking or useless (pkg/eval/mods/file/file.go).
error: could not build fake ancestor
Patch failed at 0003 Prclose and pwclose with Deprecation marked in specific file. Give feedback regarding the test cases.
hint: Use 'git am --show-current-patch=diff' to see the failed patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".
Exception: hub exited with 128
```

supreetsingh10:
Sorry for the late reply, had an exam today. Anyway, yeah I had quite a wrestle with ```git``` last time, I tried to squash push and tried to rewrite ```git history```, I will figure this out and update it.

supreetsingh10:
Mr. @krader1961 this should work. I checked git log, it should be a squash push, only 1 push is here. I think this worked. Let me know what you think of it.

krader1961:
LGTM

supreetsingh10:
> LGTM

Not gonna lie I had to look up what this meant. Thought this was some error code.

xiaq:
My plan for `prclose` and `pwclose` was to remove them, and make pipes support a `read` and a `write` index that can be used to get the read and write ends (so `prclose $p` would become `file:close $p[read]` and `pwclose $p` would become `file:close $p[write]`).

Personally, I don't insist very hard on PRs leaving the codebase in a perfectly consistent state. In particular, if leaving the codebase in a consistent state means introducing code that will be removed soon, I would merge code that is not logically 100% complete. OTOH if you do go to the trouble of doing that I don't object to it (which is the case here).

pkg/eval/mods/file/file_test.go
```diff
@@ -29,6 +29,25 @@ func TestFile(t *testing.T) {
...
+
+		That("file:prclose").Throws(
+			errs.ArityMismatch{
+				What:     "arguments here",
+				ValidLow: 1, ValidHigh: 1, Actual: 0}),
+
+		That("file:pwclose").Throws(
+			errs.ArityMismatch{
+				What:     "arguments here",
+				ValidLow: 1, ValidHigh: 1, Actual: 0}),
```
xiaq, line 41: Looking at these tests and the two above, I don't find them necessary.

The code for matching parameters and arguments is implemented generically for every function in https://github.com/elves/elvish/blob/master/pkg/eval/go_fn.go, there's no need to test it for every builtin function. I'll remove these test cases after merging.

PR #1290, opened 2021-04-11
Correct grammar of wrong type of argument error

pdietl:


pkg/eval/go_fn.go
```diff
@@ -208,7 +220,7 @@ func (b *goFn) Call(f *Frame, args []interface{}, opts map[string]interface{}) e
 		ptr := reflect.New(typ)
 		err := vals.ScanToGo(arg, ptr.Interface())
 		if err != nil {
-			return fmt.Errorf("wrong type of %d'th argument: %v", i+1, err)
```
krader1961, line 211: Sorry, @pdietl, but this change introduces more problems than it solves, IMO. The "rules", such as they are, for writing ordinal numbers using a short-hand notation are not obvious or consistent. And I say that as a 60+ year old native English speaker. See, for example, https://www.thoughtco.com/what-is-an-ordinal-number-1691459#:~:text=All%20ordinal%20numbers%20carry%20a,abbreviations%20(2nd%2C%203rd).

My preference is to modify the message to use a less ambiguous reference to the value such as `"arg number %d has wrong type: %v"`.


Starting with your replacement of `'th` with `th`; e.g., '4'th` to `4th`. Yes, `2nd` is generally recognized as an abbreviation for `second`. But `1004th` is

pdietl: Yeah makes sense. I was just thinking in terms of how the wording was previously chosen.

But even with your new suggestion, there remains a critical decision,
Is the first argument “argument 0” or “argument 1?” :p

krader1961: Elvish is explicitly zero based for indexing. See https://elv.sh/ref/language.html#indexing. Which means that in addition to my proposed change to the message the corresponding argument should be changed from `i+1` to `i`. This is, however, highly subjective. Which means that the project owner, @xiaq, might have a different opinion as to the optimal way to write such error messages.

pdietl: Indeed. One could always take the middle ground and start at index 0.5

xiaq: I endorse @krader1961's suggestion, "argument number %d" is both simpler and gramatically correct.

As for the base, I'd say stick to base 0 (so the first argument is argument 0). To reduce the possibility for confusion, maybe the name of the argument can be added too.

krader1961:
@pdietl: I've been annoyed, in the recent past, by the error message your change improves. For example, while working on a different change I saw this error output today:

```
Exception: wrong type of 1'th argument: cannot parse as number: WTF
[RC file], line 6:              echo WTF >> /tmp/x
```

Like you, I find the `1'th argument` portion of the message annoying. Please consider updating your change with the wording I suggested, or something like it.

xiaq:
This is now superseded by 2ac9d891cadb8b64aa840301e0ca068814d965f7. Thanks for the contribution, but I'm closing the PR.

PR #1293, opened 2021-04-19
Add a `-i` CLI option

### LINKED ISSUE #1292: Add support for the legacy `-i` shell CLI option
The `script` command, when invoked with no arguments, does the equivalent of `$SHELL -i`. This fails when `$SHELL` is the path to elvish because it doesn't support that POSIX legacy shell option. Supporting `elvish -i` is effectively zero cost (just a few extra bytes in the binary image and one cheap function call). Too, if elvish ever wanted to recognize a "force interactive behavior" option it should use `-i` for that purpose. So there doesn't seem to be any downside to treating `-i` as a no-op today and the upside is that commands like `script` won't unexpectedly fail.

krader1961:
POSIX shells have a `-i` short option to force interactive mode. Some
programs, such as the `script` command, assume that option is recognized
by the shell it spawns. There isn't any reason elvish shouldn't silently
ignore that option.

Resolves #1292

### DIFF
--- a/pkg/prog/prog.go
+++ b/pkg/prog/prog.go
@@ -63,6 +63,9 @@ func newFlagSet(stderr io.Writer, f *Flags) *flag.FlagSet {
 	fs.BoolVar(&f.BuildInfo, "buildinfo", false, "show build info and quit")
 	fs.BoolVar(&f.JSON, "json", false, "show output in JSON. Useful with -buildinfo.")
 
+	// The `-i` option is for compatibility with POSIX shells so that programs, such as the `script`
+	// command, will work when asked to launch an interactive Elvish shell.
+	fs.Bool("i", false, "force interactive mode; currently ignored")
 	fs.BoolVar(&f.CodeInArg, "c", false, "take first argument as code to execute")
 	fs.BoolVar(&f.CompileOnly, "compileonly", false, "Parse/Compile but do not execute")
 	fs.BoolVar(&f.NoRc, "norc", false, "run elvish without invoking rc.elv")


pkg/prog/prog.go
```diff
@@ -63,6 +63,9 @@ func newFlagSet(stderr io.Writer, f *Flags) *flag.FlagSet {
 	fs.BoolVar(&f.BuildInfo, "buildinfo", false, "show build info and quit")
 	fs.BoolVar(&f.JSON, "json", false, "show output in JSON. Useful with -buildinfo.")
 
+	// The `-i` option is for compatibility with POSIX shells so that programs, such as the `script`
+	// command, will work when asked to launch an interactive Elvish shell.
+	fs.Bool("i", false, "force interactive mode; currently ignored")
```
supreetsingh10, line 68: Can you please explain what are we trying to implement here? What interactive shell?

xiaq: This is a distinction POSIX shells make. https://tldp.org/LDP/abs/html/intandnonint.html is a good introduction to the issue.

krader1961: @supreetsingh10: See issue #1292 for more detail about "what are we trying to implement here?". The phrase "interactive Elvish shell" is currently underspecified but basically means an Elvish shell started with its stdin and stdout attached to a "terminal" (aka "tty").

supreetsingh10: Thank you Mr. @xiaq and Mr. @krader1961 this cleared out a few things.

PR #1298, opened 2021-05-06
File:truncate fixed with requested changes

supreetsingh10:
All the changes have been made, the test cases and documentation will be added once this gets approved.

### DIFF
--- a/pkg/eval/mods/file/file.go
+++ b/pkg/eval/mods/file/file.go
@@ -1,20 +1,23 @@
 package file
 
 import (
+	"math/big"
 	"os"
 
 	"src.elv.sh/pkg/eval"
+	"src.elv.sh/pkg/eval/errs"
 	"src.elv.sh/pkg/eval/vals"
 )
 
 var Ns = eval.NsBuilder{}.AddGoFns("file:", fns).Ns()
 
 var fns = map[string]interface{}{
-	"close":   close,
-	"open":    open,
-	"pipe":    pipe,
-	"prclose": prclose,
-	"pwclose": pwclose,
+	"close":    close,
+	"open":     open,
+	"pipe":     pipe,
+	"prclose":  prclose,
+	"pwclose":  pwclose,
+	"truncate": truncate,
 }
 
 //elvdoc:fn open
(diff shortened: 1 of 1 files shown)


pkg/eval/mods/file/file.go
```diff
@@ -124,3 +127,43 @@ func prclose(p vals.Pipe) error {
...
+				ValidHigh: "8 bytes",
+				Actual:    "More than 8 bytes",
+			}
+		}
+
+	default:
+		return errs.BadValue{
+			What:   "size argument for os.Truncate",
+			Valid:  "int or *big.Int",
+			Actual: "float64",
```
krader1961, line 166: Don't hardcode the `"float64"` type string since this could be some other type.

@xiaq: It seems to me this function (and some of the code in the `math` module) should be using `ScanToGo()` rather than an explicit `switch size.(type)`. However, at the present time `ScanToGo` is not sufficiently general for and doesn't support commonly used Go types like `int64`.

pkg/eval/mods/file/file.go
```diff
@@ -124,3 +127,43 @@ func prclose(p vals.Pipe) error {
...
+		if err != nil {
+			return err
+		} else {
+			return nil
+		}
+
+	case *big.Int:
+		if (size.(*big.Int)).IsInt64() {
+			s := (size.(*big.Int)).Int64()
+			err := os.Truncate(name, s)
```
krader1961, line 147: Don't duplicate this code. It's also far more verbose than necessary. This is a situation where a different pattern is clearer and less likely to contain errors:

```
var offset int64
switch size.(type) {
case ... // return early if an invalid value is detected otherwise assign the value to `offset`.
}

return os.Truncate(name, offset)
```

pkg/eval/mods/file/file.go
```diff
@@ -124,3 +127,43 @@ func prclose(p vals.Pipe) error {
...
+			if err != nil {
+				return err
+			} else {
+				return nil
+			}
+		} else {
+			return errs.OutOfRange{
+				What:      "size argument for os.Truncate",
+				ValidLow:  "0",
+				ValidHigh: "8 bytes",
```
krader1961, line 157: That string doesn't make any sense. It should be derived from a constant like `MaxInt64` (see https://golang.org/pkg/math/#pkg-constants). However, the real question is whether this case should even be using `errs.OutOfRange` since it does not explicitly verify that the value is greater than or equal to zero. This is another example why the code should be structured per my previous comment so that we only execute this if the value is not a `int64` or is not negative.

pkg/eval/mods/file/file.go
supreetsingh10: Noted, will fix this.

pkg/eval/mods/file/file.go
supreetsingh10: Thank you, this is a great piece of advice, I will implement this here as well as in future commits. Thank you, Mr. @krader1961

supreetsingh10:
https://gist.github.com/supreetsingh10/1a95da250a650c3b9ab7fcc47f9cc18c

Mr. @krader1961 and Mr. @xiaq please check this out, I have observed something, either this is a bug or I am extremely stupid. Haha, anyway apologies if I am not able to articulate the problem well, if there are more questions, I will answer them here.

PR #1301, opened 2021-05-11
file:truncate added to file module, with requested changes

supreetsingh10:
Made the requested changes, also added documentation, all feedback is welcome, thank you once again.

### DIFF
--- a/pkg/eval/mods/file/file.go
+++ b/pkg/eval/mods/file/file.go
@@ -1,20 +1,23 @@
 package file
 
 import (
+	"math/big"
 	"os"
 
 	"src.elv.sh/pkg/eval"
+	"src.elv.sh/pkg/eval/errs"
 	"src.elv.sh/pkg/eval/vals"
 )
 
 var Ns = eval.NsBuilder{}.AddGoFns("file:", fns).Ns()
 
 var fns = map[string]interface{}{
-	"close":   close,
-	"open":    open,
-	"pipe":    pipe,
-	"prclose": prclose,
-	"pwclose": pwclose,
+	"close":    close,
+	"open":     open,
+	"pipe":     pipe,
+	"prclose":  prclose,
+	"pwclose":  pwclose,
+	"truncate": truncate,
 }
 
 //elvdoc:fn open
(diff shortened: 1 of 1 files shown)


supreetsingh10:
Mr @xiaq why is this one test case failing?

pkg/eval/mods/file/file.go
```diff
@@ -124,3 +127,55 @@ func prclose(p vals.Pipe) error {
...
+// ```elvish
+// file:truncate $filename $size
+// ```
+// Truncate changes the size of
+// the named file. If the file is
+// a symbolic link, it changes the
+// size of the link's target.
+//
+// If there is an error it will
+// be of type *PathError.
```
xiaq, line 142: This seems to be copied verbatim from https://pkg.go.dev/os#Truncate; Go docs have different conventions to elvdocs and I'll amend this.

pkg/eval/mods/file/file.go
```diff
@@ -124,3 +127,55 @@ func prclose(p vals.Pipe) error {
...
+	var offset int64
+	switch size.(type) {
+	case int:
+		offset = int64(size.(int))
+		if offset < 0 {
+			return errs.BadValue{
+				What:   "Numeric input for the function is wrong",
+				Valid:  "positive 64 bit integer",
+				Actual: "Number less than 0",
+			}
```
xiaq, line 154: This should use errs.OutOfRange. The multiple checks against negative values should also be merged together.

pkg/eval/mods/file/file.go
```diff
@@ -124,3 +127,55 @@ func prclose(p vals.Pipe) error {
...
+// the named file. If the file is
+// a symbolic link, it changes the
+// size of the link's target.
+//
+// If there is an error it will
+// be of type *PathError.
+
+func truncate(name string, size vals.Num) error {
+	var offset int64
+	switch size.(type) {
```
xiaq, line 146: Using an assigned type switch helps avoid type assertions below:

```go
switch size := size.(type)
```

xiaq (approved):
The CI is failing because the overall test coverage will decrease - there is no test for the newly introduced function.

But I'll merge this anyway and do a fixup commit.

PR #1304, opened 2021-05-13
Golint cleanups

krader1961:
This cleanup is in one sense slightly pointless. But I've found that
addressing such lint tends to be beneficial as it makes it easier to notice
when real problems are introduced.

pkg/env/env.go:10:2: don't use ALL_CAPS in Go names; use CamelCase
pkg/env/env.go:12:2: don't use ALL_CAPS in Go names; use CamelCase
pkg/env/env.go:18:2: don't use ALL_CAPS in Go names; use CamelCase
pkg/eval/vals/num.go:159:10: if block ends with a return statement, so drop this else and outdent its block
pkg/persistent/hash/hash.go:32:9: if block ends with a return statement, so drop this else and outdent its block
pkg/persistent/hash/hash.go:41:9: if block ends with a return statement, so drop this else and outdent its block

### DIFF
--- a/pkg/cli/lscolors/lscolors.go
+++ b/pkg/cli/lscolors/lscolors.go
@@ -51,7 +51,7 @@ func GetColorist() Colorist {
 }
 
 func getLsColors() string {
-	lsColorString := os.Getenv(env.LS_COLORS)
+	lsColorString := os.Getenv(env.LsColors)
 	if len(lsColorString) == 0 {
 		return defaultLsColorString
 	}
--- a/pkg/edit/highlight_test.go
+++ b/pkg/edit/highlight_test.go
@@ -81,16 +81,16 @@ func TestMakeHasCommand(t *testing.T) {
 	// Set up environment.
 	testDir, cleanup := testutil.InTestDir()
 	defer cleanup()
-	oldPath := os.Getenv(env.PATH)
-	defer os.Setenv(env.PATH, oldPath)
+	oldPath := os.Getenv(env.Path)
+	defer os.Setenv(env.Path, oldPath)
 	if runtime.GOOS == "windows" {
-		oldPathExt := os.Getenv(env.PATHEXT)
-		defer os.Setenv(env.PATHEXT, oldPathExt)
-		os.Unsetenv(env.PATHEXT) // force default value
+		oldPathExt := os.Getenv(env.Pathext)
+		defer os.Setenv(env.Pathext, oldPathExt)
+		os.Unsetenv(env.Pathext) // force default value
 	}
 
 	// Set up a directory in PATH.
-	os.Setenv(env.PATH, filepath.Join(testDir, "bin"))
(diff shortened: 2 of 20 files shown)


krader1961:
With this lint cleanup just three warnings remain if you ignore the "exported... should have comment or be unexported" warnings:

```
pkg/eval/eval.go:380:6: type name will be used as eval.EvalCfg by other packages, and that stutters; consider calling this Cfg
pkg/parse/parse.go:45:6: func name will be used as parse.ParseAs by other packages, and that stutters; consider calling this As
pkg/rpc/server.go:5:1: package comment should not have leading space
```

xiaq:
Golint is deprecated: https://github.com/golang/lint, and their official stance is:

> The suggestions made by golint are exactly that: suggestions. Golint is not perfect, and has both false positives and false negatives. Do not treat its output as a gold standard. We will not be adding pragmas or other knobs to suppress specific warnings, so do not expect or require code to be completely "lint-free". In short, this tool is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process. Golint makes suggestions for many of the mechanically checkable items listed in Effective Go and the CodeReviewComments wiki page.

I'll be open to make Elvish lint-free with [staticcheck](https://staticcheck.io) instead, which also supports using `//lint:ignore` to suppress some warnings.

krader1961:
> Golint is deprecated:

Yes, but that doesn't change the fact that the cleanups in this PR seem quite reasonable. You've pointed out to me, multiple times, where a change I've proposed introduced a var that did not adhere to the Go CamelCase naming convention. Similarly, there's no good reason to end a function with a `else { return }` block. And in the case of the two functions I touched it is better to not assume the only valid pointer sizes will always be four and eight bytes.

xiaq:
> You've pointed out to me, multiple times, where a change I've proposed introduced a var that did not adhere to the Go CamelCase naming convention.

For the case of environment variable names I feel an exception is warranted, since the environment variables themselves use `ALL_CAPS` casing instead of `CamelCase`.

> Similarly, there's no good reason to end a function with a `else { return }` block. And in the case of the two functions I touched it is better to not assume the only valid pointer sizes will always be four and eight bytes.

That's a reasonable change, I overlooked it - feel free to open another PR for that.

xiaq:
Even the standard library breaks the rule when there is a strong reason to prefer the `ALL_CAPS` spelling; for example, the [os package](https://pkg.go.dev/os) has constants spelled like `O_APPEND` because their well-known C counterparts were spelled that way.

krader1961:
> For the case of environment variable names I feel an exception is warranted, since the environment variables themselves use ALL_CAPS casing instead of CamelCase.

I am not convinced that env vars deserve the same exception as OS specific symbols like `O_APPEND`. In no small part because a) env var names can contain lowercase characters and other chars that are not valid in a Go symbol name, and b) the point of the `env` package is to make it harder to mistake a literal string with a reference to an environment var name.

Issue #1053, opened 2020-06-21
Make edit:command-history easier, and faster, to use with tools like fzf


krader1961:
See #1051 for what prompted me to open this issue.

At least one common use-case of the `edit:command-history` command would be easier and faster (i.e., more efficient) with some enhancements to that command:

1) An option to output the commands in newest-to-oldest order (keeping the default oldest-to-newest).

1) An option to eliminate duplicate commands regardless of the output order. "Duplicate" here means comparing just the "cmd"  member of the default map output.

1) An option to request just the "cmd" string be emitted rather than a map of all command attributes.

1) Maybe: An option to limit the output to the _N_ newest commands; without regard to the order of the output. Note that the `take` command only works as a substitute for this option if the option to output the commands in newest-to-oldest order is used. On the other hand this option may not make much sense when coupled with oldest-to-newest order (the default). Regardless of whether or not it is treated as the first or last _N_ commands. Thus the "maybe" prefix since it may be better to omit this option.

Using the first three options makes it easy to do command selection using the popular `fzf` utility. It should also improve the time to produce a `fzf` menu by at least an order of magnitude.

The duplicate command elimination option might warrant more discussion. Specifically, whether elvish should always do duplicate elimination when updating the command data store. In all my years using ksh and zsh as my day-to-day interactive shells I can't ever recall a situation where retaining each individual use of a command in the history was useful. As opposed to being an annoyance when I listed the _N_ most recent commands I ran or was stepping backward through my command history; e.g., via [up-arrow]. Note that fish always does duplicate elimination. I can't recall ever hearing anyone complain about the deduplication.

krader1961:
See also #568 which asked for a way to access a deduplicated command history more than two years ago.

krader1961:
A recent question on IRC reminded me that I, like @zzamboni, had added some functions to my *rc.elv* several months ago to use the `fzf` program rather than the builtin histlist mode but hadn't bound those functions to Ctrl-R. So I did so and was happy with fzf but not how long it took before the fzf query prompt was usable.

On my primary MacPro server `edit:command-history >/dev/null` completes in 55 ms. The following function takes 760 ms (~13.8 times longer):

```
edit:command-history | each [hist-entry]{
    print $hist-entry[cmd]"\000"
} | perl -n0e 'print unless $h{$_}++' >/dev/null
```

A pure Elvish solution takes 2849 ms (~3.7 and ~51.8 times longer than the perl and unmodified history solutions respectively):

```
[...]

PR #1307, opened 2021-05-15
Implement edit:command-history &dedup

### LINKED ISSUE #568: API for accessing deduplicated command history
Please add a `&dedup` option to `edit:command-history`

krader1961:
This implements `&dedup` and `&newest-first` options for
`edit:command-history`. This makes it noticably cheaper to feed unique
command history into external commands like `fzf`.

Related #1053
Fixes #568

### DIFF
--- a/pkg/edit/store_api.go
+++ b/pkg/edit/store_api.go
@@ -10,24 +10,36 @@ import (
 	"src.elv.sh/pkg/eval"
 	"src.elv.sh/pkg/eval/vals"
 	"src.elv.sh/pkg/parse/parseutil"
+	"src.elv.sh/pkg/store"
 )
 
 var errStoreOffline = errors.New("store offline")
 
 //elvdoc:fn command-history
 //
-// Outputs the entire command history as a stream of maps. Each map has a `id`
-// key that identifies the sequence number of the entry, and a `cmd` key that
-// identifies the content.
+// ```elvish
+// edit:command-history &dedup &newest-first
+// ```
 //
-// Use indexing to extract individual entries. For example, to extract the
-// content of the last command, do this:
+// Outputs the command history as a stream of maps. Each map has a `id` key for the sequence number
+// of the command, and a `cmd` key for the text of the command. If `&dedup` is true only the most
+// recent instance of each command (when comparing just the `cmd` key) is output. If `&newest-first`
+// is true the output is in newest to oldest order (default is oldest to newest).
+//
+// Use indexing to extract individual entries. For example, to extract the text of the most recent
+// command do this:
 //
 // ```elvish
 // edit:command-history | put [(all)][-1][cmd]
 // ```
+//
+// @cf builtin:dir-history
 
-func commandHistory(fuser histutil.Store, ch chan<- interface{}) error {
+type cmdhistOpt struct{ Dedup, NewestFirst bool }
+
+func (o *cmdhistOpt) SetDefaultOptions() {}
+
+func commandHistory(opts cmdhistOpt, fuser histutil.Store, ch chan<- interface{}) error {
 	if fuser == nil {
 		return errStoreOffline
 	}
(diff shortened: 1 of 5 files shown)


krader1961:
> This makes it noticably cheaper to feed unique command history into external commands like fzf.

I can't stress this point enough. This change means that a Ctrl-R binding that feeds the output of `edit:command-history` to `fzf` has gone from being mildly annoying (due to the visual artifacts and the time before I can interact with `fzf`) to being almost indistinguishable with the builtin Ctrl-R implementation.

pkg/edit/store_api.go
```diff
@@ -10,37 +10,88 @@ import (
...
+		}
+	} else {
+		for _, cmd := range cmds {
+			ch <- vals.MakeMap("id", strconv.Itoa(cmd.Seq), "cmd", cmd.Text)
+		}
 	}
 	return nil
 }
 
+func uniqCmds(allCmds []store.Cmd, newestFirst bool) []store.Cmd {
```
xiaq, line 70: 1. Unnecessary for `uniqCmds` to take `newestFirst` as an argument; this merely duplicates the reverse in the previous function.

2. Since the option is called "dedup", better to be consistent and call this function `dedupCmds` too.

pkg/edit/store_api.go
```diff
@@ -10,37 +10,88 @@ import (
...
+	// The division by four is based on the empirical observation that my personal command history
```
xiaq, line 71: > my personal command history

Although code history is traceable via `git blame`, better to avoid requiring the reader to guess who "I" is. Either point out who "I" is, or just say vaguely that this is *some* sort of empirical observation.

pkg/edit/store_api_test.go
```diff
@@ -12,16 +12,65 @@ func TestCommandHistory(t *testing.T) {
...
 	testGlobal(t, f.Evaler,
 		"cmds",
 		vals.MakeList(
 			vals.MakeMap("id", "1", "cmd", "echo 1"),
-			vals.MakeMap("id", "2", "cmd", "echo 2")))
+			vals.MakeMap("id", "2", "cmd", "echo 2"),
+			vals.MakeMap("id", "3", "cmd", "echo 2"),
+			vals.MakeMap("id", "4", "cmd", "echo 1"),
+			vals.MakeMap("id", "5", "cmd", "echo 3"),
+			vals.MakeMap("id", "6", "cmd", "echo 1"),
```
xiaq, line 33: A little helper for building this map is warranted here:

```go
func cmdMap(id int, cmd string) vals.Map
```

pkg/edit/store_api.go
```diff
@@ -10,37 +10,88 @@ import (
...
+	} else if opts.NewestFirst {
+		reverse(cmds)
+	}
+	if opts.CmdOnly {
+		for _, cmd := range cmds {
+			ch <- cmd.Text
+		}
+	} else {
+		for _, cmd := range cmds {
+			ch <- vals.MakeMap("id", strconv.Itoa(cmd.Seq), "cmd", cmd.Text)
```
xiaq, line 64: Now that Elvish supports actual numbers, the seq field can be exposed as-is without converting to string.

pkg/edit/store_api.go
```diff
@@ -10,37 +10,88 @@ import (
...
 
 var errStoreOffline = errors.New("store offline")
 
 //elvdoc:fn command-history
 //
-// Outputs the entire command history as a stream of maps. Each map has a `id`
-// key that identifies the sequence number of the entry, and a `cmd` key that
-// identifies the content.
+// ```elvish
+// edit:command-history &dedup &newest-first
```
xiaq, line 21: Specify the default values for the options:

```
&dedup=$false &newest-first=$false &cmd-only=$false
```

pkg/edit/store_api.go
```diff
@@ -10,37 +10,88 @@ import (
...
+// ```
 //
-// Use indexing to extract individual entries. For example, to extract the
-// content of the last command, do this:
+// Outputs the command history as a stream of maps unless `&cmd-only` is true. Each map has a `id`
```
xiaq, line 24: > ... unless `&cmd-only` is true

In general, avoid "unless", unless the condition after "unless" is truly an edge case.

pkg/edit/store_api.go
```diff
@@ -10,37 +10,88 @@ import (
...
+// key for the sequence number of the command, and a `cmd` key for the text of the command. If
+// `&dedup` is true only the most recent instance of each command (when comparing just the `cmd`
+// key) is output. If `&newest-first` is true the output is in newest to oldest order (default is
+// oldest to newest). If `&cmd-only` is true then only the command text is output (default is a map
+// as described previously).
+//
+// Use indexing to extract individual entries. For example, to extract the text of the most recent
```
xiaq, line 31: This is a bit of an out-of-place comment here since it applies to all commands that output a stream of values. It's sufficient to just say that the following are useful examples.

xiaq (approved):

pkg/edit/store_api.go
xiaq: Hmm, I realized that the reason you made this function take `newestFirst` as an argument is that it naturally reverses the commands, so I'll keep that.

PR #1314, opened 2021-05-20
Fix isDrive

krader1961:
I was looking at "TODO" comments that could be eliminated and noticed the
one preceding the sole use of `isDrive()`. Which caused me to notice its
implementation is broken. The `s[1] < 'z'` shnould be `s[0] < 'z'`

Rather than make the mechanical fix simplify the implementation by using
`unicode.IsLetter`. That will probably be marginally slower but this
function should never appear in a hot execution path.

### DIFF
--- a/pkg/glob/glob.go
+++ b/pkg/glob/glob.go
@@ -8,7 +8,9 @@ import (
 	"unicode/utf8"
 )
 
-// TODO: Use native path separators instead of always using /.
+// TODO: Use native path separators (i.e., os.PathSeparator) instead of always using `/` when
+// building a path. When parsing a path note that Windows supports both `/` and `\` so Elvish should
+// as well. TBD is whether os.IsPathSeparator handles Windows correctly.
 
 // PathInfo keeps a path resulting from glob expansion and its FileInfo. The
 // FileInfo is useful for efficiently determining if a given pathname satisfies
@@ -40,7 +42,7 @@ func (p Pattern) Glob(cb func(PathInfo) bool) bool {
 		segs = segs[1:]
 		dir += "/"
 	} else if runtime.GOOS == "windows" && len(segs) > 1 && IsLiteral(segs[0]) && IsSlash(segs[1]) {
-		// TODO: Handle UNC.
+		// TODO: Handle Windows UNC paths.
 		elem := segs[0].(Literal).Data
 		if isDrive(elem) {
 			segs = segs[2:]
@@ -51,9 +53,14 @@ func (p Pattern) Glob(cb func(PathInfo) bool) bool {
 	return glob(segs, dir, cb)
 }
 
+// isLetter returns true if the byte is an ASCII letter.
+func isLetter(chr byte) bool {
+	return ('a' <= chr && chr <= 'z') || ('A' <= chr && chr <= 'Z')
+}
+
+// isDrive returns true if the string looks like a Windows drive letter path prefix.
 func isDrive(s string) bool {
-	return len(s) == 2 && s[1] == ':' &&
-		(('a' <= s[0] && s[1] <= 'z') || ('A' <= s[0] && s[0] <= 'Z'))
+	return len(s) == 2 && s[1] == ':' && isLetter(s[0])
 }
 
 // glob finds all filenames matching the given Segments in the given dir, and


krader1961:
FYI: I considered adding a dependency on https://github.com/scott-ainsworth/go-ascii/ to minimize the run-time overhead. I decided that, for Elvish, the cost of that dependency was greater than the slower builtin unicode package.

pkg/glob/glob.go
```diff
@@ -40,7 +41,7 @@ func (p Pattern) Glob(cb func(PathInfo) bool) bool {
 		segs = segs[1:]
 		dir += "/"
 	} else if runtime.GOOS == "windows" && len(segs) > 1 && IsLiteral(segs[0]) && IsSlash(segs[1]) {
-		// TODO: Handle UNC.
+		// TODO: Handle Windows UNC paths. Google "windows unc path".
```
xiaq, line 44: IMO clarifying UNC to the more Google-able "Windows UNC paths" is sufficient; if the reader is not familiar with the concept they should know how to search for that themselves. Otherwise we'd litter the codebase with comments that say "Google $concept_x", which is unnecessary and a bit patronizing TBH.

pkg/glob/glob.go
```diff
@@ -52,8 +53,7 @@ func (p Pattern) Glob(cb func(PathInfo) bool) bool {
 }
 
 func isDrive(s string) bool {
-	return len(s) == 2 && s[1] == ':' &&
-		(('a' <= s[0] && s[1] <= 'z') || ('A' <= s[0] && s[0] <= 'Z'))
+	return len(s) == 2 && s[1] == ':' && unicode.IsLetter(rune(s[0]))
```
xiaq, line 56: `unicode.IsLetter` does much more than checking the rune is between A and Z. The behaviour *happens* to be correct since s[0] is a single byte, but I find this unnecessarily clever. A plain function to check ASCII letters is better:

```go
func isASCIILetter(b byte) bool
```

pkg/glob/glob.go
krader1961: Agreed. I added that because I was surprised **I* had to google it.

pkg/glob/glob.go
krader1961: I have no idea why you think this is clever since the unicode functions produce the expected results for ASCII bytes. If we're going to implement ASCII specific analogues it is probably better to bite the bullet and add a dependency on https://github.com/scott-ainsworth/go-ascii/. Yay or nay?

krader1961: I ran `grep "'0'" **.go` and found fewer instances (five?) of checking if a byte is an ASCII digit than I expected. Nonetheless, it illustrates that we should either depend on the `go-ascii` package for such tests or implement our own, equivalent, set of functions.

krader1961: One of our ASCII char tests is for an octal digit (see *pkg/parse/parse.go*). Which isn't supported explicitly by the `scott-aisnsworth/go-ascii` package. So that particular test would have to remain in its current form (`if r < '0' || r > '7' {`) unless abstracted into an elvish specific function. Argh! At this point I'm inclined to introduce package `ascii` (*pkg/ascii/ascii.go*) that provides the ASCII classifications required by Elvish. Then consistently use those functions rather than hand-coded equality tests like `r < '0' || r > '7'`.

xiaq: > I have no idea why you think this is clever since the unicode functions produce the expected results for ASCII bytes.

Clever code is when the writer tries to save a few keystrokes and incurs cognitive burden on the reader.

To understand that the version using unicode.IsLetter is working correctly, the reader needs to:

* Know that the Unicode definition of letters happen to include A-Z and a-z, and nothing more, in the ASCII range

* Read the code carefully and realize that it is converting a byte to a rune, so the rune can only be in the ASCII range

> If we're going to implement ASCII specific analogues it is probably better to bite the bullet and add a dependency on https://github.com/scott-ainsworth/go-ascii/. Yay or nay?

No. A little copying is better than a little dependency.

krader1961: > Know that the Unicode definition of letters happen to include A-Z and a-z, and nothing more, in the ASCII range

I would expect anyone minimally competent with respect to the Go language and Unicode to know that fact. Even programmers looking at this code who don't satisfy that assumption will probably draw the obvious, correct, conclusion.  It's also no more clever than the existing logic. Consider, for example, that the existing code is incorrect if the encoding is EBCDIC. Yes, I know that isn't possible in practice but that isn't the point. There is nothing clever about using `unicode.IsLetter` other than the explicit type coercion from byte to rune. That, I will grant you, is clever in the sense you are using the term and it would be preferable to avoid it. So I'll introduce an `ascii` package to make such tests easier to read and less likely to be misunderstood by a human or the Go tool chain.

P.S., The unit tests which exercise `isDrive()` only pass uppercase `C:` or strings of length one. I learned this by using my un-merged trace package.

xiaq: > I would expect anyone minimally competent with respect to the Go language and Unicode to know that fact.

The [godoc for unicode.IsLetter](https://pkg.go.dev/unicode#IsLetter) reads:

> IsLetter reports whether the rune is a letter (category L).

If your definition of minimal competency with Unicode requires immediately knowing that the intersection between Unicode category L and ASCII is [a-zA-Z] without looking up how Unicode defines category L, I'd happily admit that I'm not minimally competent with Unicode.

My preferred definition of minimal competency with Unicode would be knowing enough Unicode to have a healthy fear of any Unicode concept - including seemingly banal concepts as "letter" in the Unicode context, and not using them in contexts that shouldn't require an understanding of Unicode.

pkg/glob/glob.go
```diff
@@ -8,7 +8,9 @@ import (
 	"unicode/utf8"
 )
 
-// TODO: Use native path separators instead of always using /.
+// TODO: Use native path separators (i.e., os.PathSeparator) instead of always using `/` when
+// building a path. When parsing a path note that Windows supports both `/` and `\` so Elvish should
+// as well. TBD is whether os.IsPathSeparator handles Windows correctly.
```
xiaq, line 13: It does - https://cs.opensource.google/go/go/+/refs/tags/go1.16.4:src/os/path_windows.go;l=12-16

xiaq (approved):

PR #1319, opened 2021-05-22
Fix bug identified by staticcheck

krader1961:
This resolves this staticcheck warning:

    pkg/cli/term/file_reader_unix.go:75:2: empty critical section (SA2001)

### DIFF
--- a/pkg/cli/term/file_reader_unix.go
+++ b/pkg/cli/term/file_reader_unix.go
@@ -70,9 +70,9 @@ func (r *bReader) ReadByteWithTimeout(timeout time.Duration) (byte, error) {
 }
 
 func (r *bReader) Stop() error {
-	_, err := r.wStop.Write([]byte{'q'})
 	r.mutex.Lock()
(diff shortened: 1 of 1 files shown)


xiaq:
No, this isn't a bug. The empty critical section is useful here, and this PR now introduces a race condition. Instead a comment should be added to suppress this lint check.

krader1961:
@xiaq: Okay, it looks like I need to take a closer look at how the mutex is being used. Presumably you're using it as a way to synchronize with (i.e., wait on) a different goroutine before continuing. If so there either needs to be a detailed "Here be dragons" style comment (here and wherever else the mutex is manipulated), in addition to adding a staticcheck comment to suppress the warning. Or, this should be replaced with a clearer and more straightforward mechanism if possible.

xiaq:
Adding a `//lint:ignore` comment in https://github.com/elves/elvish/commit/e142c6b6a0c23fa687056c49b190bd36d3a0e18a.

Issue #1316, opened 2021-05-21
`file:pipe` should return an object that supports a value channel


krader1961:
Issue #1315 reminded me of a question I've been thinking about for a while: Why doesn't `file:pipe` return an object that supports both byte and value streams? If it did so we could use the `file:pipe` object as a queue between value producers and consumers that were not directly related by a `lhs | rhs` pipeline. Which would allow meaningfully backgrounding value consuming/producing functions that do so by reading from their value input or writing to their value output (see issue #1069) if their was an explicit redirection from an enhanced `file:pipe` object.

There are many questions I don't yet have answers for. Should there always be both a byte and value component to the object or should one or the other have to be explicitly requested so the user has maximum control over resource usage? Should the value channel always be unbuffered or should there be an option for the user to request the amount of buffering? How closely should Elvish adhere to the Go "channel" model? Should the current `file:prclose` and `file:pwclose` affect the value counterpart, or should closing the byte and value components be independent?

krader1961:
> Should the current file:prclose and file:pwclose affect the value counterpart, or should closing the byte and value components be independent?

To answer my own question.... Closing byte and value endpoints should be independent. In no small part because we want to replace `prclose` and `pwclose` with `file:close`, but more generally which endpoint is closed should be under the control of the user and there is no benefit to coupling the ops.

PR #1320, opened 2021-05-23
Replace `prclose` and `pwclose` with `file:close`

krader1961:
Rather than having specialized commands make a `file:pipe` object
indexable so we can use the generic `file:close` command. This does not
address existing problems; such as builtins not failing when writing to
a `file:pipe` object if the read-end is closed.

Related #1316


xiaq (approved):
Thanks for doing change. However, this only implements the Index operation, not IterateKeys or HasKey; since a pipe is nothing more than a read end and a write end, making it a StructMap is the easier way to get these operations implemented on it automatically. I'll do that after merging.

PR #1321, opened 2021-05-24
Fix `sleep` handling of the new `num` type

### LINKED ISSUE #1317: Builtin `sleep` doesn't accept arbitrary `num` types
I was fixing some problems identified by the `staticcheck` Go lint tool and noticed that the builtin `sleep` command only handles string and float64 data types. Other `num` types are not supported but should be. In particular, doing `sleep (num 1)` should work rather than failing with `Exception: invalid sleep duration`.

krader1961:
Fixes #1317

### DIFF
--- a/pkg/eval/builtin_fn_misc_test.go
+++ b/pkg/eval/builtin_fn_misc_test.go
@@ -107,8 +107,11 @@ func TestSleep(t *testing.T) {
 		That(`sleep -7`).Throws(errors.New("sleep duration must be >= zero"), "sleep -7"),
 		That(`sleep -3h`).Throws(errors.New("sleep duration must be >= zero"), "sleep -3h"),
 
-		// Verify the correct behavior if a number rather than a string is
-		// passed to the command.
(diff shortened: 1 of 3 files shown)


pkg/eval/builtin_fn_misc_test.go
```diff
@@ -106,10 +106,11 @@ func TestSleep(t *testing.T) {
 		That(`sleep -7`).Throws(ErrNegativeSleepDuration, "sleep -7"),
 		That(`sleep -3h`).Throws(ErrNegativeSleepDuration, "sleep -3h"),
 
-		// Verify the correct behavior if a number rather than a string is
-		// passed to the command.
-		//
-		// TODO: Augment these tests to use `(num ...)` as part of resolving https://b.elv.sh/1317.
+		That(`sleep 33/3`).Puts(11*time.Second), // rational number string
```
xiaq, line 109: 33/3 is internally normalized to the same as 11. To test a real rational number, use 1/2

Issue #1234, opened 2021-02-11
Increase test coverage to 88% before 0.16.0 release


xiaq:

krader1961:
I was curious why the existing unit tests did not have sufficient coverage to satisfy this issue. That resulted in me noticing that `IsExecutable()` in *pkg/fsutil/search.go* is unused. Not just in a unit test but unused by the elvish program. That function is theoretically useful. Especially in the context of issues like #1135. TBD is whether to retain and use it.

PR #1322, opened 2021-05-25
Increase `eawk` test coverage to 100%

krader1961:
Related #1234

### DIFF
--- a/pkg/eval/builtin_fn_str.go
+++ b/pkg/eval/builtin_fn_str.go
@@ -147,8 +147,9 @@ var eawkWordSep = regexp.MustCompile("[ \t]+")
 // eawk $f $input-list?
 // ```
 //
-// For each input, call `$f` with the input followed by all its fields. The
-// function may call `break` and `continue`.
+// For each input, call `$f` with the input followed by all its fields. A
+// [`break`](./builtin.html#break) command will cause `eawk` to stop processing inputs. A
+// [`continue`](./builtin.html#continue) command is ignored.
 //
 // It should behave the same as the following functions:
 //
--- a/pkg/eval/builtin_fn_str_test.go
+++ b/pkg/eval/builtin_fn_str_test.go
@@ -3,6 +3,7 @@ package eval_test
 import (
 	"testing"
 
+	. "src.elv.sh/pkg/eval"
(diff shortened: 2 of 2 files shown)


pkg/eval/builtin_fn_str.go
```diff
@@ -147,8 +147,9 @@ var eawkWordSep = regexp.MustCompile("[ \t]+")
 // eawk $f $input-list?
 // ```
 //
-// For each input, call `$f` with the input followed by all its fields. The
-// function may call `break` and `continue`.
+// For each input, call `$f` with the input followed by all its fields. A
+// [`break`](./builtin.html#break) command will cause `eawk` to stop processing inputs. A
+// [`continue`](./builtin.html#continue) command is ignored.
```
xiaq, line 152: The exception is ignored, but the command itself is not.

pkg/eval/builtin_fn_str_test.go
```diff
@@ -50,7 +51,35 @@ func TestWcswidth(t *testing.T) {
...
 	Test(t,
+		That(`
+			echo "1\n2\n3\n4" | eawk [@a]{
+				if (==s 3 $a[1]) {
+					fail "stop eawk"
+				}
+				put $a[1]
+			}
+		`).Puts("1", "2").Throws(FailError{"stop eawk"}),
+		That(`put (num 42) | eawk [@a]{ fail "this should not run" }`).
```
xiaq, line 62: Just say `num 42` instead of `put (num 42)`

pkg/eval/builtin_fn_str_test.go
```diff
@@ -50,7 +51,35 @@ func TestWcswidth(t *testing.T) {
 
 func TestEawk(t *testing.T) {
 	Test(t,
+		That(`
+			echo "1\n2\n3\n4" | eawk [@a]{
```
xiaq, line 55: 1. These `echo` commands are worth rewriting to `put line1 line2 | to-lines`; with a lot of "\n" the strings are a bit hard to read.
2. Add comments on which behavior of `eawk` is tested by each test case.

xiaq (approved):

xiaq:
Thanks!

PR #1323, opened 2021-05-26
AltGr handling

### LINKED ISSUE #752: Wrong key parsing with non-US layout on Windows
Reported by SitiSchu on Telegram: on a German keyboard you use AltGr + ß to type a backslash; but Elvish parses this as Ctrl-Alt-[ and complains that it is unbound.

exactly-one-kas:
Should fix #752 at least for German keyboard layouts

### DIFF
--- a/pkg/cli/term/reader_windows.go
+++ b/pkg/cli/term/reader_windows.go
@@ -189,7 +189,7 @@ func convertEvent(event sys.InputEvent) Event {
 			// If only the shift is held down, we try and see if this is a
 			// non-functional key by looking if the rune generated is a
 			// printable ASCII character.
-			if 0x20 <= r && r < 0x7f {
+			if 0x20 <= r && r != 0x7f {
 				return KeyEvent(ui.Key{Rune: r})
 			}
 		} else if filteredMod&(leftCtrl|rightAlt) == leftCtrl|rightAlt && 0x20 <= r && r != 0x7f && currentLayoutHasAltGr() {
--- a/pkg/cli/term/reader_windows.go
+++ b/pkg/cli/term/reader_windows.go
@@ -114,7 +114,7 @@ const (
 	shift     uint32 = 0x10
 )
 
-type HKL (uint32)
+type HKL uint32
 
 type keyboardLayout struct {
 	lastCheck time.Time
--- a/pkg/cli/term/reader_windows.go
+++ b/pkg/cli/term/reader_windows.go
@@ -106,11 +106,11 @@ var keyCodeToRune = map[uint16]rune{
 // A subset of constants listed in
 // https://docs.microsoft.com/en-us/windows/console/key-event-record-str
 const (
-	leftAlt   uint32 = 0x02
-	leftCtrl  uint32 = 0x08
-	rightAlt  uint32 = 0x01
-	rightCtrl uint32 = 0x04
-	shift     uint32 = 0x10
+	leftAlt   = 0x02
+	leftCtrl  = 0x08
+	rightAlt  = 0x01
+	rightCtrl = 0x04
+	shift     = 0x10
 )
 
 // convertEvent converts the native sys.InputEvent type to a suitable Event
@@ -137,10 +137,10 @@ func convertEvent(event sys.InputEvent) Event {
 			if 0x20 <= r && r != 0x7f {
 				return KeyEvent(ui.Key{Rune: r})
 			}
-		} else if filteredMod&(leftCtrl|rightAlt) == leftCtrl|rightAlt && 0x20 <= r && r != 0x7f && currentKeyboardLayoutHasAltGr {
+		} else if filteredMod&(leftCtrl|rightAlt) == leftCtrl|rightAlt && 0x20 <= r && r != 0x7f {
 			// Handle AltGr key combinations if they result in a rune
 			// Shift is also ignored, since it is required for some chars
-			return KeyEvent(ui.Key{Rune: r})
+			return KeyEvent(ui.Key{Rune: r, Mod: ui.Mod(filteredMod &^ (leftCtrl | rightAlt | shift))})
 		}
 		mod := convertMod(filteredMod)
 		if mod == 0 && event.WVirtualKeyCode == 0x1b {
--- a/pkg/cli/term/reader_windows.go
+++ b/pkg/cli/term/reader_windows.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"sync"
+	"syscall"
 	"time"
 
 	"golang.org/x/sys/windows"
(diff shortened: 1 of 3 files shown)


pkg/cli/term/reader_windows.go
```diff
@@ -106,13 +107,67 @@ var keyCodeToRune = map[uint16]rune{
...
-	rightCtrl = 0x04
-	shift     = 0x10
+	leftAlt   uint32 = 0x02
+	leftCtrl  uint32 = 0x08
+	rightAlt  uint32 = 0x01
+	rightCtrl uint32 = 0x04
+	shift     uint32 = 0x10
 )
 
+type HKL (uint32)
```
krader1961, line 117: It doesn't seem like the parens are needed based on my understanding of Go and empirical testing on Windows 10 that shows it compiles and passes existing tests with the parens removed. Were they leftover from a previous version of this change or are they really needed?

pkg/cli/term/reader_windows.go
```diff
@@ -106,13 +107,67 @@ var keyCodeToRune = map[uint16]rune{
...
+}
+
+var (
+	user32                                = syscall.NewLazyDLL("user32.dll")
+	procGetKeyboardLayout                 = user32.NewProc("GetKeyboardLayout")
+	procVkKeyScanExA                      = user32.NewProc("VkKeyScanExA")
+	currentKeyboardLayout *keyboardLayout = nil
+)
+
+const keyboadLayoutCheckTimeout = time.Duration(1e+9) // 1 second
```
krader1961, line 132: Thanks for the comment but I'm still unhappy with the need for a seemingly magic constant (the `1e+9` component). Not to mention that doing the check roughly once a second seems like it is arbitrary and either not quite often enough or too often depending on whether you think people frequently change their keyboard definition or rarely do so. Ideally this would be done via a OS callback that runs whenever the user changes the keyboard definition. For example, does MS Windows provide a Go compatible signal that could be used to change the Elvish keyboard layout only when the user does change the layout as an alternative to polling? Is polling the accepted way to handle this in MS Windows apps?

pkg/cli/term/reader_windows.go
exactly-one-kas: No, that's just because I'm new to Go. I'll remove them

pkg/cli/term/reader_windows.go
exactly-one-kas: > [...] seems like it is arbitrary [...]

It is. Changing the keyboard layout usually isn't done frequently (probably mostly for compatibility reasons with older software). I chose one second to not have too much delay between changing the layout and Elvish noticing the change, but also to not check too frequently.

> [...] does MS Windows provide a Go compatible signal [...]

Windows only provides `WM_INPUTLANGCHANGE`, but since it's a message for the UI subsystem, I'd have to handle another main loop. Do you have a suggestion of which package to use?

krader1961: > Do you have a suggestion of which package to use?

Sorry, no. I haven't done any substantive Windows programming in twenty years.

Do Windows users really change their keyboard layout with any frequency other than configuring it once? Would it make more sense to detect the layout just once in setup_windows.setup() rather than doing it every four or five chars typed by the user (assuming normal typing speeds)?

krader1961: Also, instead of `time.Duration(1e+9)` just use the constant `time.Second`. Assuming you retain this var rather than switch to doing the keyboard layout setup just once.

exactly-one-kas: I personally only change keyboard layouts on accident by pressing `Alt + Shift`. However, I think that this feature has been implemented for a reason.

Then again, the only thing the code does with an AltGr layout is allowing runes to be generated from `Ctrl + Alt + <Key>`, which probably doesn't happen on non-AltGr layouts.

TL;DR: Yes, I think it would suffice to simply detect the layout once on startup. I'll make the change.

pkg/cli/term/setup_windows.go
```diff
@@ -34,6 +43,22 @@ func setup(in, out *os.File) (func() error, error) {
...
+			scancode, _, err := syscall.Syscall(procVkKeyScanExA.Addr(), 2, uintptr(char), layout, 0)
+
+			if err == windows.ERROR_SUCCESS && scancode&0x0600 == 0x0600 {
+				// At least one ASCII char requires CTRL and ALT to be pressed
+				currentKeyboardLayoutHasAltGr = true
+				break
+			}
+		}
+	}
+
```
xiaq, line 61: I don't think Elvish should assume that the user never changes the keyboard layout during a shell session. I believe the check here is unnecessary too; see my other comment.

pkg/cli/term/reader_windows.go
```diff
@@ -106,11 +106,11 @@ var keyCodeToRune = map[uint16]rune{
...
-	leftAlt   = 0x02
-	leftCtrl  = 0x08
-	rightAlt  = 0x01
-	rightCtrl = 0x04
-	shift     = 0x10
+	leftAlt   uint32 = 0x02
+	leftCtrl  uint32 = 0x08
+	rightAlt  uint32 = 0x01
+	rightCtrl uint32 = 0x04
+	shift     uint32 = 0x10
```
xiaq, line 113: Is it necessary to add types to these constants? If not, revert.

pkg/cli/term/setup_windows.go
xiaq: Handling AltGr correctly does not require detecting the layout. I won't go into too much detail here, but this should be quite clear once you understand how reader_windows.go parses the event.

pkg/cli/term/reader_windows.go
exactly-one-kas: At some point I had errors that complained about the missing type annotations, but they're not necessary (anymore). I'll remove them

pkg/cli/term/reader_windows.go
```diff
@@ -134,9 +134,13 @@ func convertEvent(event sys.InputEvent) Event {
...
 			// non-functional key by looking if the rune generated is a
 			// printable ASCII character.
-			if 0x20 <= r && r < 0x7f {
+			if 0x20 <= r && r != 0x7f {
 				return KeyEvent(ui.Key{Rune: r})
 			}
+		} else if filteredMod&(leftCtrl|rightAlt) == leftCtrl|rightAlt && 0x20 <= r && r != 0x7f && currentKeyboardLayoutHasAltGr {
+			// Handle AltGr key combinations if they result in a rune
+			// Shift is also ignored, since it is required for some chars
+			return KeyEvent(ui.Key{Rune: r})
```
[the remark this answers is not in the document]
exactly-one-kas: There are also chars that require shift to be pressed, for example `ẞ` (uppercase ß), which won't be input if I use `filteredMod & ^(leftCtrl|rightAlt)`.

Additionally, I noticed that pressing `AltGr + A` results in `a` being input, which also isn't exactly right.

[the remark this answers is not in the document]
xiaq: > There are also chars that require shift to be pressed, for example ẞ (uppercase ß), which won't be input if I use filteredMod & ^(leftCtrl|rightAlt).

I see, that's quite a bit of finger gymnastics :) But I guess no character would require Ctrl or Alt to input? You can still use `filteredMod & ^(leftCtrl|rightAlt|shift)` as the mod instead of leaving it unset.

> Additionally, I noticed that pressing AltGr + A results in a being input, which also isn't exactly right (the A key has no alternate graving, so this should not create any character input).

Which terminal are you using?

I noticed the same behavior when testing in VS Code's integrated terminal, it's an event identical to a simple press on the `A` key, so there's nothing Elvish can do here - `cmd` also insert `a` when pressing `AltGr-A`.

However, neither Windows's default console nor the Windows Terminal app exhibits this behavior for me. Presumably VS Code does something funny with the keyboard events?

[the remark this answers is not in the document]
exactly-one-kas: I can confirm that the default console doesn't exhibit that behavior, but the VSCode console and wezterm do. I'll just use `filteredMod & ^(leftCtrl|rightAlt|shift)` and leave it at that.

xiaq (approved):
Thanks. I would add some more comments and factor out `0x20 <= r && r != 0x7f` into a function, but I can do that after merging.

Issue #893, opened 2020-01-02
Case insensitive completion


hanche:
Given this setup:

```
edit:completion:matcher[argument] = [seed]{ edit:match-prefix $seed &ignore-case=$true }
```

If I try to complete a filename and there is only one match, but the case of what I typed does not match the filename, I get a completion list containing just one element. For example, after typing `ls doc` (all lower case) and hitting TAB, I get this (`⬥` is my prompt. I removed the right prompt)
```
⬥ ls Documents/
 COMPLETING argument
Documents
```
What I expected is what is seen on the command line above, without the two extra lines and no need to hit RETURN to continue. That is what happens if I type `ls Doc` and hit TAB.

xiaq:
OK, this actually is a feature. Citing https://elv.sh/ref/edit.html#editcompletionsmart-start:

> Starts the completion mode. However, if all the candidates share a non-empty prefix and that prefix starts with the seed, inserts the prefix instead.

There is no special handling for automatically accepting a single candidate; it falls in the general case; when there is only a single candidate, the non-empty prefix is that candidate itself.

However, `edit:completion:smart-start` only inserts the prefix if it also *starts with the seed*. In this case, the seed is `de` (lower case), while the prefix is `Desktop` (with an upper-case D). So the prefix insertion path is not triggered.

---

Consider this case: you have a *substring matcher*, and you have `foo` and `bar` in a directory. Typing `vim o<Tab>` completes to a single candidate `foo`. Since `foo` does not start with the seed `o`, Elvish opens up the completion menu. This is because there is a non-zero chance that you have mistyped and did not mean to complete `foo`. So this is Elvish asking you "are you sure you want to complete `foo`"?

The same can be said for this issue, but the point is a bit weaker. Often we know that `Desktop` starts with a capital `D`, but we are just too lazy to press that Shift key. So when I type `vim de<Tab>`, there is actually a very high chance that I **do** mean to complete `Desktop`.

I think there is a case for `edit:completion:smart-start` to ignore the case or apply smart-case when determining whether prefix completion should be done. It can accept `&smart-case` or `&ignore-case` like the matchers do, and we can default to `&smart-case`.

hanche:
Thanks for the explanation. I don't suppose it would be possible to write elvish code to override the behaviour of `edit:completion:smart-start`? The documentation seems a bit sparse on that point.

xiaq:
No, you can't do that in Elvish code. The Go code needs to be changed.

[...]

PR #1326, opened 2021-05-28
Implement case insensitive command and file completion for Windows

exactly-one-kas:
This changes the following core mechanics:

- Determining file executability: On UNIXoids this still looks at the executable bit. On Windows, it checks the file extension for `.exe`, `.bat` and `.cmd`
- Filtering completion candidates: Prefix checking has been moved into `RawItem` to enable case insensitive matching for commands and files on Windows.
  - This also introduced the option `CaseInsensitive` to `ComplexItem` to facilitate file name matching, but it can be used for various other purposes, too.

Caveat: At least Linux supports mounting with case insensitivity. This hasn't been implemented here, as this would require to somehow check if the directory to be enumerated is mounted that way. It's also probably used very seldomly.

### DIFF
--- a/pkg/edit/complete/util_unix.go
+++ b/pkg/edit/complete/util_unix.go
@@ -1,0 +1,5 @@
+// +build !windows,!plan9
+
+package complete
+
+const IsCaseInsensitiveOs bool = false
--- a/pkg/cli/lscolors/feature.go
+++ b/pkg/cli/lscolors/feature.go
@@ -2,6 +2,8 @@ package lscolors
 
 import (
 	"os"
+
+	"src.elv.sh/pkg/fsutil"
 )
 
 type feature int
@@ -37,8 +39,7 @@ const (
 // Weirdly, permission masks for group and other are missing on platforms other
 // than linux, darwin and netbsd. So we replicate some of them here.
 const (
-	worldWritable = 02   // Writable by other
-	executable    = 0111 // Executable
+	worldWritable = 02 // Writable by other
 )
 
 func determineFeature(fname string, mh bool) (feature, error) {
@@ -105,7 +106,7 @@ func determineFeature(fname string, mh bool) (feature, error) {
 		return featureSetuid, nil
 	case is(m, os.ModeSetgid):
 		return featureSetgid, nil
-	case m&executable != 0:
+	case fsutil.IsExecutableByInfo(stat):
 		return featureExecutable, nil
 	}
 
--- a/pkg/edit/complete/complete.go
+++ b/pkg/edit/complete/complete.go
@@ -50,6 +50,7 @@ type Result struct {
 // RawItem represents completion items before the quoting pass.
 type RawItem interface {
 	String() string
+	HasPrefix(seed string) bool
 	Cook(parse.PrimaryType) modes.CompletionItem
 }
 
--- a/pkg/edit/complete/filterers.go
+++ b/pkg/edit/complete/filterers.go
@@ -1,13 +1,11 @@
 package complete
 
-import "strings"
-
 // FilterPrefix filters raw items by prefix. It can be used as a Filterer in
 // Config.
 func FilterPrefix(ctxName, seed string, items []RawItem) []RawItem {
 	var filtered []RawItem
 	for _, cand := range items {
-		if strings.HasPrefix(cand.String(), seed) {
+		if cand.HasPrefix(seed) {
 			filtered = append(filtered, cand)
 		}
 	}
--- a/pkg/edit/complete/generators.go
+++ b/pkg/edit/complete/generators.go
@@ -58,7 +58,7 @@ func generateCommands(seed string, ev PureEvaler) ([]RawItem, error) {
 	if strings.HasPrefix(seed, "e:") {
 		// Generate all external commands with the e: prefix, and be done.
 		ev.EachExternal(func(command string) {
-			addPlainItem("e:" + command)
+			cands = append(cands, ExternalCommand{Name: command, IsNamespaced: true})
 		})
 		return cands, nil
 	}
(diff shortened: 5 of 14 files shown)


krader1961:
@exactly-one-kas: Is this related to an open issue? In general, I believe significant behavioral changes should have an open issue where discussion about the change in behavior can be discussed. I'm not opposed to this change in behavior but it would be nice to have more context for this change; i.e., what prompted its creation.

> Caveat: At least Linux supports mounting with case insensitivity.

This is true for macOS, Windows, and BSD platforms and probably many others AFAIK. Even if only by virtue of the filesystem type rather than an explicit mount option. It's also unclear why that should affect the behavior of Elvish. It seems to me that whether the underlying filesystem is case sensitive, case-preserving, or case-insensitive should not affect how Elvish performs pathname matching. Which is why I think that these types of changes should, ideally, be discussed in an enhancement issue before a PR is created. I know that many people don't agree with me on this point but I think they're wrong :-)

pkg/edit/complete/util_unix.go
```diff
@@ -0,0 +1,5 @@
+// +build !windows,!plan9
+
+package complete
+
+const IsCaseInsensitiveOs bool = false
```
krader1961, line 5: Ugh! MacOS (a UNIX family OS and which I use on my home server and laptop) is case preserving by default; a special-case of case insensitive. So this is wrong on my systems. It's also why I suggested that this sort of change should be discussed before being implemented. I don't understand why this behavior depends on the platform. Especially since, as you noted in your commit comment, that some platforms may allow mounting a filesystem with case insensitive behavior even if the underlying filesystem is case sensitive.

exactly-one-kas:
> […] it would be nice to have more context for this change; i.e., what prompted its creation.

The context is that matching files and external commands for completion differs in its behavior from the user's expectation for Windows based CLIs. Since the OS is case insensitive by default, executables and files exist in arbitrary casings and are unable to be properly discovered using tab completion.
Missing implementation for determining executability on Windows and using a colon as a path list splitter made it impossible to have external command tab completion at all.

> I don't understand why this behavior depends on the platform.

The following was my reasoning:
- Windows has case insensitivity by default, so it's expected of software running on it to be aware of that
- There were no open issues about file path casing
- Matching external commands straight up didn't work
- I wanted the changes to be non-invasive for other platforms

It seems that `IsCaseInsensitiveOs` is a misnomer, perhaps `IsPrimarilyCaseInsensitiveOs` would be a better fit? Or should I make a configuration variable that the user can override in their `rc.elv` (with the default filled in depending on the OS)?

> […] these types of changes should, ideally, be discussed in an enhancement issue […]

Fair, I might've been a bit over-enthusiastic. But if I decide to make Elvish my daily driver, I'd have implemented case-insensitive matching anyways, even if only in my own fork.

krader1961:
FWIW, @exactly-one-kas, I've been mildly annoyed by the fact that pathname tab completion on macOS is not case-insensitive. Only mildly because the Elvish "completion mode" does show the case-insensitive matches. So `ls /V[Tab]` on macOS immediately expands to `ls /Volumes/`, but `ls /v[Tab]` shows a completion mode list of candidates that includes the single value `/Volumes/`. Personally I'd like the behavior in your change to be the norm on all the UNIX systems I use. I say "the norm" because this may be a rare case where the behavior should be configurable.

exactly-one-kas:
> […] the Elvish "completion mode" does show the case-insensitive matches.

I just tried this again with the current `master` commit and can confirm that this somehow doesn't apply to Windows (at least in my case). Does `HasPrefix` work differently on OSX???

Which of the following changes would you like me to do?
- Make file system completion matching configurable (I'd propose `$edit:completion:case-insensitive-fs` as the variable to configure)
  - Preset the variable according to the OS
- Rename `IsCaseInsensitiveOs` to `IsPrimarilyCaseInsensitiveOs`
- Drop the case insensivity part completely
  - Keep the `CaseInsensitive` option for `ComplexItem`

xiaq:
The change to fix IsExecutable and PATH searching on Windows is good.

The change to make completions case-insensitive is too invasive. Elvish's completion system handles matching by [matchers](https://elv.sh/ref/edit.html#matcher), and it can be set to a case-insensitive one:

```
edit:completion:matcher[argument] = [seed]{ edit:match-prefix $seed &ignore-case=$true }
```

This can probably be made the default on Windows and macOS. The default on other OS'es can be kept unchanged.

One thing that remains the the prefix insertion behavior - when all candidates share a common prefix, [edit:completion:smart-start](https://elv.sh/ref/edit.html#editcompletionsmart-start) will insert that prefix instead. This is implemented [here](https://github.com/elves/elvish/blob/3368744d1b847347b20ca35c639a1bfc65b9e544/pkg/edit/completion.go#L170-L198) and it is always case-sensitive. It is possible to change that to be optionally case-insensitive (controllable by an option), although I have some changes to the matcher system in mind which will solve this in a slightly cleaner way. In the meanwhile this is hopefully only a small annoyance.

xiaq:
Hmm, I overlooked one thing: although the matcher API allows controlling command completion precisely (by setting `edit:completion:matcher[command]`), it doesn't has the same support for file completion; one has to set `edit:completion:matcher[argument]`, but that will apply to other non-file arguments. This is a deficiency of the current API and I'll address it as part of my redesign for the matcher system.

krader1961:
> The change to fix IsExecutable and PATH searching on Windows is good.

I'm going to create a PR that includes only that aspect of this PR since this PR is now nine months old, and @exactly-one-kas has apparently abandoned it. Also, it seems to me Elvish should honor `$E:PATHEXT` if set.  I'll augment the code to do so.

krader1961:
I created PR #1499 to implement the improved completion generation when searching PATH for commands on Windows. The case insensitive aspect of this PR should probably be moved to an issue where the pros and cons of the current implementation can be discussed. Personally, I'm in favor of making the logic case insensitive regardless of platform because a) case preserving but otherwise insensitive is now the norm and b) even on systems which have historically been case sensitive (e.g., legacy UNIX behavior) file system mount options (or similar knobs) can change that behavior.

In light of the above, and because it appears this PR has been abandoned, I think this PR should be closed.

xiaq:
> a) case preserving but otherwise insensitive is now the norm and b) even on systems which have historically been case sensitive (e.g., legacy UNIX behavior) file system mount options (or similar knobs) can change that behavior

Most UNIX systems still default to case-sensitive filesystems; macOS is the odd one out. Maybe what you said will be true in distant future, not right now.

The filesystem question is only tangentially related to whether Elvish should offer case-insensitive matching. Even on case-sensitive filesystems, it may be desirable to do case-insensitive matching. This is already possible with the matcher API, although there is some minor usability issues with that (#893).

Closing this as there's been no activity for a while.

PR #1327, opened 2021-06-04
Fix oldest command being ignored in `edit:command-history &dedup`

mmlb:
The loop iteration was stopping at i == 0 and so allCmds[0] was never
examined.

### DIFF
--- a/pkg/edit/store_api.go
+++ b/pkg/edit/store_api.go
@@ -76,7 +76,7 @@ func dedupCmds(allCmds []store.Cmd, newestFirst bool) []store.Cmd {
 	// Capacity allocation below is based on some personal empirical observation.
 	uniqCmds := make([]store.Cmd, 0, len(allCmds)/4)
 	seenCmds := make(map[string]bool, len(allCmds)/4)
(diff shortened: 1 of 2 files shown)


pkg/edit/store_api.go
```diff
@@ -76,7 +76,7 @@ func dedupCmds(allCmds []store.Cmd, newestFirst bool) []store.Cmd {
 	// Capacity allocation below is based on some personal empirical observation.
 	uniqCmds := make([]store.Cmd, 0, len(allCmds)/4)
 	seenCmds := make(map[string]bool, len(allCmds)/4)
-	for i := len(allCmds) - 1; i != 0; i-- {
+	for i := len(allCmds) - 1; i >= 0; i-- {
```
krader1961, line 79: Great catch of a classic off-by-one bug. I'm doing a "Picard face-palm" at myself 😄 This is also a good example of a fencepost error (a special-case of the off-by-one error).

mmlb: Yes indeed, happens to us all ;)

PR #1329, opened 2021-06-05
Replace ErrArgs with more specific errors

krader1961:
This also replaces the slightly awkward "arguments here" reason with
"argument count" as the "what" for a typical errs.ArityMismatch
exception. It also reformats most of the constructors so that the "what"
is on the same line. This makes `grep errs.ArityMismatch **.go` more
useful as a result.

### DIFF
--- a/pkg/eval/builtin_fn_io.go
+++ b/pkg/eval/builtin_fn_io.go
@@ -17,6 +17,8 @@ import (
 
 // Input and output.
 
+var ErrInvalidTerminator = errors.New("terminator must be a single ASCII char")
+
 func init() {
 	addBuiltinFns(map[string]interface{}{
 		// Value output
@@ -121,7 +123,7 @@ func put(fm *Frame, args ...interface{}) {
 
 func readUpto(fm *Frame, terminator string) (string, error) {
 	if len(terminator) != 1 || terminator[0] > 127 {
-		return "", ErrArgs
+		return "", ErrInvalidTerminator
 	}
 	in := fm.InputFile()
 	var buf []byte
@@ -714,8 +716,6 @@ func fromJSONInterface(v interface{}) (interface{}, error) {
 // ```
 //
 // @cf from-lines read-upto to-terminated
-
-var ErrInvalidTerminator = errors.New("terminator must be a single ASCII char")
 
 func fromTerminated(fm *Frame, terminator string) error {
 	if len(terminator) != 1 || terminator[0] > 127 {
--- a/pkg/eval/builtin_fn_num.go
+++ b/pkg/eval/builtin_fn_num.go
@@ -1,6 +1,8 @@
 package eval
 
 import (
+	"errors"
+	"fmt"
 	"math"
 	"math/big"
 	"math/rand"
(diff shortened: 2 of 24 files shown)


krader1961:
@xiaq, In light of other recent commits I rebased this on master and resolved two merge errors. I also changed the `read-upto` error for an invalid terminator to be the same as the `{from to}-terminated` commands.

xiaq:
Replacing ErrArgs with more specific errors in general is good.

Regarding the change to the `What` field, I'm OK with removing the "here", I was never sure whether it increases or decreases readability.

But adding "count" makes the error messages worse. For example, `errs.ArityMismatch{What: "argument count", ValidLow: 1, ValidHigh: 2, Actual: 3}` reads:

> arity mismatch: argument count must be 1 to 2 values, but is 3 values

"Argument count" is a single value; "arguments" are multiple values, so the logical thing to do is to set `What: "arguments"`.

---

I suppose you were only looking at the field names, not the error message, where it's more logical to contrast "argument count" (instead of "arguments") with `1`, `2`, `3`. This can probably changing the field names; e.g. adding a "Count" suffix - `errs.ArityMismatch{What: "argument count", ValidLowCount: 1, ValidHighCount: 2, ActualCount: 3}`. But this is a bit verbose, so I'm more inclined to just live with the current names. It's the error message that's more important anyway.

krader1961:
> ...so the logical thing to do is to set What: "arguments".

SGTM. Done.

pkg/eval/builtin_fn_io.go
```diff
@@ -121,7 +123,7 @@ func put(fm *Frame, args ...interface{}) {
 
 func readUpto(fm *Frame, terminator string) (string, error) {
 	if len(terminator) != 1 || terminator[0] > 127 {
-		return "", ErrArgs
+		return "", ErrInvalidTerminator
```
xiaq, line 126: This can use errs.BadValue instead to include the actual provided value.

pkg/eval/builtin_fn_num.go
```diff
@@ -588,9 +587,12 @@ func rem(a, b int) (int, error) {
 // ▶ 6
 // ```
 
+var ErrRandIntArgs = errors.New("$low >= $high")
```
xiaq, line 590: This doesn't seem to be used.

pkg/eval/go_fn_test.go
```diff
@@ -50,7 +50,7 @@ func TestGoFnCall(t *testing.T) {
 		t.Helper()
 		err := f.Call(fm, args, opts)
 		if !matchErr(wantErr, err) {
-			t.Errorf("Calling f didn't return error")
+			t.Errorf("Calling f returned wrong error:\nexp %v\ngot %v", wantErr, err)
```
xiaq, line 53: > exp %v

Use "want" for consistency with other test error messages

xiaq (approved):

pkg/eval/builtin_fn_num.go
```diff
@@ -588,9 +587,12 @@ func rem(a, b int) (int, error) {
...
+
 func randint(low, high int) (int, error) {
 	if low >= high {
-		return 0, ErrArgs
+		return 0, errs.BadValue{What: "low value",
```
xiaq, line 594: Since the high value comes after the low value, it's more intuitive to think of the constraint that `high` should be larger than `low` and invert the logic of the error here.

PR #1330, opened 2021-06-05
Increase parse/string.go coverage from 44% to 70%

krader1961:
This is an unusual change in that it exists solely to exercise generated
code and thus increase test coverage. On the one hand, if we never care
about the string value of type `RedirMode` then it should be removed
from the stringified types. On the other hand, verifying a redirection
is correctly represented in the output of the parse.pprint...() family
of functions is useful given the importance of I/O redirections in shells
like Elvish.

### DIFF
--- a/pkg/parse/pprint_test.go
+++ b/pkg/parse/pprint_test.go
@@ -7,7 +7,7 @@ import (
 	"src.elv.sh/pkg/tt"
 )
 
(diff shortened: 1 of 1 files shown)


xiaq:
Well, the `pprint` functions are not user facing, only for debugging test failures, so it doesn't matter that much how well formatted their outputs are. But improving test coverage is fine.

PR #1332, opened 2021-06-06
Increase pkg/eval/vars/env_list.go coverage to 100%

krader1961:
Related #1234

### DIFF
--- a/pkg/eval/vars/env_list.go
+++ b/pkg/eval/vars/env_list.go
@@ -1,5 +1,8 @@
 package vars
 
+// Note: This doesn't have an associated env_list_tests.go because most of its functionality is
+// tested by TestSetEnv_PATH and related tests.
+
 import (
 	"errors"
 	"os"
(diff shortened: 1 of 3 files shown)


pkg/eval/vars/env_list.go
```diff
@@ -18,8 +21,8 @@ var (
 
 // Errors
 var (
-	ErrPathMustBeString           = errors.New("path must be string")
-	ErrPathCannotContainColonZero = errors.New(`path cannot contain colon or \0`)
+	ErrPathMustBeString = errors.New("path must be string")
+	ErrInvalidPathVal   = errors.New(`path cannot contain (semi)colon or \x00`)
```
xiaq, line 25: > (semi)colon

I see how this is an attempt to cover the Windows case, but this is now slightly wrong on both UNIX and Windows; on UNIX semicolons *are* allowed while on Windows colons *are* allowed.

Better to say explicitly which character isn't allowed on which platform.

xiaq (approved):

pkg/eval/vars/env_list.go
xiaq: Also the variable name is unnecessarily broad; `ErrPathContainsForbiddenChar` is better.

krader1961: > this is now slightly wrong on both UNIX and Windows; on UNIX semicolons are allowed while on Windows colons are allowed.

I think you got that backwards but I take your point. I did this because it's unlikely anyone will ever see this error and if they do it seems likely they will understand which of the two chars is relevant. I don't have any objection to specializing the text based on the platform.

PR #1335, opened 2021-06-09
(docs) update README.md

vladdoster:
- fix punctuation

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,14 +8,14 @@
 [![GoDoc](https://img.shields.io/badge/godoc-api-blue.svg)](https://godoc.elv.sh)
 [![Twitter](https://img.shields.io/twitter/url/http/shields.io.svg?style=social)](https://twitter.com/ElvishShell)
 
-Elvish is an expressive programming language and a versatile interactive shell,
-combined into one seamless package. It runs on Linux, BSDs, macOS and Windows.
+Elvish is an expressive programming language and a versatile interactive shell
+combined into one seamless package. It runs on Linux, BSDs, macOS, and Windows.
 
 Despite its pre-1.0 status, it is already suitable for most daily interactive
 use.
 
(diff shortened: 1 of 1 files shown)


xiaq:
I don't think the existing punctuations are objectively wrong?

Oxford commas as a stylistic choice is not universally followed.

vladdoster:
> I don't think the existing punctuations are objectively wrong?
> 
> Oxford commas as a stylistic choice is not universally followed.

They are used in other places throughout the document.

xiaq:
> > I don't think the existing punctuations are objectively wrong?
> > Oxford commas as a stylistic choice is not universally followed.
> 
> They are used in other places throughout the document.

Ha. I went through the doc and there are two places that use Oxford commas, two places that don't, so it's in fact a tie.

My personal default style is to only use Oxford commas when it is actually needed for disambiguating a sentence, which is not the case in any of the uses here. I'll just remove them.

README.md
```diff
@@ -8,14 +8,14 @@
 [![GoDoc](https://img.shields.io/badge/godoc-api-blue.svg)](https://godoc.elv.sh)
 [![Twitter](https://img.shields.io/twitter/url/http/shields.io.svg?style=social)](https://twitter.com/ElvishShell)
 
-Elvish is an expressive programming language and a versatile interactive shell,
```
xiaq, line 11: I'm not sure which style guide you're following, but I'm not aware of any that would ban this particular comma. It doesn't feel "wrong" to me, in fact it makes the otherwise long sentence a bit easier to read.

xiaq:
Unnecessary Oxford commas removed in 9b733b14d86828cd84f3d15e13987f0db7bea80c, so I'll close this.

vladdoster:
No worries @xiaq! Thanks for reviewing my PR. 

I am loving `elivish`!

PR #1337, opened 2021-06-15
Build without support for the interactive daemon

krader1961:
This implements a mechanism for building an elvish program without
support for shared interactive history provided by the interactive
database daemon. This is done via a stub implementation of the daemon
client API. Selected at build time via:

    make ELVISH_MAKE_TAGS=elv_daemon_stub

This does not eliminate the overhead of the Bbolt DB code. That
improvement will require a different set of changes. I wanted to keep
this change small and focused on the daemon.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,7 @@
 ELVISH_MAKE_BIN ?= $(shell go env GOPATH)/bin/elvish
+ifdef ELVISH_MAKE_TAGS
+	override ELVISH_MAKE_TAGS := -tags ${ELVISH_MAKE_TAGS}
+endif
 
 default: test get
 
@@ -8,7 +11,7 @@ get:
 		export GOFLAGS=-buildmode=pie; \
 	fi; \
 	mkdir -p $(shell dirname $(ELVISH_MAKE_BIN))
-	go build -o $(ELVISH_MAKE_BIN) -trimpath -ldflags \
+	go build -o $(ELVISH_MAKE_BIN) $(ELVISH_MAKE_TAGS) -trimpath -ldflags \
 		"-X src.elv.sh/pkg/buildinfo.VersionSuffix=-dev.$$(git rev-parse HEAD)$$(git diff HEAD --quiet || printf +%s `uname -n`) \
 		 -X src.elv.sh/pkg/buildinfo.Reproducible=true" ./cmd/elvish
 
@@ -17,12 +20,12 @@ generate:
 
 # Run unit tests, with race detection if the platform supports it.
 test:
-	go test $(shell ./tools/run-race.sh) ./...
+	go test $(shell ./tools/run-race.sh) $(ELVISH_MAKE_TAGS) ./...
 
 # Generate a basic test coverage report, and open it in the browser. See also
 # https://apps.codecov.io/gh/elves/elvish/.
 cover:
-	go test -coverprofile=cover -coverpkg=./pkg/... ./pkg/...
+	go test -coverprofile=cover -coverpkg=./pkg/... $(ELVISH_MAKE_TAGS) ./pkg/...
 	go tool cover -html=cover
 	go tool cover -func=cover | tail -1 | awk '{ print "Overall coverage:", $$NF }'
 
--- a/README.md
+++ b/README.md
@@ -61,6 +61,12 @@ make get ELVISH_MAKE_BIN=./elvish # Install to the repo root
 make get ELVISH_MAKE_BIN=/usr/local/bin/elvish # Install to /usr/local/bin
 ```
 
+To build it without support for the interactive database daemon:
+
+```sh
+make get ELVISH_MAKE_TAGS=elv_daemon_stub
+```
+
 ## Packaging Elvish
 
 See [PACKAGING.md](PACKAGING.md) for notes for packagers.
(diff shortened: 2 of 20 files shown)


xiaq:
Custom build tags in general are best avoided unless necessary. They introduce another dimension that needs to be tested (both locally and in CI), and they break Go's build cache.

It is entirely feasible to do this using just language constructs (instead of compiler constructs). Elvish already has a mechanism for building different variants with different modules, and that is by defining different main packages under `cmd`. Adding `cmd/nodaemon/elvish/main.go` with the same content as https://github.com/elves/elvish/blob/master/cmd/elvish/main.go, just without the `daemon.Program` will already get rid of the server.

To get rid of the client, use [dependency inversion](https://en.wikipedia.org/wiki/Dependency_inversion_principle):

* Pull the stub implementation of the client into a separate package, say `daemonstub`.

* Make `shell.Program` a type instead of a var, and let it keep the `NewClient` and `Spawn` callbacks

  ```go
  package shell
  
  type Program {
    SpawnDaemon(*SpawnCfg) error
    NewDaemonClient(string) Client
  }
  ```

* In the default entry point (cmd/elvish), populate those callbacks with the non-stub version.

* In the no-daemon variant (cmd/nodaemon/elvish), populate those callbacks with the stub version.

krader1961:
> ... and they break Go's build cache.

Yes, but that isn't really an issue for this change in functionality. The only people using this will be "busybox" type projects like u-root and they're not going to be building both the stub and normal variants.

When I began tackling this I intended to use the dependency inversion approach you suggested. I decided that was more work than it's worth. In no small part because I hope the daemon is eliminated in the not too distant future.

While my approach "introduce(s) another dimension that needs to be tested" I'll point out that AFAICT we don't explicitly test any of the *./cmd/* variants; e.g., `./cmd/withweb/...`. My approach actually makes it easy to run all the unit tests that are known not to depend on a working daemon by simply doing `make test ELVISH_MAKE_TAGS=elv_daemon_stub` without having to explicitly test a hypothetical `./cmd/nodaemon/...` target. I modified my local VM test runner to support that and verified every test that isn't explicitly excluded due to the lack of a daemon passes on macOS, Linux, FreeBSD, and Windows. Too, running the resulting no-daemon binary by hand behaves as expected; whether interactive or not.

Finally, using custom build tags is potentially useful for creating other custom elvish variants. We might add a feature, such as a `net` module, that something like the u-root project deems unacceptable and won't want to include in their build. We can certainly debate whether supporting such variants is desirable but custom build tags is arguably the easiest way to do so.

xiaq:
> Yes, but that isn't really an issue for this change in functionality. The only people using this will be "busybox" type projects like u-root and they're not going to be building both the stub and normal variants.

True, but I consider my responsibility to not break the build for them, so realistically I will have to run tests locally with and without the custom build flag.

> When I began tackling this I intended to use the dependency inversion approach you suggested. I decided that was more work than it's worth. In no small part because I hope the daemon is eliminated in the not too distant future.

Maybe, but I don't want to count on a future project to justify doing something sub-optimally now.

> While my approach "introduce(s) another dimension that needs to be tested" I'll point out that AFAICT we don't explicitly test any of the _./cmd/_ variants; e.g., `./cmd/withweb/...`.

They don't have tests now, but `go test ./...` ensures that they at least build. With a custom build tag, you'll need to run any build or test command twice.

> My approach actually makes it easy to run all the unit tests that are known not to depend on a working daemon by simply doing `make test ELVISH_MAKE_TAGS=elv_daemon_stub` without having to explicitly test a hypothetical `./cmd/nodaemon/...` target. I modified my local VM test runner to support that and verified every test that isn't explicitly excluded due to the lack of a daemon passes on macOS, Linux, FreeBSD, and Windows. Too, running the resulting no-daemon binary by hand behaves as expected; whether interactive or not.

Yes, this is an advantage with the build tag approach in theory - it is easy to run all the tests under the two different implementations. However, the advantage is only real if most of the tests actual depend on the concrete implementation.

There are two packages that depend on `pkg/daemon`, `pkg/eval` and `pkg/shell`. The former does not depend on the concrete implementation, only on the interface. The latter does depend on the concrete implementation, but most of the tests are written in such a way to actually avoid invoking the daemon; the only test that does is [this one](https://github.com/elves/elvish/blob/f3c08d78b24251b36edfb57acad799bc0eae5763/pkg/shell/shell_unix_test.go#L52-L58), with an explicit `SpawnDaemon: true`; other tests leave out this config which suppresses the daemon.

With dependency inversion, it is still possible to test the two different implementations by explicit injection; it is a bit more verbose, but still doable, and has the advantage that it doesn't require running `go test` twice with different tags. Custom build tags also break Go's test cache.

> Finally, using custom build tags is potentially useful for creating other custom elvish variants. We might add a feature, such as a `net` module, that something like the u-root project deems unacceptable and won't want to include in their build. We can certainly debate whether supporting such variants is desirable but custom build tags is arguably the easiest way to do so.

That can be achieved by dependency inversion too.

---

In conclusion, using custom build tags in this way won't fly with me. I guess you don't like dependency inversion, but that ship sailed for me a long time ago.

Also, come to think of it, I don't think implementing this requires a stub implementation of the daemon at all; the client could just be left out as `nil`, and places that expect a daemon client in `pkg/shell` and `pkg/eval` already deals with a nil client.

d6464280f27b8ff359c63eea217d95d77cb35ff6 implemented this with dependency inversion, so I'm going to close this PR now.

PR #1338, opened 2021-06-15
Add plugin support

kolbycrouch:
Adds golang plugin support to the `use` keyword.

With this change, `use` prioritizes loading modules with the extension `.so`.

A plugin module must contain a symbol called `Ns` of type `*eval.Ns`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,10 @@
 ELVISH_MAKE_BIN ?= $(shell go env GOPATH)/bin/elvish
+ELVISH_PLUGINS ?= 0
 
 default: test get
 
 get:
-	export CGO_ENABLED=0; \
+	export CGO_ENABLED=$(ELVISH_PLUGINS); \
 	if go env GOOS GOARCH | egrep -qx '(windows .*|linux (amd64|arm64))'; then \
 		export GOFLAGS=-buildmode=pie; \
 	fi; \
--- a/README.md
+++ b/README.md
@@ -61,6 +61,12 @@ make get ELVISH_MAKE_BIN=./elvish # Install to the repo root
 make get ELVISH_MAKE_BIN=/usr/local/bin/elvish # Install to /usr/local/bin
 ```
 
+To install with plugin support, override `ELVISH_PLUGINS` in the `make` command:
+
+```sh
+make get ELVISH_PLUGINS=1 # Install with plugin support
+```
+
 ## Packaging Elvish
 
 See [PACKAGING.md](PACKAGING.md) for notes for packagers.
--- a/pkg/eval/builtin_special.go
+++ b/pkg/eval/builtin_special.go
@@ -345,7 +345,7 @@ func use(fm *Frame, spec string, r diag.Ranger) (*Ns, error) {
 				return nil, err
 			}
 		}
-		path := filepath.Clean(dir + "/" + spec + ".elv")
+		path := filepath.Clean(dir + "/" + spec)
 		return useFromFile(fm, spec, path, r)
 	}
 	if ns, ok := fm.Evaler.modules[spec]; ok {
@@ -359,7 +359,7 @@ func use(fm *Frame, spec string, r diag.Ranger) (*Ns, error) {
 	if libDir == "" {
 		return nil, noSuchModule{spec}
 	}
-	return useFromFile(fm, spec, libDir+"/"+spec+".elv", r)
+	return useFromFile(fm, spec, libDir+"/"+spec, r)
 }
 
 // TODO: Make access to fm.Evaler.modules concurrency-safe.
@@ -367,14 +367,29 @@ func useFromFile(fm *Frame, spec, path string, r diag.Ranger) (*Ns, error) {
 	if ns, ok := fm.Evaler.modules[path]; ok {
 		return ns, nil
 	}
-	code, err := readFileUTF8(path)
+	_, err := os.Stat(path + ".so")
 	if err != nil {
-		if os.IsNotExist(err) {
+		code, err := readFileUTF8(path + ".elv")
+		if err != nil {
+			if os.IsNotExist(err) {
+				return nil, noSuchModule{spec}
+			}
+			return nil, err
+		}
+		return evalModule(fm, path, parse.Source{Name: path, Code: code, IsFile: true}, r)
+	} else {
+		plug, err := pluginOpen(path + ".so")
+		if err != nil {
 			return nil, noSuchModule{spec}
 		}
-		return nil, err
+		sym, err := plug.Lookup("Ns")
+		if err != nil {
+			return nil, err
+		}
+		ns := sym.(**Ns)
+		fm.Evaler.modules[path] = *ns
+		return *ns, nil
 	}
-	return evalModule(fm, path, parse.Source{Name: path, Code: code, IsFile: true}, r)
 }
 
 // TODO: Make access to fm.Evaler.modules concurrency-safe.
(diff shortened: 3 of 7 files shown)


krader1961:
Maybe document how to build an Elvish compatible plugin?

Makefile
```diff
@@ -1,9 +1,10 @@
 ELVISH_MAKE_BIN ?= $(shell go env GOPATH)/bin/elvish
+ELVISH_PLUGINS ?= 0
 
 default: test get
 
 get:
-	export CGO_ENABLED=0; \
+	export CGO_ENABLED=$(ELVISH_PLUGINS); \
```
xiaq, line 7: The `go build` line below also needs to be updated to set `buildinfo.Reproducible=false` when cgo is enabled; cgo builds are in general not reproducible. You can leave this as is and I can handle that change.

README.md
```diff
@@ -61,6 +61,12 @@ make get ELVISH_MAKE_BIN=./elvish # Install to the repo root
 make get ELVISH_MAKE_BIN=/usr/local/bin/elvish # Install to /usr/local/bin
 ```
 
+To install with plugin support, override `ELVISH_PLUGINS` in the `make` command:
+
+```sh
+make get ELVISH_PLUGINS=1 # Install with plugin support
+```
+
```
xiaq, line 69: I'd also add instructions of how to build such plugins; you can leave it as is.

pkg/eval/builtin_special.go
```diff
@@ -345,7 +346,13 @@ func use(fm *Frame, spec string, r diag.Ranger) (*Ns, error) {
 				return nil, err
 			}
 		}
-		path := filepath.Clean(dir + "/" + spec + ".elv")
+		var path string
+		_, plugin := os.Stat(filepath.Clean(dir + "/" + spec + ".so"))
```
xiaq, line 350: Name error return value `err`, not something else; this is a strong convention followed in the codebase.

pkg/eval/builtin_special.go
```diff
@@ -359,22 +366,43 @@ func use(fm *Frame, spec string, r diag.Ranger) (*Ns, error) {
...
 	}
-	return useFromFile(fm, spec, libDir+"/"+spec+".elv", r)
+	var path string
+	_, plugin := os.Stat(libDir + "/" + spec + ".so")
+	if os.IsNotExist(plugin) {
+		path = libDir + "/" + spec + ".elv"
+	} else {
+		path = libDir + "/" + spec + ".so"
+	}
+	return useFromFile(fm, spec, path, r)
```
xiaq, line 376: This is duplicate with the code above.

Since both places call `useFromFile` and these are the only call sites of `useFromFile`, one could change the semantics of the `path` argument to not include the extension, and let `useFromFile` determine whether to use the `.so` file or the `.elv` file. It'd also eliminate the need for the `strings.HasSuffix` check in `useFromFile`.

pkg/eval/builtin_special.go
```diff
@@ -359,22 +366,43 @@ func use(fm *Frame, spec string, r diag.Ranger) (*Ns, error) {
...
+		if err != nil {
 			return nil, noSuchModule{spec}
 		}
-		return nil, err
+		sym, err := plug.Lookup("Ns")
+		if err != nil {
+			return nil, err
+		}
+		var ns **Ns
+		ns = sym.(**Ns)
```
xiaq, line 391: No need for the `var` declaration here:

```go
ns := sym.(**Ns)
```

pkg/eval/builtin_special.go
```diff
@@ -359,22 +366,43 @@ func use(fm *Frame, spec string, r diag.Ranger) (*Ns, error) {
...
+		return *ns, nil
+	} else {
+		if ns, ok := fm.Evaler.modules[path]; ok {
+			return ns, nil
+		}
```
xiaq, line 396: These two lines should remain at the beginning of the function - `.go` modules should be only loaded once too, like `.elv` modules.

pkg/eval/builtin_special.go
```diff
@@ -359,22 +366,43 @@ func use(fm *Frame, spec string, r diag.Ranger) (*Ns, error) {
...
 			return nil, noSuchModule{spec}
 		}
-		return nil, err
+		sym, err := plug.Lookup("Ns")
+		if err != nil {
+			return nil, err
+		}
+		var ns **Ns
+		ns = sym.(**Ns)
+		return *ns, nil
```
xiaq, line 392: Before returning, save the module:

```go
fm.Evaler.modules[path] = *ns
```

As for the key here, it should be the extensionless name of the module (so e.g. `foo/bar`, not `foo/bar.so`).

website/ref/language.md
```diff
@@ -2227,6 +2227,9 @@ fn f {
 }
 ```
 
+Note: If you build elvish with `ELVISH_PLUGINS=1` you can use go plugins with
+the extension `.so`.
+
```
xiaq, line 2232: I'd expand the explanation here, but you can leave it as is in the PR.

pkg/eval/builtin_special.go
```diff
@@ -19,6 +19,7 @@ package eval
 import (
 	"os"
 	"path/filepath"
+	"plugin"
```
xiaq, line 22: This breaks the build on gccgo, which lacks the `plugin` package. It'd be nicer if gccgo has a `plugin` package with a stub implementation (I asked for that in https://github.com/golang/go/issues/36403#issuecomment-864474859), but for now, you'll need to manually stub out the plugin API like this:

plugin.go:

```go
//+build !gccgo

package eval

import "plugin"

var pluginOpen = plugin.Open
```

plugin_gccgo.go:

```go
//+build gccgo

package eval

import "errors"

var errPluginNotImplemented = errors.New("plugin not implemented")

type pluginStub struct{}

func pluginOpen(name string) (pluginStub, error) {
  return pluginStub{}, errPluginNotImplemented
}

func (pluginStub) Lookup(symName string) (interface{}, error) {
  return nil, errPluginNotImplemented
}
```

kolbycrouch:
Hopefully this is what you had in mind.

Yea it definitely needs more documenting.

My problem was that a plugin either needs to reside inside the same directory structure as its `src.elv.sh` import
or you need a go.mod file.

If you make a go.mod file (the proper thing to do AFAICT) you need to setup replacements I think if your using a fork
of the desired import.

So if I documented exactly how to set that up it would be hard to test since the change isn't upstream, and the plugin
has to built against exactly the same version of the elvish that loads it.

xiaq:
Thanks, code LGTM!

> My problem was that a plugin either needs to reside inside the same directory structure as its `src.elv.sh` import
> or you need a go.mod file.
> 
> If you make a go.mod file (the proper thing to do AFAICT) you need to setup replacements I think if your using a fork
> of the desired import.

Hmm, I don't see why it's necessary to set up replacement. You just need to ensure that you're using the same Elvish version as declared in `go.mod`. I'll add some examples, which hopefully makes it clear.

kolbycrouch:
I read that you need to setup replacements in the go.mod if you want to depend on a fork of something imported with the import statement. So until the change was upstream the go.mod needed to use my repo to satisfy `src.elv.sh` imports. That was my understanding when I was trying to setup an example.

xiaq:
> I read that you need to setup replacements in the go.mod if you want to depend on a fork of something imported with the import statement. So until the change was upstream the go.mod needed to use my repo to satisfy `src.elv.sh` imports. That was my understanding when I was trying to setup an example.

Right, you would have needed that before this PR was merged, but you shouldn't need that anymore - I published an example at https://github.com/elves/sample-plugin.

PR #1339, opened 2021-06-17
`make cover` now ignores same files as Codecov

krader1961:
Running `make cover` includes code in its report that is explicitly
excluded from the https://codecov.io/gh/elves/elvish/ report. This change
causes both reports to include/exclude the same source files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ test:
 # https://apps.codecov.io/gh/elves/elvish/.
 cover:
 	go test -coverprofile=cover -coverpkg=./pkg/... ./pkg/...
+	tools/cover-prune.sh .codecov.yml cover
 	go tool cover -html=cover
 	go tool cover -func=cover | tail -1 | awk '{ print "Overall coverage:", $$NF }'
 
--- a/tools/cover-prune.sh
+++ b/tools/cover-prune.sh
@@ -1,0 +1,23 @@
+#!/bin/sh -e
+
+# Prune the same objects from the "make cover" report that we tell Codecov
+# (https://codecov.io/gh/elves/elvish/) to ignore.
+
+if test $# != 2
+then
+    echo 'Usage: cover_prune.sh ${codecov.yml} $cover' >&2
+    exit 1
+fi
+yaml="$1"
+data="$2"
+
+# This approach to ignoring code from the "make cover" report is suboptimal
+# but efficient enough for our purposes. Especially when weighed against a
+# more complicated approach that constructs a single regexp that is used to
+# filter the data just once since the number of exclusions is small.
+sed -ne '/^ignore:/,/^[^ \t]/s/^[ \t]*- "\(.*\)"/\1/p' $yaml |
+    while read pattern
+    do
+        grep -v "$pattern" $data > $data.tmp
+        mv $data.tmp $data
+    done


krader1961:
This increases `make cover` overall coverage 88.4% => 90.2% on macOS. TBD is why Codecov has a dramatically lower, 85.3%, overall coverage. But this does, AFAICT, result in both reports measuring the same set of source files.

krader1961:
> TBD is why Codecov has a dramatically lower, 85.3%, overall coverage.

The answer to my previous question appears to be a bug, possibly an off by one bug, in `go tool cover`. The Codecov value of this file is 96.55%: https://codecov.io/gh/elves/elvish/src/master/pkg/cli/tk/colview.go. The `make cover` value is 100%. The difference is because the go coverage tool apparently considers the two lambdas assigned to `spec.OnLeft` and `spec.OnRight` to have been executed. Which is not only wrong but unexpected since the coloring of those lines has the left-paren in green (representing the definition) and the right-paren in red (since the body was not executed). Adding a `panic()` call to each lambda does not result in any test failures. Adding a `panic()` to a related lambda that Codecov marks as fully covered results in a test failure when running `make cover`. The raw coverage data in the *./cover* file is consistent with those lambda bodies never being executed:

```
src.elv.sh/pkg/cli/tk/colview.go:62.24,63.31 1 0
src.elv.sh/pkg/cli/tk/colview.go:63.32,63.33 0 0
src.elv.sh/pkg/cli/tk/colview.go:65.25,66.32 1 0
src.elv.sh/pkg/cli/tk/colview.go:66.33,66.34 0 0
```

Changing those empty lambdas to include a `panic("WTF")` call changes the `make cover` result so it matches the Codecov value. It also shows that those lambdas are not executed since the panics never occur when executing the tests.

Having said the above I still think this filtering of the data used by `make cover` is a good idea. The `make cover` report should exclude the same files as the Codecov report.

krader1961:
I opened a Go issue about the incorrect treatment of empty lamdas: https://github.com/golang/go/issues/46825. TBD is what, if anything, we want to do about this since a fix from the Go project is unlikely to be available until far into the future. We can always add dummy `Nop()` calls; although anything but an empty lambda fixes the coverage percentage. It's probably better just to live with the difference and note that the coverage would be increased if empty lambdas were actually executed by unit tests.

krader1961:
And, unsurprisingly, the aforementioned Go coverage bug only accounts for 0.1% of the difference compared to the Codecov values. The much more significant reason for the difference in coverage percentage is that Go considers each branch to be fully covered if any branch is taken; whereas Codecov considers each branch independently. Which metric is more useful is a philosophical question. Obviously the Go and Codecov teams disagree on which is more useful.

A concrete example is: https://codecov.io/gh/elves/elvish/src/master/pkg/eval/vals/aliased_types.go. Which shows 75% coverage while the Go toolchain shows 90%. The difference is because this expression only evaluates to false during testing:

```go
if len(a)%2 == 1 {
```

The Go toolchain considers that single outcome to have fully covered the expression. While Codecov treats the true and false outcomes separately and thus the expression was only 50% covered.

Changing the `make cover` behavior to address the empty lambda bug turns out to be trivial; even without permanently modifying the source. A simple `sed` command included in the `make cover` command sequence "fixes" the related metric:

```
sed -i -e 's/) {}/) { func(){}() }/' **.go
```

But since that represents an insignificant portion of the difference between the Go and Codecov metrics it isn't worth even that trivial cost. Especially since it becomes more complicated when you want to support `make cover` with uncommitted changes.

This [stackoverflow](https://stackoverflow.com/questions/37363691/getting-branch-coverage-for-golang-tests) question is relevant. As is this [Go issue](https://github.com/golang/go/issues/28888). In that issue Rob Pike said:

> It was a consequence of the design, not deliberate and not accidental. I think it's fine and would prefer to leave things as they are. The simplicity of the current implementation is worth a lot.

I can't figure out if the Go team is open to the idea of supporting more pessimistic coverage metrics based on some branches not being taken. Also, it's not obvious that Codecov does the right thing with respect to more complex expressions; e.g., `if x && y && z {` for all 2^3 combinations. Which argues for the simpler approach taken by the Go tool chain.

krader1961:
Consider the coverage of this file: https://codecov.io/gh/elves/elvish/src/master/pkg/eval/vals/aliased_types.go

Codecov says it is 75% while Go says it is 90%. The difference is because only the false branch of `len(a)%2 == 1` is taken. In practice we don't really care if the panic path is taken during unit tests. I suspect a majority of the differences in coverage are due to similar situations. So I'm inclined to use the Go toolchain coverage metrics for deciding whether there is adequate coverage and the Codecov metrics for where small, incremental, improvements can be made in test coverage.

xiaq (approved):
I've mentioned the divergence in branch coverage before, which is a more common issue and likely more impactful than lambda coverage: https://github.com/elves/elvish/issues/1062#issuecomment-761516167. Judging from https://github.com/golang/go/issues/28888#issuecomment-440463919, which you referenced, it's likely that the Go team will prefer to keep their simplistic calculation algorithm in the case of lambda coverage too.

I'll merge this PR anyway as it is still a net improvement, but unless one re-implements Codecov's post-processing algorithm, the coverage data from `make cover` will never match Codecov's data and one will always have to rely on Codecov.

PR #1342, opened 2021-06-22
Generalize the places that handle `case float64:` to other number types where appropriate

krader1961:


pkg/eval/builtin_fn_io.go
```diff
@@ -673,6 +673,7 @@ func fromJSONInterface(v interface{}) (interface{}, error) {
 	case nil, bool, string:
 		return v, nil
 	case float64:
+		// TODO: Decide if we want to normalize ints here.
```
xiaq, line 676: JSON only has one number type and it corresponds to Go's float64; nothing can be done here to recover the "original" type.

pkg/eval/builtin_fn_styled_test.go
```diff
@@ -68,17 +68,26 @@ func TestStyledConcat(t *testing.T) {
 		That("print (styled-segment abc &bg-color=red)(styled-segment abc &fg-color=red)").Prints("\033[41mabc\033[m\033[31mabc\033[m"),
 		// segment+text
 		That("print (styled-segment abc &underlined=$true)(styled abc bright-cyan)").Prints("\033[4mabc\033[m\033[96mabc\033[m"),
-	)
-
-	Test(t,
+		// segment+num
+		That("print (float64 99)(styled-segment abc &blink)").Prints("99.0\033[5mabc\033[m"),
+		That("print (num 66)(styled-segment abc &blink)").Prints("66\033[5mabc\033[m"),
+		That("print (num 33)(styled-segment abc &blink)").Prints("33\033[5mabc\033[m"),
```
xiaq, line 74: Both cases test the `case int` branch. Did you mean to make one of them test a rational number?

pkg/eval/vals/equal_test.go
```diff
@@ -22,6 +23,14 @@ func TestEqual(t *testing.T) {
 		Args(true, false).Rets(false),
 
 		Args(1.0, 1.0).Rets(true),
+		Args(1.0, 1.1).Rets(false),
+		Args("1.0", 1.0).Rets(false),
+		Args(1, 1.0).Rets(false),
+		Args(1, 1).Rets(true),
+		Args(big.NewInt(1), big.NewInt(1)).Rets(true),
+		Args(big.NewInt(1), 1).Rets(false),
```
xiaq, line 31: Internally, Elvish never represents integers within the range of `int` as a `*big.Int`; this should test some actual big numbers; num_test.go has some utilities for doing this.

pkg/eval/vals/hash.go
```diff
@@ -24,6 +24,7 @@ func Hash(v interface{}) uint32 {
 			return 1
 		}
 		return 0
+	// TODO: Add support for the other num types: int, *big.Int, *big.Rat.
```
xiaq, line 27: Oops, that's an oversight on my part. I'll implement these.

pkg/eval/vals/string.go
```diff
@@ -20,6 +20,8 @@ func ToString(v interface{}) string {
 	case int:
 		return strconv.Itoa(v)
 	case float64:
+		// Float64 is special-cased for historical reasons. The other number types are handled by
+		// the Stringer interface.
```
xiaq, line 24: Not really for historical reasons; it's because I don't like any of Go's builtin ways to print float64. The point about other number types are correct.

xiaq (approved):

PR #1348, opened 2021-06-30
Add buildinfo and version vars

krader1961:
Introduce `builtin:buildinfo` and `builtin:version` vars. This also
changes the `-buildinfo -json` implementation to use the standard
encoding/json package rather than handcrafting the JSON string.

This also fixes three incorrect spellings of "overridden". Normally I would
do those in a separate change but since there are only three instances,
one of which one was legitimately part of this change, I decided to bundle
the other two.


pkg/buildinfo/buildinfo.go
```diff
@@ -15,38 +16,58 @@ import (
...
 type program struct{}
 
 func (program) ShouldRun(f *prog.Flags) bool { return f.Version || f.BuildInfo }
 
-func (program) Run(fds [3]*os.File, f *prog.Flags, _ []string) error {
-	fullVersion := Version + VersionSuffix
-	if f.Version {
-		fmt.Fprintln(fds[1], fullVersion)
-		return nil
+func GetBuildInfo() Buildinfo {
```
xiaq, line 44: This needs to have a godoc, and it can be a simple `var` declaration instead of a function.

(If you're worried that the overridden values won't be picked up by a `var` declaration, those overrides happen at link time and directly modify the initial value of the variables.)

pkg/buildinfo/buildinfo.go
```diff
@@ -15,38 +16,58 @@ import (
...
-		fmt.Fprintln(fds[1], "Go version:", runtime.Version())
-		fmt.Fprintln(fds[1], "Reproducible build:", Reproducible)
+}
+
+func (program) Run(fds [3]*os.File, f *prog.Flags, _ []string) error {
+	bi := GetBuildInfo()
+	switch {
+	case f.JSON:
+		// Note: The only way this will be executed is if option -buildinfo is also present.
+		// See the ShouldRun() method above.
```
xiaq, line 57: > // Note: The only way this will be executed is if option -buildinfo is also present.
> // See the ShouldRun() method above.

That's not true? `-json -version` will also trigger this code path, and this PR changes the behavior from printing the version string to printing the entire buildinfo. Neither of them is correct; the correct behavior should be printing out the version string as a JSON-encoded string.

website/ref/language.md
```diff
@@ -378,19 +378,24 @@ lorem
...
-A pseudo-map typically has a fixed set of keys, called **fields**. The values of
-a field can be accessed by [indexing](#indexing). However, unlike maps, it is
-usually not possible to add new keys, remove existing keys, or even assigning a
-new value to an existing key.
+A pseudo-map has a fixed set of keys whose values can be accessed by
+[indexing](#indexing) like you would for a regular [map](#map). Similarly, you
+can use commands like [`keys`](./builtin.html#keys) and
+[`has-key`](./builtin.html#keys) on such objects. However, unlike a normal map
+it is not possible to add new keys, remove existing keys, or even assign a new
+value to an existing key. In other words, a pseudo-map is immutable.
```
xiaq, line 394: > In other words, a pseudo-map is immutable.

This is potentially misleading because normal maps are also immutable. The "mutation" operation supported by normal maps create copies.

xiaq (approved):

pkg/buildinfo/buildinfo.go
krader1961: The new implementation behaves the same as the old implementation. I don't think there is a good reason to have `elvish -json -version` emit the version as a JSON encoded string. Consider that doing so simply turns the value into a quoted string:

```
> put $version | to-json
"0.16.0-dev.bfb990f932611060edc198886396076004e48c53"
```

pkg/buildinfo/buildinfo.go
krader1961: Good point. This is a function because an earlier iteration of this change had more complicated logic in a misguided attempt to normalize the components of the full version string.

website/ref/language.md
krader1961: Yes, normal maps are also technically immutable but can be mutated using `assoc` explicitly or implicitly via magic LHS syntax. A pseudo-map cannot be mutated by `assoc`. That subtlety should probably be clarified but I couldn't see a way to word this that didn't hide the most important point of this paragraph.

pkg/buildinfo/buildinfo.go
xiaq: Probably not very useful but consistent.

website/ref/language.md
xiaq: This concept is sometimes called "nondestructive mutation" but that concept would require its own paragraph to explain. The doc now says "it is currently not possible to create a modified version of an existing pseudo-map".

PR #1349, opened 2021-06-30
Fix `go vet` error about unreachable code

krader1961:


krader1961:
We probably do not want to predicate every `make test` on `go vet` not reporting errors. It would add a non-negligible cost to each `make test` or `make get` (2.2 seconds, best case, on my main server). We could add an explicit `go vet` to the github/gitlab test workflows but that wouldn't catch mistakes made by people able to merge changes before those test workflows runs. 😸

P.S., The sole test failure has nothing to do with this change, AFAICT:

```
ok  	src.elv.sh/pkg/buildinfo	0.026s
--- FAIL: TestReadCode_DoesNotReadMoreEventsThanNeeded (0.00s)
    app_test.go:520: got event <nil>, want {98 0}
FAIL
FAIL	src.elv.sh/pkg/cli	0.074s
```

xiaq:
Hmm, it's not the first time I saw that test fail on the FreeBSD CI environment. I'll try to reproduce it on my FreeBSD VM later.

PR #1354, opened 2021-07-04
Lint cleanups

krader1961:
This change addresses issues reported by the `staticcheck` tool and a
couple reported by the `golint` tool. It also adds missing issue links
to a couple of entries in the release notes. This change deliberately
does not address these warnings since it is unclear whether the project
owner would prefer to suppress or address them:

    pkg/store/cmd.go:8:2: should not use dot imports (ST1001)
    pkg/store/db_store.go:10:2: should not use dot imports (ST1001)
    pkg/store/dir.go:8:2: should not use dot imports (ST1001)

### DIFF
--- a/pkg/eval/external_cmd_unix_internal_test.go
+++ b/pkg/eval/external_cmd_unix_internal_test.go
@@ -11,23 +11,22 @@ import (
 	"src.elv.sh/pkg/env"
 	"src.elv.sh/pkg/parse"
 	"src.elv.sh/pkg/testutil"
-	. "src.elv.sh/pkg/testutil"
 )
 
 func TestExec_Argv0Argv(t *testing.T) {
-	dir, cleanupFS := InTestDir()
+	dir, cleanupFS := testutil.InTestDir()
 	defer cleanupFS()
-	ApplyDir(Dir{
-		"bin": Dir{
-			"elvish": File{Perm: 0755},
-			"cat":    File{Perm: 0755},
+	testutil.ApplyDir(testutil.Dir{
+		"bin": testutil.Dir{
+			"elvish": testutil.File{Perm: 0755},
+			"cat":    testutil.File{Perm: 0755},
 		},
 	})
 
-	restorePATH := WithTempEnv("PATH", dir+"/bin")
+	restorePATH := testutil.WithTempEnv("PATH", dir+"/bin")
 	defer restorePATH()
 
-	restoreSHLVL := WithTempEnv(env.SHLVL, "1")
+	restoreSHLVL := testutil.WithTempEnv(env.SHLVL, "1")
 	defer restoreSHLVL()
 
 	var tests = []struct {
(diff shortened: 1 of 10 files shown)


pkg/eval/external_cmd_unix_internal_test.go
```diff
@@ -11,23 +11,22 @@ import (
...
+			"cat":    testutil.File{Perm: 0755},
 		},
 	})
 
-	restorePATH := WithTempEnv("PATH", dir+"/bin")
+	restorePATH := testutil.WithTempEnv("PATH", dir+"/bin")
 	defer restorePATH()
 
-	restoreSHLVL := WithTempEnv(env.SHLVL, "1")
+	restoreSHLVL := testutil.WithTempEnv(env.SHLVL, "1")
```
xiaq, line 29: I prefer keeping the dot import on testutil.

Staticcheck has a [`dot_import_whitelist` flag](https://staticcheck.io/docs/options#dot_import_whitelist) that allows certain packages to be dot-imported. I plan to add staticcheck as part of CI and a Makefile target and I'll add the `testutil` package to that list.

pkg/eval/vals/num.go
```diff
@@ -201,13 +197,11 @@ func ConvertToFloat64(num Num) float64 {
 	case int:
 		return float64(num)
 	case *big.Int:
-		if num.IsInt64() {
-			// Might fit in float64
+		if num.IsInt64() { // might fit in float64
+			// TODO: Make this more robust so the "might fit" is "will fit".
```
xiaq, line 201: There is no loss of robustness here. If the number overflows `float64`, the conversion below will result in +Inf or -Inf. I guess the comment isn't clear about that, I'll expand the comment.

pkg/eval/vals/num.go
```diff
@@ -201,13 +197,11 @@ func ConvertToFloat64(num Num) float64 {
...
 			return float64(num.Int64())
-		} else {
-			// Definitely won't fit in float64
-			return math.Inf(num.Sign())
 		}
+		return math.Inf(num.Sign()) // definitely won't fit in float64
```
xiaq, line 204: I personally prefer putting comments on their own lines.

xiaq (approved):

pkg/eval/external_cmd_unix_internal_test.go
krader1961: > I prefer keeping the dot import on testutil.

Okay, but at present the code is inconsistent with the vast majority of the imports not using the dot form. It's far from obvious why this file should use the dot form and the other test modules don't.

xiaq: It's probably because there are multiple nested `testutil.Dir` that I introduced the dot import. I don't have a rigorous rule I stick to though. I'll just say that I'm more tolerant of dot imports in test files when they make the code more readable; but the exact balance of is highly subjective and not entirely consistent.

I do avoid dot imports in non-test files as a rule. The only exception to that rule is allowing dot-importing `pkg/store/storedefs` from `pkg/store` (reflected in `pkg/store/staticcheck.conf`). In this case it's because the entire point of the `storedefs` package is to provide types used in the `store` package's API, but they have to be artificially split into different packages so that other packages that only need depend on those types can import `storedefs` instead of `store`.

PR #1368, opened 2021-07-14
Update run-parallel example

### LINKED ISSUE #1357: Capture stdout and stderr into different variables
It would be really nice if Elvish had this solved without using temp files, for example to catch `stdout` and `stderr` into separate vars

```sh
error_file=$(mktemp)
out=$(some_command 2>$error_file)
err=$(< $error_file)
rm $error_file
```

https://stackoverflow.com/questions/11027679/capture-stdout-and-stderr-into-different-variables

krader1961:
Courtesy of @zzamboni this updates the example of how to use the
`run-parallel` command to capture the stdout and stderr byte streams
independent of each other.

Resolves #1357

### DIFF
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -947,6 +947,10 @@ and byte output might not agree with the order in which they happened:
 ▶ a
 ```
 
+If you want to capture the stdout and stderr byte streams independent of each
+other see the example in the [run-parallel](./builtin.html#run-parallel)
+documentation.
+
 ## Exception capture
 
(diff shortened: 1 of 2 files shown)


website/ref/language.md
```diff
@@ -947,6 +947,10 @@ and byte output might not agree with the order in which they happened:
 ▶ a
 ```
 
+If you want to capture the stdout and stderr byte streams independent of each
```
xiaq, line 950: This needs a leading "note" since it's non-normative.

xiaq (approved):

PR #1379, opened 2021-08-09
epm: fix for xdg-dir

kolbycrouch:
This fixes the bundled `epm` module to respect the new XDG support.

### DIFF
--- a/pkg/eval/mods/bundled/epm.elv.go
+++ b/pkg/eval/mods/bundled/epm.elv.go
@@ -109,7 +109,14 @@ debug-mode = $false
 ]
 
 # Internal configuration
--data-dir = ~/.elvish
+-data-dir = (
+  if (not-eq $E:XDG_DATA_HOME '') {
+    put $E:XDG_DATA_HOME/elvish
+  }  else {
+    put ~/.local/share/elvish
+  }
+)
+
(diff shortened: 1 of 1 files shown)


pkg/eval/mods/bundled/epm.elv.go
```diff
@@ -109,7 +109,14 @@ debug-mode = $false
 ]
 
 # Internal configuration
--data-dir = ~/.elvish
+-data-dir = (
+  if (has-env XDG_CONFIG_HOME) {
+    put $E:XDG_CONFIG_HOME/elvish
```
xiaq, line 114: An empty XDG_CONFIG_HOME should be treated as if it is unset; the condition here should be

`(not (eq $XDG_CONFIG_HOME ''))`

xiaq: Ah, and the design is to use XDG_DATA_HOME, not XDG_CONFIG_HOME: https://github.com/elves/elvish/issues/383#issuecomment-869217648

kolbycrouch: Oops. Sorry about that. Fixed it.

xiaq:
This needs to be amended to work correctly on Windows too; I'll do that after merging this.

PR #1386, opened 2021-08-23
Bump gccgo to 11

zhsj:
gcc 11 implements go1.16

The default symlink(`/usr/bin/go -> /usr/bin/go-10`) provided by gccgo-go package hasn't been bumped in Debian unstable. It might take sometime since there are regressions in gcc-11, ref https://tracker.debian.org/pkg/gcc-11

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -10,12 +10,11 @@ test_task:
         image: debian:unstable-slim
       setup_script:
         - apt-get -y update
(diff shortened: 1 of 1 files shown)


krader1961:
FWIW, the change comment confused me. I think I understand the gist of the comment but the straightforward interpretation appears to be at odds with the actual change.

zhsj:
@krader1961  Which part confused you?

PR #1394, opened 2021-08-29
Replace remaining deprecated ioutil.ReadDir uses

### LINKED ISSUE #1350: Replace uses of Go package "io/ioutil" when the minimum supported Go is 1.16
I noticed that the Go "io/ioutil" package has been deprecated: https://golang.org/doc/go1.16#ioutil. Elvish makes extensive use of that package. When Go 1.17 is released we should update this project to use the alternatives documented by that link.

krader1961:
Resolves #1350

### DIFF
--- a/pkg/edit/complete/generators.go
+++ b/pkg/edit/complete/generators.go
@@ -2,7 +2,6 @@ package complete
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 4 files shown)


pkg/edit/complete/generators.go
```diff
@@ -108,14 +107,15 @@ func generateFileNames(seed string, onlyExecutable bool) ([]RawItem, error) {
 	// Make candidates out of elements that match the file component.
 	for _, info := range infos {
 		name := info.Name()
+		finfo, _ := info.Info()
```
xiaq, line 110: This can return an error; do not ignore it.

Having two variables named "info" and "finfo" is confusing. Better rename the original info variable as it is no longer a FileInfo value.

xiaq (approved):

pkg/edit/complete/generators.go
krader1961: Yes, it can return an error. But ignoring it in this context should be okay; although, the logic is subtle. Since the `FileInfo` object will be initialized to the zero value(s) the property tests will all evaluate to false causing it to be silently ignored. In hindsight I agree that an explicit `if err != nil { continue }` is clearer and thus preferable.

PR #1417, opened 2021-10-27
Add transpose char and transpose word

### LINKED ISSUE #726: Add `transpose` functions in insert mode
In emacs and bash, there is a "transpose" functionality that swaps two adjacent runes or words.  `Ctrl-t` swaps two runes adjacent to the cursor; `Alt-t` swaps two words adjacent to the cursor.  It would be nice to implement this in elvish too.

---

I would suggest, if #721 is merged, that the transpose function be implemented also with a pure, transformer-like behavior, e.g.

```go
func transposeRunes(buffer string, dot int) (string, int) {
  if dot == 0 || dot == len(buffer) {
    return buffer, dot
  }

  return buffer[:dot-1] + buffer[dot] + buffer[dot-1] + buffer[dot+1:], dot+1
}
```

So that the transformation is applied with a mutator function (which is analogous in feature to `applyKill`, `applyMove`, as suggested in #722), as suggested in my last comment on #722 as `modifyBuffer`:

```go
func (ed *editor) modifyBuffer(fn func(string, int) (string, int)) {
  ed.buffer, ed.dot = fn(ed.buffer, ed.dot)
}
```

If we like this suggestion, then we would also change the `applyMove` and `applyKill` functions to be pure transformer functions, i.e.

```go
func applyMove(fn func(string, int) int) {
  return ed.buffer, fn(ed.buffer, ed.dot)
}

func applyKill(moveFn func(string, int) int) {
  // do computation to convert move sequence to buffer deletion
  return newBuffer, newDot
}
```

---

Alternatively, we could implement the transpose functions directly on `*editor`, but I feel like that's a less-safe and less functional kind of approach, with the more functional `modifyBuffer` approach a bit cleaner and more general.  Just my two cents.

sauyon:
resolves #726

### DIFF
--- a/pkg/edit/builtins.go
+++ b/pkg/edit/builtins.go
@@ -274,6 +274,8 @@ var bufferBuiltinsData = map[string]func(*tk.CodeBuffer){
 	"kill-right-alnum-word": makeKill(moveDotRightAlnumWord),
 	"kill-line-left":        makeKill(moveDotSOL),
 	"kill-line-right":       makeKill(moveDotEOL),
+
+	"transpose-rune": makeTransform(transposeRunes),
 }
 
 func initBufferBuiltins(app cli.App, nb eval.NsBuilder) {
@@ -423,6 +425,47 @@ func moveDotDown(buffer string, dot int) int {
 	sol := strutil.FindLastSOL(buffer[:dot])
 	width := wcwidth.Of(buffer[sol:dot])
 	return nextSOL + len(wcwidth.Trim(buffer[nextSOL:nextEOL], width))
+}
+
+//elvdoc:fn transpose-rune
+//
+// Swaps the characters to the left and right of the dot. If the dot is
+// at the beginning of the buffer, swaps the first two characters, and
+// if it is at the end, it swaps the last two.
+
+func transposeRunes(buffer string, dot int) (string, int) {
+	if len(buffer) == 0 {
+		return buffer, dot
+	}
+
+	var res string
+	var newDot int
+	// transpose at the beginning of the buffer transposes the first two
+	// characters, and at the end the last two
+	if dot == 0 {
+		first, firstLen := utf8.DecodeRuneInString(buffer)
+		if firstLen == len(buffer) {
+			return buffer, dot
+		}
+		second, secondLen := utf8.DecodeRuneInString(buffer[firstLen:])
+		res = string(second) + string(first) + buffer[firstLen+secondLen:]
+		newDot = firstLen + secondLen
+	} else if dot == len(buffer) {
+		second, secondLen := utf8.DecodeLastRuneInString(buffer)
+		if secondLen == len(buffer) {
+			return buffer, dot
+		}
+		first, firstLen := utf8.DecodeLastRuneInString(buffer[:len(buffer)-secondLen])
+		res = buffer[:len(buffer)-firstLen-secondLen] + string(second) + string(first)
+		newDot = len(res)
+	} else {
+		first, firstLen := utf8.DecodeLastRuneInString(buffer[:dot])
+		second, secondLen := utf8.DecodeRuneInString(buffer[dot:])
+		res = buffer[:dot-firstLen] + string(second) + string(first) + buffer[dot+secondLen:]
+		newDot = dot + secondLen
+	}
+
+	return res, newDot
 }
 
 // TODO(xiaq): Document the concepts of words, small words and alnum words.
--- a/pkg/edit/builtins.go
+++ b/pkg/edit/builtins.go
@@ -324,6 +324,16 @@ func makeKill(m pureMover) func(*tk.CodeBuffer) {
 	}
 }
 
+// A pure function that takes the current buffer and dot, and returns a new
+// value for the buffer and dot.
+type transformer func(buffer string, dot int) (string, int)
+
+func makeTransform(t transformer) func(*tk.CodeBuffer) {
+	return func(buf *tk.CodeBuffer) {
+		buf.Content, buf.Dot = t(buf.Content, buf.Dot)
+	}
+}
+
 // Implementation of pure movers.
 
 //elvdoc:fn move-dot-left
--- a/pkg/edit/builtins.go
+++ b/pkg/edit/builtins.go
@@ -563,60 +563,93 @@ func categorizeAlnum(r rune) int {
 // different word categories.
 type categorizer func(rune) int
 
+// Skip only word characters left from pos in buffer and return that position.
+//
+// Note that a word character must be to the left of pos.
+func (categorize categorizer) skipWordLeft(buffer string, pos int) int {
+	if pos == 0 {
+		return pos
+	}
+
+	r, _ := utf8.DecodeLastRuneInString(buffer[:pos])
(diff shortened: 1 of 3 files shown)


pkg/edit/builtins.go
```diff
@@ -563,60 +641,134 @@ func categorizeAlnum(r rune) int {
...
+	// after skipping whitespace, find the end of the right word
+	pos := categorize.skipWsRight(buffer, dot)
+	var rightEnd int
+	if pos == len(buffer) {
+		// there is only whitespace to the right of the dot
+		rightEnd = categorize.skipWsLeft(buffer, pos)
+	} else {
+		rightEnd = categorize.skipWordRight(buffer, pos)
+	}
+	rightStart := categorize.skipWordLeft(buffer, rightEnd)
```
xiaq, line 750: IIUC this should be the same as `pos` if the `else` branch above was taken? If so, consider rewriting this as:

```go
var rightStart, rightEnd int
if pos == len(buffer) {
  ...
} else {
  rightEnd = ...
  rightLeft = pos
}
```

pkg/edit/builtins_test.go
```diff
@@ -208,6 +208,71 @@ var bufferBuiltinsTests = []struct {
...
+	},
+	{
+		"transpose-word",
+		tk.CodeBuffer{Content: " \t\n  ", Dot: 4},
+		tk.CodeBuffer{Content: " \t\n  ", Dot: 4},
+	},
+	{
+		"transpose-word",
+		tk.CodeBuffer{Content: "ab  bc cd", Dot: 9},
+		tk.CodeBuffer{Content: "ab  cd bc", Dot: 9},
```
xiaq, line 274: Please add at least one test case for `transpose-small-word` and `transpose-alnum-word`.

pkg/edit/builtins_test.go
```diff
@@ -208,6 +208,71 @@ var bufferBuiltinsTests = []struct {
 		tk.CodeBuffer{Content: "ab", Dot: 1},
 		tk.CodeBuffer{Content: "a", Dot: 1},
 	},
+	{
+		"transpose-char",
```
xiaq, line 212: Give these test cases unique, meaningful names. This one can be called `transpose-char with dot at beginning`

pkg/edit/builtins.go
```diff
@@ -563,60 +641,134 @@ func categorizeAlnum(r rune) int {
...
-func moveDotLeftGeneralWord(categorize categorizer, buffer string, dot int) int {
-	left := buffer[:dot]
-	skipCat := func(cat int) {
-		left = strings.TrimRightFunc(left, func(r rune) bool {
-			return categorize(r) == cat
-		})
+// Skip only word characters left from pos in buffer and return that position.
+//
+// Note that a word character must be to the left of pos.
+func (categorize categorizer) skipWordLeft(buffer string, pos int) int {
```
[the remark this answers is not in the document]
sauyon: Simply that it made the code in the functions look slightly cleaner. I'll make them functions.

pkg/edit/builtins.go
sauyon: No, if `dot` is in the middle of a word, then `pos` is `dot`, so we still need to skip word characters to the left to the find the left of the word. Will document this.

pkg/edit/builtins.go
```diff
@@ -274,6 +274,11 @@ var bufferBuiltinsData = map[string]func(*tk.CodeBuffer){
 	"kill-right-alnum-word": makeKill(moveDotRightAlnumWord),
 	"kill-line-left":        makeKill(moveDotSOL),
 	"kill-line-right":       makeKill(moveDotEOL),
+
+	"transpose-char":       makeTransform(transposeRunes),
```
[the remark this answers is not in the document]
sauyon: Not sure how I missed that like 10 lines above where I defined this...

sauyon:
I've applied review comments in a few fixup commits.

pkg/edit/builtins_test.go
```diff
@@ -208,6 +209,76 @@ var bufferBuiltinsTests = []struct {
...
+	},
+	{
+		"transpose-word with no words",
+		tk.CodeBuffer{Content: " \t\n  ", Dot: 4},
+		tk.CodeBuffer{Content: " \t\n  ", Dot: 4},
+	},
+	{
+		"transpose-word with complex input",
+		tk.CodeBuffer{Content: "cd ~/downloads;", Dot: 4},
+		tk.CodeBuffer{Content: "~/downloads;", Dot: 15},
```
xiaq, line 270: Fix this test - you should be able to see the test errors.

sauyon: Whoops, sorry.

pkg/edit/builtins.go
```diff
@@ -563,60 +649,137 @@ func categorizeAlnum(r rune) int {
...
-	left := buffer[:dot]
-	skipCat := func(cat int) {
-		left = strings.TrimRightFunc(left, func(r rune) bool {
-			return categorize(r) == cat
-		})
+// Skip only word characters left from pos in buffer and return that position.
+//
+// Note that a word character must be to the left of pos. If it is a
+// whitespace character, this will instead skip whitespace.
+func skipWordLeft(categorize categorizer, buffer string, pos int) int {
```
xiaq, line 656: It's better to just name the function `skipSameCatLeft`.

pkg/edit/builtins.go
```diff
@@ -563,60 +649,137 @@ func categorizeAlnum(r rune) int {
...
+	return len(left)
+}
 
-	// get category of last rune
-	r, _ := utf8.DecodeLastRuneInString(left)
+// Skip only word characters right from pos in buffer and return that position.
+//
+// Note that a word character must be to the right of pos. If it is a
+// whitespace character, this will instead skip whitespace.
+func skipWordRight(categorize categorizer, buffer string, pos int) int {
```
xiaq, line 683: Better to just name this `skipSameCatRight`.

pkg/edit/builtins.go
```diff
@@ -563,60 +649,137 @@ func categorizeAlnum(r rune) int {
...
+	var rightEnd int
+	if pos == len(buffer) {
+		// there is only whitespace to the right of the dot
+		rightEnd = skipWsLeft(categorize, buffer, pos)
+	} else {
+		rightEnd = skipWordRight(categorize, buffer, pos)
+	}
+	// if the dot started in the middle of a word, 'pos' is the same as dot,
+	// so we should skip word characters to the left to find the start of the
+	// word
```
xiaq, line 760: I see you added this as a response to my question, thanks. The comment feels a bit out of place because it's responding to a particular doubt about why this is not equal to `pos`, which the prospective reader does not necessarily have. I think it's probably better just to remove this comment, the code is actually clear enough.

xiaq (approved):
Some minor comments that I'll act on after merging the PR.

Some more overall style comments:

- `builtins.go` is getting rather large. I'll factor out the buffer builtins into a separate `buffer_builtins.go`.

- Elvish source files, for the most part, are written top-down, so helper functions tend to appear later in the file. I'll be reordering some functions.

PR #1419, opened 2021-11-01
Add `use` unit tests and tweak documentation

krader1961:
This explicitly tests a common error case not currently verified by the
existing unit tests. Thus improving test coverage by one line.  :-)

### DIFF
--- a/pkg/eval/builtin_special.go
+++ b/pkg/eval/builtin_special.go
@@ -36,9 +36,11 @@ var builtinSpecials map[string]compileBuiltin
 // intended for external consumption, e.g. the syntax highlighter.
 var IsBuiltinSpecial = map[string]bool{}
 
-type noSuchModule struct{ spec string }
+// NoSuchModule encodes an error where a module spec cannot be resolved.
+type NoSuchModule struct{ spec string }
 
-func (err noSuchModule) Error() string { return "no such module: " + err.spec }
+// Error implements the error interface.
+func (err NoSuchModule) Error() string { return "no such module: " + err.spec }
 
 func init() {
 	// Needed to avoid initialization loop
@@ -316,7 +318,12 @@ func (op useOp) exec(fm *Frame) Exception {
 	return nil
 }
 
+// TODO: Add support for module specs relative to a package/workspace.
+// See https://github.com/elves/elvish/issues/1421.
 func use(fm *Frame, spec string, r diag.Ranger) (*Ns, error) {
+	// Handle relative imports. Note that this deliberately does not support Windows backslash as a
+	// path separator because module specs are meant to be platform independent. If necessary, we
+	// translate a module spec to an appropriate path for the platform.
 	if strings.HasPrefix(spec, "./") || strings.HasPrefix(spec, "../") {
 		var dir string
 		if fm.srcMeta.IsFile {
@@ -331,6 +338,8 @@ func use(fm *Frame, spec string, r diag.Ranger) (*Ns, error) {
 		path := filepath.Clean(dir + "/" + spec)
 		return useFromFile(fm, spec, path, r)
 	}
+
+	// Handle imports of pre-defined modules like `builtin` and `str`.
 	if ns, ok := fm.Evaler.modules[spec]; ok {
 		return ns, nil
 	}
@@ -338,16 +347,21 @@ func use(fm *Frame, spec string, r diag.Ranger) (*Ns, error) {
 		return evalModule(fm, spec,
 			parse.Source{Name: "[bundled " + spec + "]", Code: code}, r)
 	}
+
+	// Handle imports relative to the Elvish module search directories.
+	//
 	// TODO: For non-relative imports, use the spec (instead of the full path)
 	// as the module key instead to avoid searching every time.
 	for _, dir := range fm.Evaler.LibDirs {
 		ns, err := useFromFile(fm, spec, filepath.Join(dir, spec), r)
-		if _, isNoSuchModule := err.(noSuchModule); isNoSuchModule {
+		if _, isNoSuchModule := err.(NoSuchModule); isNoSuchModule {
 			continue
 		}
 		return ns, err
 	}
-	return nil, noSuchModule{spec}
+
+	// Sadly, we couldn't resolve the module spec.
+	return nil, NoSuchModule{spec}
 }
 
 // TODO: Make access to fm.Evaler.modules concurrency-safe.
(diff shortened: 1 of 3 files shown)


website/ref/language.md
```diff
@@ -2392,18 +2393,27 @@ itself or defined by the user.
 
 ### Importing modules with `use`
 
-Modules are imported using the `use` special command, which accepts a **module
-spec** and an optional alias:
+Modules are imported using the `use` special command. It requires a module spec
+and an optional alias:
```
xiaq, line 2397: Not sure about this change. "Requires ... an optional..." is an oxymoron; "accept" is more appropriate. Also not sure why you'd removed the emphasis on `module spec`; it is the first time this term appears in the documentation and should be emphasized.

website/ref/language.md
```diff
@@ -2392,18 +2393,27 @@ itself or defined by the user.
...
 
 ```elvish
 use $spec $alias?
 ```
 
-Both the module spec and the alias must appear as a single string literal.
+The module spec and the alias must both be a [string literal](#string).
```
xiaq, line 2403: Not sure why you removed "single" - it is needed to exclude cases where the spec of the alias appear as a compound of multiple string literals, which is not valid:

```
use 'a'/b/c
```

website/ref/language.md
```diff
@@ -2392,18 +2393,27 @@ itself or defined by the user.
...
 
-The module spec specifies which module to import, and the alias, if given,
-specifies the namespace to import the module under. By default, the namespace is
-derived from the module spec, by taking the part after the last slash.
+The module spec specifies which module to import. The alias, if given, specifies
+the namespace to import the module under. By default, the namespace is derived
+from the module spec by taking the part after the last slash.
+
+Module specs fall into two categories:
```
xiaq, line 2409: This description leaves out the important case of modules in module search directories.

pkg/eval/builtin_special.go
```diff
@@ -36,9 +36,9 @@ var builtinSpecials map[string]compileBuiltin
 // intended for external consumption, e.g. the syntax highlighter.
 var IsBuiltinSpecial = map[string]bool{}
 
-type noSuchModule struct{ spec string }
+type NoSuchModule struct{ spec string }
```
xiaq, line 39: This needs a godoc now that it is exported.

pkg/eval/builtin_special.go
```diff
@@ -36,9 +36,9 @@ var builtinSpecials map[string]compileBuiltin
...
 
-func (err noSuchModule) Error() string { return "no such module: " + err.spec }
+func (err NoSuchModule) Error() string { return "no such module: " + err.spec }
```
xiaq, line 41: This needs a godoc now that it is exported.

pkg/eval/builtin_special.go
```diff
@@ -316,7 +316,12 @@ func (op useOp) exec(fm *Frame) Exception {
 	return nil
 }
 
+// TODO: Add support for module specs relative to a package/workspace.
+// See https://github.com/elves/elvish/issues/1400.
```
xiaq, line 320: This references a closed issue, which is confusing. I filed https://b.elv.sh/1421 that can be referenced here instead.

pkg/eval/builtin_special.go
```diff
@@ -331,23 +336,30 @@ func use(fm *Frame, spec string, r diag.Ranger) (*Ns, error) {
...
+	// Handle imports of pre-defined modules like `builtin` and `str`.
 	if ns, ok := fm.Evaler.modules[spec]; ok {
 		return ns, nil
 	}
 	if code, ok := fm.Evaler.BundledModules[spec]; ok {
 		return evalModule(fm, spec,
 			parse.Source{Name: "[bundled " + spec + "]", Code: code}, r)
 	}
+
+	// Handle imports relative to the Elvish config lib directories.
```
xiaq, line 349: The "config" here is superfluous since config directory and lib directories are now separate concepts.

pkg/eval/builtin_special.go
```diff
@@ -331,23 +336,30 @@ func use(fm *Frame, spec string, r diag.Ranger) (*Ns, error) {
...
 		ns, err := useFromFile(fm, spec, filepath.Join(dir, spec), r)
-		if _, isNoSuchModule := err.(noSuchModule); isNoSuchModule {
+		if _, isNoSuchModule := err.(NoSuchModule); isNoSuchModule {
 			continue
 		}
 		return ns, err
 	}
-	return nil, noSuchModule{spec}
+
+	// Sadly, we couldn't resolve the module spec.
```
xiaq, line 361: This comment doesn't say anything that's not clear from the code.

Some codebases use a pseudo-literate style where the comments form a complete description of the behavior of the code, but Elvish's codebase does not use that style and I don't intend to adopt that either.

website/ref/language.md
krader1961: I would argue that your example is a "string literal".; albeit not a "single" string literal. And it is not obvious to me why a "single" (which I interpret as "simple") string literal is relevant in this context. Also, neither your, or my, definition (with or without the "single" prefix) appears anywhere in the documentation AFAICT. Why can't a "single string literal" in this context be composed of a composition of string literals? This definition, obviously,

pkg/eval/builtin_special.go
krader1961: Fair enough. Removed. I added that comment to be explicit about the logic but agree it isn't needed for any competent contibutor.

website/ref/language.md
krader1961: I settled on this verbiage: "It requires a module spec and allows a namespace alias:". I removed the bolding because it doesn't tell the user anything meaningful but I reinstated it since I, sort of, see your point.

website/ref/language.md
krader1961: I opted for something a bit more verbose with a link to the "Compounding" section.

website/ref/language.md
krader1961: D'oh! Of course. I'll add an item for that case.

pkg/eval/builtin_special.go
krader1961: done

pkg/eval/builtin_special.go
krader1961: done

pkg/eval/builtin_special.go
krader1961: done

pkg/eval/builtin_special.go
krader1961: I decided "module search directories." is clearer and consistent with the user-facing documentation.

krader1961:
PTAL

xiaq:
Thanks!

website/ref/language.md
xiaq: Thanks, the wording you chose is the indeed clearer.

PR #1428, opened 2021-11-15
Replace AnyError catchall with the specific error expected by a test

krader1961:


krader1961:
Gah! I didn't notice this depends on a different PR I already sent for review and that needs updating. Closing until that older PR is merged.

PR #1429, opened 2021-11-16
Added leading 0 format to num documentation

### LINKED ISSUE #1372: Octal Format Specifier
From the documentation:

> Integers are written in decimal (e.g. `10`), hexadecimal (e.g. `0xA`), octal (e.g. `0o12`) or binary (e.g. `0b1010`).

However, from my testing, `(num '010')` -> `(num 8)`

It is clear that a leading zero maps to octal, and that the `o` is optional. I think that this should not be the default behavior. As a language that is meant to be used for reading and parsing human-readable text, I think that leading zeros due to simple formatting is far more likely than it being an octal value.

However, I understand that a simple `0` prefix alone is a widely used way of formatting octal values. If the behavior is not changed, at the very least I think the documentation should be updated to reflect the optional nature of the `o`.

fennewald:
Closes #1372 

Saw I still had this issue open. Just added a line to the docs to clear it up.

### DIFF
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -220,7 +220,7 @@ can be constructed by passing their **string representation** to the
 [`num`](builtin.html#num) builtin command:
 
 -   **Integers** are written in decimal (e.g. `10`), hexadecimal (e.g. `0xA`),
-    octal (e.g. `0o12`) or binary (e.g. `0b1010`).
+    octal (e.g. `012`, or `0o12`) or binary (e.g. `0b1010`).
 
 -   **Rationals** are written as two exact integers joined by `/`, e.g. `1/2` or
     `0x10/100` (16/100).


website/ref/language.md
```diff
@@ -220,7 +220,7 @@ can be constructed by passing their **string representation** to the
 [`num`](builtin.html#num) builtin command:
 
 -   **Integers** are written in decimal (e.g. `10`), hexadecimal (e.g. `0xA`),
-    octal (e.g. `0o12`) or binary (e.g. `0b1010`).
+    octal (e.g. `012`, or `0o12`) or binary (e.g. `0b1010`).
```
krader1961, line 223: Fine by me although I would have swapped the two octal examples since, in some sense, the `0o` prefix is preferred and the `0` prefix exists for historical reasons. /pedantic /only-semi-serious 😄

krader1961: On a more serious note please change "Digits may be separately by underscores" to "Digits may be separated by underscores" a few lines farther down as part of this change. That is, change "separately" to "separated". That sentence needs to be rewritten but that is outside the scope of your change. Simply fixing "separately" to "separated" is sufficient and justifies an update to the P.R. that also changes the order of the octal examples per my previous comment. :smile:

krader1961: I'll also point out the examples should include at least one silly, yet legal, example  such as `1_2_3` being equivalent to `123`. Whether that improvement to the doc should be part of this change is debatable.

xiaq:
This PR is fine but I'd prefer to note that the `010` syntax may go away, which I did in 6ea9753769e2c7c85eb56e3bb5ef5016af8703de. Thanks for the PR but I'll close it now.

PR #1434, opened 2021-11-22
Improve the "wrong type" error message

### LINKED ISSUE #715: Wrong error message "need nil, got nil"
```
~> ns = (ns [])
Exception: wrong type of 1'th argument: need !!<nil>, got list
```

This should say "need map, got list"

### LINKED ISSUE #744: Confusing error message for `x~ = foo`
Reported by @zzamboni:

```
> foo~ = hi
Exception: multiple errors: need !!<nil>, got string; need !!<nil>, got string
[tty], line 1: foo~ = hi
```

krader1961:
Output correct error messages caused by statements such as these:

    var x~ = $true
    ns $true
    each abc

Fixes #715
Fixes #744

### DIFF
--- a/pkg/edit/completion.go
+++ b/pkg/edit/completion.go
@@ -283,7 +283,8 @@ func (c complexItem) IterateKeys(f func(interface{}) bool) {
 	vals.Feed(f, "stem", "code-suffix", "display")
 }
 
-func (c complexItem) Kind() string { return "map" }
+// Kind is used by vals.Kind() to cause it to return the correct "kind" for these objects.
+func (complexItem) Kind() string { return "map" }
 
 func (c complexItem) Equal(a interface{}) bool {
 	rhs, ok := a.(complexItem)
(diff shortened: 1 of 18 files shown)


xiaq:
This is a rather intrusive and convoluted change to handle the error reporting when the `ptr` argument to `ScanToGo` is a pointer to an interface. Adding `Kind() string` to `Callable` doesn't do anything here, and the comment in the `default` branch of `Kind` is quite misguided since it contradicts with how interfaces actually work - you may want to read Russ Cox's writeup on the internals of Go's interface at  https://research.swtch.com/interfaces. The parts relevant to Go's formal semantics are actually described in the [spec itself](https://go.dev/ref/spec), but the details about dynamic type and conversion to interfaces are a bit scattered and Russ's introduction is much clearer even if you don't care about the machine representation of interfaces.

I've fixed the "need nil" message in ab88de7c15f842356dd8910ca593665aa29b4e91. Yes it'd be better if the error message says "need map" and "need fn" instead of "need !!hashmap.Map" or "need !!eval.Callable", but since the "kind" system is a stopgap solution before there's a real type system, I'm going to keep it more or less in the current state and avoid complicating it.

PR #1435, opened 2021-11-22
Add a builtin:compare command

### LINKED ISSUE #1347: We need a "compare" command independent of "order"
I have been working on an Elvish way to compare values in order to implement [semantic version](https://semver.org) comparisons. The [default comparison function](https://github.com/elves/elvish/blob/9981b2a777961f3669200e9b890c16cbb6935d3d/pkg/eval/builtin_fn_container.go#L899) used by the `order` command appears to do what I want. It is a shame that I have to reinvent that wheel using Elvish commands.

krader1961:
Expose the default comparison function used by the `builtin:order`
command as a command in its own right. This command is useful when
writing, in Elvish, something like `builtin:order`. Such as a semantic
version comparison command.

Resolves #1347


0.17.0-release-notes.md
```diff
@@ -66,3 +66,7 @@ New features in the standard library:
 
 -   The `store:` module now exposes all functionalities of Elvish's persistent
     store.
+
+-   New `compare` command to allow efficient comparison of numbers, strings, and
+    lists ([#1347](https://b.elv.sh/1347)). This is the default comparator used
+    by `order`.
```
xiaq, line 72: This is inaccurate - `order` does not use a comparator, it uses a less-than callback. Better to say that this is consistent with the criteria used by `order`.

pkg/eval/builtin_fn_container.go
```diff
@@ -896,7 +899,58 @@ const (
...
+//
+// ```elvish-transcript
+// ~> compare a b
+// ▶ 1
+// ~> compare b a
+// ▶ -1
+// ~> compare x x
+// ▶ 0
+// ~> order (float64 10) (float64 1)
+// ▶ 1
```
xiaq, line 933: These outputs seem wrong since `order` outputs int values.

pkg/eval/builtin_fn_container_test.go
```diff
@@ -203,17 +235,20 @@ func TestOrder(t *testing.T) {
 		That("put 10 1 5 2 | order").Puts("1", "10", "2", "5"),
 
 		// Ordering numbers
-		// Only small integers
+		// Only small integers normalized to the "num" type.
```
xiaq, line 238: "Normalization" has a specific meaning when it comes to number types. To clarify that these are values of number types instead of strings, say "typed numbers".

pkg/eval/builtin_fn_container.go
```diff
@@ -896,7 +899,58 @@ const (
...
-func compare(a, b interface{}) ordering {
+//elvdoc:fn compare
+//
+// ```elvish
+// compare $a $b
+// ```
+//
+// Outputs -1 if `$a` is less than `$b`, 0 if equal, and 1 if greater than.
+//
+// The following comparison algorithm is used:
```
xiaq, line 910: This algorithm is now documented in both `order` and `compare`; better to avoid the duplication.

xiaq (approved):
Some issues, but I'll merge and push a fixup commit later.

0.17.0-release-notes.md
xiaq: I am also unsure what the "efficient" part refers to so I'll just remove it. If there's a specific reason to keep it I can add it back, probably with a different wording.

PR #1439, opened 2021-11-26
Augment `builtin:range` to support counting down

### LINKED ISSUE #1436: Augment `range` to support counting down
While pondering augmenting `range` to accept a lambda arg (issue #1433) it occurred to me there is no good reason the command has to be restricted to counting up. In other words, `range 5 0` should be equivalent to `range &step=-1 5 0` to produce the sequence 5, 4, 3, 2, 1. Obviously the argument (and `&step` option) sanity checks would need to be adjusted. Too, the documentation should replace the terms `low?` and `high` with something like `start?` and `end`.

TBD is whether the `end` value should be excluded or included when counting down. The former would be consistent with the current counting up behavior. While the latter feels more natural and is consistent with the `seq` command (at least on macOS).

krader1961:
Resolves #1436

### DIFF
--- a/pkg/eval/builtin_fn_container.go
+++ b/pkg/eval/builtin_fn_container.go
@@ -137,27 +137,52 @@ func makeMap(input Inputs) (vals.Map, error) {
 //elvdoc:fn range
 //
 // ```elvish
-// range &step=1 $low? $high
+// range &step=0 $start? $end
 // ```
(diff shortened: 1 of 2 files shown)


xiaq (approved):
Some issues, but I'll merge and do a fixup afterwards.

Issue #664, opened 2018-04-19
Closure syntax requires arbitrary lookahead


thomcc:
You need unbounded lookahead to distinguish between function parameters and a list. This makes syntax highlighting and other parsing or analysis much more difficult.

Specifically, when you encounter `[` you don't know if the items inside are function parameters or members of a list until you hit the `]{`. This is generally something you want to avoid, especially in cases where you're defining symbols into a scope, or doing other things that tooling might want to understand.

As a result, syntax highlighting this correctly for elvish isn't really possible using common methods (tmLanguage, sublime-syntax, etc) without fragile hacks. It also means that it's harder to report errors if a user puts a space between `]` and `{` (you have no way of knowing if it was intentional or not).

Many parser generators also have trouble with cases like these, as well (although I'm not sure when someone would be implementing a parser generator for the language they use with their shell).

The easiest fix would probably be adding a token that comes before the parameter list. `fn[a b c]{ ... }` would be somewhat consistent with function declaration. A sigil already used by the language like `$`, `@`, `&` in the place of `fn` would also work as well.

There are a lot of other choices as well, and I have no preference for which you end up going with (so long as it doesn't also have the same issue).

That said, this would require breaking a large amount of existing elvish code, so I'd understand being reluctant about changing it. A tool to fix code like that automatically would not be hard to implement though.

zzamboni:
@xiaq commented about this a few days ago in the Telegram group. Search for "unapologetically haskellesque" in the group to find the discussion :)

thomcc:
> hence the working syntax-highlighting support already existing for multiple editors.

Do they work if a new line is in the middle of the argument list? You can do some hacks where match the whole decl as a regex, and pull it apart, but at least in editors that use tmLanguage (sublime text also has sublime-syntax, but that has this restriction as well), those regexps cannot go over multiple lines (there are other issues with doing this as well, but this is the biggest one).

Maybe that's enough of an edge case where it doesn't matter though, although not being able to offer good help for someone who accidentally separates the `]` and `{` with a space like `] {` also is an issue caused by this, which might matter more to some.

zzamboni:
I agree about the thing with `] {`, and it has bitten me more than once. IIRC, that was one of the points mentioned by @xiaq in the discussion about maybe changing the syntax for closure declarations. 

The Elvish mode in Emacs uses regex-based syntax highlighting, so many cases are not properly handled - e.g. the new-line-in-the-middle (not the color change):
[...]

PR #1440, opened 2021-11-28
Replace legacy lambda with new lambda examples

krader1961:
I was surprised to see so many legacy lambda syntax examples in the
documentation. This replaces all of them with the new syntax -- excluding
the handful of cases meant to explicitly verify the legacy form is still
valid. This also adds a link to the issue in the release notes which
documents the change in syntax.

Related #664

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -477,7 +477,7 @@ type Primary struct {
 	node
 	ExprCtx ExprCtx
 	Type    PrimaryType
-	// Legacy lambda uses [args]{ body } instead of { |args| body }
+	// Legacy lambda uses {|args| body } instead of { |args| body }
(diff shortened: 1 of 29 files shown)


pkg/parse/parse.go
```diff
@@ -477,7 +477,7 @@ type Primary struct {
 	node
 	ExprCtx ExprCtx
 	Type    PrimaryType
-	// Legacy lambda uses [args]{ body } instead of { |args| body }
+	// Legacy lambda uses {|args| body } instead of { |args| body }
```
xiaq, line 480: Revert

xiaq (approved):
Seems that some places now have 2 or 3 spaces after `{|...|`; I'll change them to 1 space.

PR #1442, opened 2021-12-06
Replace `AnyError` in tests with a specific error

krader1961:
The `AnyError` placeholder error can cause tests to succeed for errors
other than what was expected. That is, the use of `AnyError` can mask
bugs in a unit test. So replace it with the specific error, or error type,
the test expects to be raised.

This does not remove the anyError structure because it is used in
the TestCase.DoesNotCompile() method. To keep the size of this change
as small as possible I want to defer updating that use to a separate
change. However, remove the public AnyError var so future test writers
don't attempt to use it.


krader1961:
Hold on looking at this P.R. I just realized there were new uses of `AnyError` introduced since I wrote this change several months ago (was waiting for my rejected fix for the "wrong type" bug to be merged). Should have an update soon.

PR #1450, opened 2021-12-16
Use vals.NumType constants rather than zero

krader1961:
While working on my implementation of unix:ulimit I realized I needed
to use vals.UnifyNums(). Looking at the existing uses I was dismayed to
find several instances that pass a literal zero (a magic value) rather
than the symbolic vals.Int constant. This replaces those literal zeros
with Vals.Int and modifies the implementation to make using a literal
int where a const symbol is expected more difficult.

### DIFF
--- a/pkg/eval/builtin_fn_container.go
+++ b/pkg/eval/builtin_fn_container.go
@@ -1119,7 +1119,7 @@ func cmp(a, b interface{}) ordering {
 	case int, *big.Int, *big.Rat, float64:
 		switch b.(type) {
 		case int, *big.Int, *big.Rat, float64:
-			a, b := vals.UnifyNums2(a, b, 0)
+			a, b := vals.UnifyNums2(a, b, vals.Int)
 			switch a := a.(type) {
 			case int:
 				return compareInt(a, b.(int))
--- a/pkg/eval/builtin_fn_num.go
+++ b/pkg/eval/builtin_fn_num.go
@@ -248,7 +248,7 @@ func chainCompare(nums []vals.Num,
 
 	for i := 0; i < len(nums)-1; i++ {
 		var r bool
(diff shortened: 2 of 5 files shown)


xiaq:
The use of 0 is intentional to signify that no minimal NumType is needed.

`Int` happens to be the lowest in the order of implicit conversion now, but suppose another type is introduced and takes the value of 0. In that case the places that don't need a minimal NumType should continue to use 0 instead of `Int`.

krader1961:
> The use of 0 is intentional to signify that no minimal NumType is needed.
> 
> `Int` happens to be the lowest in the order of implicit conversion now, but suppose another type is introduced and takes the value of 0. In that case the places that don't need a minimal NumType should continue to use 0 instead of `Int`.

Okay, but magic values like zero are still problematic in this type of context where a reader would expect a   constant identifier. It seems to me to be preferable to introduce a distinct symbol to be used in those contexts where the caller doesn't care what minimal type is returned. Perhaps `vals.AnyType` for the zero value. The problem is that a magic integer literal, rather than an identifier, makes understanding the code harder. For example, someone using an IDE or editor that supports the Go LSP (language specific processor) protocol can't simply double-click on the function call argument to understand the meaning of the argument. You certainly wouldn't allow passing a literal one (`1`) where the caller meant `vals.BigInt`. I don't see how a placeholder for the minimal type (currently `Int`) is any different.

Also, while orthogonal to my above argument the only numeric type I'm familiar with more minimal than the set of integers is the set of natural numbers (i.e., unsigned integers). Which we might want to support explicitly since it has obvious applications. Such as the `unix:ulimit` command I'm working on. 😄

xiaq:
046e02869d935fdf6c4df23020f2cf97ce6f97f4 documented the use of 0 as the argument to `UnifyNums`.

PR #1451, opened 2021-12-16
Use ENABLE_WRAP_AT_EOL_OUTPUT to show output fully on Windows Terminal on Windows.

moko256:
On master, the command output longer than terminal width is not shown fully.
It does not occur on WSL Ubuntu on Windows Terminal.
This PR will fix it by using ENABLE_WRAP_AT_EOL_OUTPUT ([MSDN SetConsoleMode function](https://docs.microsoft.com/en-us/windows/console/setconsolemode)).

Steps to reproduce:
- Set up elvish with Windows Terminal.
- Do `echo` with the string longer than terminal width.

Behavior on Console Windows Host:
- The string is wrapped and shown all.

Behavior on Windows Terminal:
- The string overflowed is not shown.


moko256:
Sorry, I missed the comments in `pkg/cli/term/setup_unix.go`. I'll rewrite this.

PR #1452, opened 2021-12-20
More improvements to `edit:add-var` documentation

krader1961:
The existing `edit:add-var` documentation incorrectly implies the `$init`
value is optional. It also doesn't include any examples that are useful
in practice. This addresses both shortcomings.

### DIFF
--- a/pkg/edit/vars.go
+++ b/pkg/edit/vars.go
@@ -22,10 +22,10 @@ func initVarsAPI(ed *Editor, nb eval.NsBuilder) {
 // edit:add-var $name $init
 // ```
 //
-// Declares a new variable in the REPL with an initial value. The new variable
-// becomes available during the next REPL cycle.
+// Defines a new variable in the interactive REPL with an initial value. The new variable becomes
+// available during the next REPL cycle.
 //
-// Equivalent to running `var $name = $init` at the REPL, but `$name` can be
+// Equivalent to running `var $name = $init` at a REPL prompt, but `$name` can be
 // dynamic.
 //
 // This is most useful for modules to modify the REPL namespace. Example:
@@ -41,9 +41,8 @@ func initVarsAPI(ed *Editor, nb eval.NsBuilder) {
 // ▶ (num 2)
 // ```
 //
-// Note that if you pass a variable as the `$init` argument, `edit:add-var`
-// doesn't add the variable "itself" to the REPL namespace, it creates a new
-// variable, only using the existing variable's value:
+// Note that `edit:add-var` doesn't add (or alias) the `$init` variable "itself" to the REPL
+// namespace, It creates a new REPL variable using the existing value of the variable:
 //
 // ```elvish-transcript
 // ~> cat .config/elvish/lib/b.elv
@@ -55,6 +54,34 @@ func initVarsAPI(ed *Editor, nb eval.NsBuilder) {
 // ~> set foo = bar
 // ~> echo $b:foo
 // foo
+// ```
+//
+// Because of the above behavior this mechanism is perhaps most useful for adding a module namespace
+// or function to the REPL namespace. For example, you might include this in your
+// `~/.config/elvish/rc.elv` in order to make your Elvish interactive config work on both Windows
+// and UNIX-like systems:
+//
+// ```elvish-transcript
+// use platform
+// if $platform:is-unix {
+//   use unix
+//   edit:add-var unix: $unix:
+// }
+//
+// Similarly you might have functions, such as `$util:ff~` to "find files", that you want to be able
+// to use in the REPL without needing the module prefix (i.e., `ff` rather than `util:ff`) to use
+// the functions interactively. You might add a statement like this to your
+// `~/.config/elvish/rc.elv`:
+//
+// ```elvish-transcript
+// edit:add-var ff~ $util:ff~
+// ```
+//
+// Alternatively, in the `~/.config/elvish/lib/util.elv` module:
+//
+// ```elvish-transcript
+// fn ff { ... }
+// edit:add-var ff~ $ff~
 // ```
 
 func addVar(fm *eval.Frame, name string, val interface{}) error {


xiaq:
> The existing `edit:add-var` documentation incorrectly implies the `$init`
> value is optional.

Where is that implied and how does this PR fix that?

pkg/edit/vars.go
```diff
@@ -56,6 +55,34 @@ func initVarsAPI(ed *Editor, nb eval.NsBuilder) {
...
+// and UNIX-like systems:
+//
+// ```elvish-transcript
+// use platform
+// if $platform:is-unix {
+//   use unix
+//   edit:add-var unix: $unix:
+// }
+//
+// Similarly you might have functions, such as `$util:ff~` to "find files", that you want to be able
```
xiaq, line 71: The "find files" bit in the example doesn't feel very useful. Either an abstract function (just say `f`) or a more common example (like `ll` for `ls -l`) would be more useful.

xiaq (approved):

pkg/edit/vars.go
```diff
@@ -41,9 +41,8 @@ func initVarsAPI(ed *Editor, nb eval.NsBuilder) {
 // ▶ (num 2)
 // ```
 //
-// Note that if you pass a variable as the `$init` argument, `edit:add-var`
-// doesn't add the variable "itself" to the REPL namespace, it creates a new
-// variable, only using the existing variable's value:
+// Note that `edit:add-var` doesn't add (or alias) the `$init` variable "itself" to the REPL
```
xiaq, line 44: I'm not sure why you removed the "if you pass variable ..." wording. This note is really only relevant when passing a variable to `$init`. "Aliasing" is not a concern when you do something like `edit:add-var foo foo`. You can't alias a value since values themselves are immutable.

krader1961: The implication of the original statement is that the `$init` argument is optional which isn't true. The problem here is the existing command notation doesn't readily lend itself, to a casual reader, to distinguishing between a value and a reference to a variable. In fact, you can't pass a "variable" as the initial value to this command -- you can only pass a value (possibly derived from an existing variable). That is the source of my (and no doubt other's) misunderstanding despite the otherwise clear syntax and what caused me to mistakenly think `edit:add-var umask $unix:umask` would work.

pkg/edit/vars.go
krader1961: Fair enough, but I use an abbreviation for `ll` to map to `ls -lhtr` so that isn't a good example in my case. I also don't think a more abstract is a good idea. IMHO, it's usually better to provide a concrete example the user might relate to in this situation.

pkg/edit/vars.go
```diff
@@ -56,6 +55,34 @@ func initVarsAPI(ed *Editor, nb eval.NsBuilder) {
 // ~> echo $b:foo
 // foo
 // ```
+//
+// Because of the above behavior this mechanism is perhaps most useful for adding a module namespace
+// or function to the REPL namespace. For example, you might include this in your
+// `~/.config/elvish/rc.elv` in order to make your Elvish interactive config work on both Windows
```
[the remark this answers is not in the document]
krader1961: Yes, which gets back to a question I asked some weeks ago about how to represent the `rc.elv` file in the documentation. It seemed to me you had tentatively decided to use the UNIX path for the time being. See the output of this pipeline (which uses my find-files function):

```
grep 'rc\.elv' (ff) | grep -v '"src.elv.sh'
```

pkg/edit/vars.go
```diff
@@ -56,6 +55,34 @@ func initVarsAPI(ed *Editor, nb eval.NsBuilder) {
 // ~> echo $b:foo
 // foo
 // ```
+//
+// Because of the above behavior this mechanism is perhaps most useful for adding a module namespace
```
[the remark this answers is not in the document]
krader1961: Good point. I would argue it's prescriptive since I don't believe anyone has identified a use case for creating a non-namespace, non-function, variable in the REPL using this mechanism. Hence my earlier comments on IRC that perhaps this command should only allow such objects by default; possibly providing an option (tongue-in-cheek: `&i-know-what-i-am-doing`) to define a REPL var with some other value (scalar, list, map).

pkg/edit/vars.go
```diff
@@ -56,6 +55,34 @@ func initVarsAPI(ed *Editor, nb eval.NsBuilder) {
...
+// }
+//
+// Similarly you might have functions, such as `$util:ff~` to "find files", that you want to be able
+// to use in the REPL without needing the module prefix (i.e., `ff` rather than `util:ff`) to use
+// the functions interactively. You might add a statement like this to your
+// `~/.config/elvish/rc.elv`:
+//
+// ```elvish-transcript
+// edit:add-var ff~ $util:ff~
+// ```
```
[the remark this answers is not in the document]
krader1961: Not entirely since the `edit:add-var` is needed if the statement appears in a narrower scope; e.g., an `if` block. Which is the whole point of this command. Also, and more importantly, a hypothetical *lib/util.elv* module which had an `edit:add-var ff~ $ff~` statement makes it impossible to `use util` in a non-interactive context since doing so results in this exception: `Exception: exec: "edit:add-var": executable file not found in $PATH`.

pkg/edit/vars.go
xiaq: > The implication of the original statement is that the `$init` argument is optional which isn't true.

I'm still not quite sure how that misunderstanding arises. Is it because the wording "if you pass" can lead the reader to think that they don't have to pass anything at all as `$init`?

krader1961: > Is it because the wording "if you pass" can lead the reader to think that they don't have to pass anything at all as $init? 

Yes, and where I wrote "a value" in my previous comment I meant a literal string, number, etc. as opposed to a variable deref.

PR #1454, opened 2021-12-23
Cleanups to rc.elv references

krader1961:


website/ttyshot/tour/history-walk.html
```diff
@@ -1,2 +1,2 @@
-~&gt; <span class="sgr-4 sgr-32">vim</span><span class="sgr-4"> .elvish/rc.elv </span>                       <span class="sgr-7">xiaq@macarch</span>
+~&gt; <span class="sgr-4 sgr-32">vim</span><span class="sgr-4"> ~/.config/elvish/rc.elv </span>                       <span class="sgr-7">xiaq@macarch</span>
```
xiaq, line 1: Don't change the layout of the ttyshots; they have the same width of 58.

It also seems that you're editing this by hand - it's better to just retake those ttyshots, the process is documented in https://github.com/elves/elvish/tree/master/website/ttyshot.

xiaq (approved):

website/ttyshot/tour/history-walk.html
krader1961: > It also seems that you're editing this by hand 

Yes. I was attempting to keep the length of the spanned text the same by replacing the spaces with the same number of non-space chars added/removed. I did not double-check the result so it's possible I got it wrong.

krader1961: And I see that I did get it wrong in this instance but got it right in the history-list.html and lastcmd.html examples. /facepalm

krader1961: @xiaq, I would recommend modifying the `edit:-dump-buf` ttyshot generator to include a comment in the generated HTML that reminds people to use the directions at https://github.com/elves/elvish/tree/master/website/ttyshot rather than manually editing the HTML.

krader1961: FYI, I've created a change that implements my suggestion vis-a-vis the generated ttyshot output and the documentation for creating them. However, since I can't get the existing directions to work without my change I can't verify my change is correct.

PR #1458, opened 2021-12-26
Small fixup for Function doc

iandol:
The chunk starting "There must be no space between ] and {; ..." is no longer relevant.


xiaq:
Thanks!

PR #1466, opened 2022-01-03
LSP that only supports parsing errors

pontaoski:


pontaoski:
the main purpose of this more than anything else is laying down the infrastructure/boilerplate you need for an lsp authored in go. adding things from here is fairly easy: add the method with the correct types, and note it in the method map. additional diagnostics can be done in functions called from evaluate.

xiaq:
There's a CI error caused by the dependency not being present in `go.mod` in `website`, which depends on the main Go module. Try running `go mod download` in the `website` directory and that should fix it.

xiaq (approved):
Thanks for the PR, this is a decent starting point.

I'd like this to be part of the `elvish` binary itself (i.e. `elvish -lsp`) instead of a separate command. The benefit is that the user will always have an LSP server that is consistent with their actual shell. The downside is making the binary larger; but most (if not all) of LSP functionalities should share the same implementation as the corresponding editor, the only added code is the LSP server, which seems to be small enough.

I've noted down all the places that should be improved; but I'll merge this and make these improvements myself.

cmd/elvish-lsp/server.go
```diff
@@ -0,0 +1,132 @@
...
+		if c == '\n' {
+			col = 0
+			line++
+		} else {
+			col++
+		}
+
+		if i == idx {
+			return lsp.Position{Line: line, Character: col}
+		}
```
xiaq, line 80: This return should be placed at the beginning of the loop body, instead of the end.

PR #1474, opened 2022-01-13
Implement unix:ulimit

### LINKED ISSUE #959: Implement a unix:ulimit command
Not many people need the ability to manipulate resource limits and there are awkward workarounds for elvish not having a `ulimit` builtin. So this isn't a high priority but it would be nice if elvish had a `unix:ulimit` command. The question is how to implement the options. For example, traditionally you would use `ulimit -c` to view or change the core size limit. Rather than map those to Elvish option syntax (e.g., `unix:ulimit &core-size`) I think subcommand style makes more sense: `unix:ulimit core-size`. The only options would be `&soft` and `&hard` to specify whether the soft or hard limit is to be displayed or changed. Running `unix:ulimit` with no other arguments would be equivalent to the traditional `ulimit -a` to display all limits.

krader1961:
Resolves #959

### DIFF
--- a/0.18.0-release-notes.md
+++ b/0.18.0-release-notes.md
@@ -41,3 +41,5 @@ or compiled, even if it is not executed:
 
 -   A new `call` command to call a command, using a list for and a map for
     options.
+
+-   A new `unix:ulimit` command ([#959](https://b.elv.sh/959)).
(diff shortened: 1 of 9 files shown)


xiaq:
The command implemented in this PR is a reasonable improvement over bash's `ulimit`, but a text-oriented API isn't an idiomatic one for Elvish.

I looked into the getrlimit/setrlimit API and implemented a binding via a variable in fb6d2fceafe1c08f942db64ac26ce82158cd0e61, so I'll close this now. Thanks for the interest anyway.

PR #1475, opened 2022-01-13
Update language spec metacharacter table for new lambda syntax

cherryblossom000:


website/ref/language.md
```diff
@@ -48,17 +48,17 @@ A **whitespace** is any of the following:
...
-| `;`           | Separating pipelines                                        |
-| `<` and `>`   | Introducing IO redirections                                 |
-| `(` and `)`   | Enclosing output captures                                   |
-| `[` and `]`   | Enclosing list literals, map literals or function signature |
-| `{` and `}`   | Enclosing lambda literals or brace expressions              |
+| Metacharacter | Use                                                          |
+| ------------- | ------------------------------------------------------------ |
+| `$`           | Referencing variables                                        |
+| `*` and `?`   | Forming wildcard                                             |
+| `\|`          | Separating forms in a pipeline; enclosing function signature |
```
krader1961, line 55: I would wait for feedback from @xiaq before changing this but I think verbiage like `function parameter signature` is clearer. Too, it would help if the phrase was a link to https://elv.sh/ref/language.html#function. Lastly, the related text at https://elv.sh/ref/language.html#function could use some improvement. Among other things the current verbiage implies the opening brace must be followed by whitespace but that isn't true since a parameter signature can immediately follow the opening brace without whitespace; e.g., `{|arg| ... }`.

xiaq: "Function signature" is fine. I'll add the links.

Issue #1476, opened 2022-01-14
Elvish noDaemon causes panic on cd command


MDr164:
Using 0.17.0 with the current noDaemon example code produces a stacktrace whenever one tries to change the directory. This might be related to chdir calling into daemon specific functions which are nil and then cause a nil pointer dereference.

```
goroutine 1 [running]:
src.elv.sh/pkg/sys.DumpStack()
	/home/user/go/src/github.com/u-root/u-root/vendor/src.elv.sh/pkg/sys/dumpstack.go:10 +0x66
src.elv.sh/pkg/shell.handlePanic()
	/home/user/go/src/github.com/u-root/u-root/vendor/src.elv.sh/pkg/shell/interact.go:144 +0x5a
panic({0x764940, 0x990a60})
	/usr/lib/go/src/runtime/panic.go:1038 +0x215
src.elv.sh/pkg/edit.initLocation.func2({0xc000018fe3, 0x8})
	/home/user/go/src/github.com/u-root/u-root/vendor/src.elv.sh/pkg/edit/listing.go:128 +0x58
src.elv.sh/pkg/eval.(*Evaler).Chdir(0xc0001a8000, {0xc000018fe3, 0x8})
	/home/user/go/src/github.com/u-root/u-root/vendor/src.elv.sh/pkg/eval/eval.go:322 +0xd5
src.elv.sh/pkg/eval.cd(0x2, {0xc00005d960, 0x1, 0x1})
	/home/user/go/src/github.com/u-root/u-root/vendor/src.elv.sh/pkg/eval/builtin_fn_fs.go:57 +0xbe
reflect.Value.call({0x758fa0, 0x7bb208, 0x30}, {0x7a5b9d, 0x4}, {0xc0003724e0, 0x2, 0x18})
	/usr/lib/go/src/reflect/value.go:556 +0x845
reflect.Value.Call({0x758fa0, 0x7bb208, 0x748a00}, {0xc0003724e0, 0x2, 0x2})
	/usr/lib/go/src/reflect/value.go:339 +0xc5
src.elv.sh/pkg/eval.(*goFn).Call(0xc0000f72d0, 0xc000301650, {0xc00005d930, 0x1, 0x7}, 0xc0003724b0)
	/home/user/go/src/github.com/u-root/u-root/vendor/src.elv.sh/pkg/eval/go_fn.go:230 +0xb46
src.elv.sh/pkg/eval.(*formOp).exec(0xc000344370, 0xc000301650)
[...]

PR #1477, opened 2022-01-14
pkg/edit: Don't call AddDir on storedefs if it is not ready

walterchris:
If storedef.store is not initialized i.e. == nil this leads to a panic. Before calling we should check if it is save to call storedef.store or not.

Signed-off-by: Christian Walter <christian.walter@9elements.com>

### DIFF
--- a/pkg/edit/listing.go
+++ b/pkg/edit/listing.go
@@ -124,9 +124,12 @@ func initLocation(ed *Editor, ev *eval.Evaler, st storedefs.Store, commonBinding
 			// TODO(xiaq): Surface the error.
 			return
 		}
-		st.AddDir(wd, 1)
+		// If st is not ready - don't try to call it
+		if st != nil {
+			st.AddDir(wd, 1)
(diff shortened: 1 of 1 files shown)


MDr164:
Related to https://github.com/elves/elvish/issues/1476

pkg/edit/listing.go
```diff
@@ -124,9 +124,12 @@ func initLocation(ed *Editor, ev *eval.Evaler, st storedefs.Store, commonBinding
...
-		st.AddDir(wd, 1)
+		// If st is not ready - don't try to call it
+		if st != nil {
+			st.AddDir(wd, 1)
+		}
 		kind, root := workspaceIterator.Parse(wd)
-		if kind != "" {
+		if kind != "" && st != nil {
 			st.AddDir(kind+wd[len(root):], 1)
 		}
```
xiaq, line 134: This entire block can be surrounded by `if st != nil`.

PR #1479, opened 2022-01-17
Document all elvish CLI flags

krader1961:
There was a recent question for which the answer was to use the
`-compileonly` flag but the user didn't realize that because there was
no mention on https://elv.sh/ref/command.html. Similarly, other flags,
like `-decpreation-level`, would benefit from more verbiage than shown by
`elvish -help`.

### DIFF
--- a/website/ref/command.md
+++ b/website/ref/command.md
@@ -90,5 +90,70 @@ If the legacy `~/.elvish/lib` directory exists, it is also searched.
 
 # Other command-line flags
 
-Running `elvish -help` lists all supported command-line flags, which are not
-repeated here.
+-   `-buildinfo`: Output information about how `elvish` was built then quit. See
+    also `-version` and `-json`. For example:
+
+    ```
+    ~> elvish -buildinfo
+    Version: 0.18.0-dev.4350ad48b1bb798068ec6a01f149382cd0607e2f
+    Go version: go1.17.2
+    Reproducible build: true
+    ```
+
+-   `-compileonly`: Compile the Elvish program without risking any side-effects
+    from executing the program. This is useful to identify problems with an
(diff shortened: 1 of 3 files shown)


krader1961:
Updated the PR to resolve conflicts with recent commits that refactored how CLI flags are registered.

website/ref/command.md
```diff
@@ -90,5 +90,70 @@ If the legacy `~/.elvish/lib` directory exists, it is also searched.
...
+    Version: 0.18.0-dev.4350ad48b1bb798068ec6a01f149382cd0607e2f
+    Go version: go1.17.2
+    Reproducible build: true
+    ```
+
+-   `-compileonly`: Compile the Elvish program without risking any side-effects
+    from executing the program. This is useful to identify problems with an
+    Elvish program that can be detected at compile time. At present it cannot be
+    used for verifying that the interactive config is syntactically valid
+    without executing the config. See also `-json`.
```
xiaq, line 107: > At present it cannot be
> used for verifying that the interactive config is syntactically valid
> without executing the config.

What is this referring to?

krader1961: That you can't simply run `elvish -compileonly` to verify the syntax of your *rc.elv*. I was surprised by that behavior so I added that sentence. But it is arguably a corner case that 99.9% of users will never care about. Note that you can't explicitly verify it due to the magic `edit` namespace:

```
> elvish -compileonly ~/.config/elvish/rc.elv
compilation error: cannot find variable $edit:completion:matcher
/Users/krader/.config/elvish/rc.elv, line 96: set edit:completion:matcher[''] = {|p| edit:match-prefix &smart-case $p }
Exception: elvish exited with 2
[tty 8], line 1: elvish -compileonly ~/.config/elvish/rc.elv
```

Issue #1005, opened 2020-05-05
Input buffering problem on Windows


Xjs:
In my quest to use more Go software I use [gopass](https://github.com/gopasspw/gopass) as my password manager. However, on Windows, using Elvish, gopass doesn't accept any input.

I have boiled down the problem to the [ReadLine method used by gopass](https://github.com/gopasspw/gopass/blob/master/pkg/termio/reader.go) and can reproduce the issue with the following little Go program:

```go
package main

import (
	"bytes"
	"fmt"
	"io"
	"log"
	"os"
)

func main() {
	fmt.Print("Enter line: ")

	var input bytes.Buffer
	buf := make([]byte, 1)

	for {
		n, err := os.Stdin.Read(buf)
		if err == io.EOF {
			break
		} else if err != nil {
			log.Println(err)
			break
		}
		if n == 0 {
			continue
		}
		if buf[0] == '\n' {
			break
		}
		input.WriteByte(buf[0])
	}

	fmt.Printf("%s\n", input.String())
}

```

This program, when executed from Elvish under Windows with `go run`, will not echo typed input from the terminal, and only terminate on ^Z (which signals EOF) -- but correctly read the input in this case.

When executed from Elvish under Linux, or when executed from other shells under Windows (tried MSYS sh bundled with Git for Windows and Powershell), the program will read and echo input as expected, and terminate on newlines.

I'm not sure how to start debugging, but I'm happy to help. If you think this is not the way a Go program should read unbuffered input at all, I'm happy to raise an issue with gopass as well. :)

Xjs:
I just noticed it might be the local changes I made to elvish in https://github.com/Xjs/elvish/commit/28fa32e00c7de50322e354b1f2327b2f1e8f4906 that cause this behaviour. I'm sorry.

Issue #798, opened 2019-03-06
Command output uses unsupported sequences on Windows


capnspacehook:
I know that in the README it states that Windows support is experimental, but I'd thought I'd post this issue here as it doesn't seem like a hard problem to fix.

Colors work great on elvish on Windows, but whenever an exception is thrown, for example when a program is called that is outside of the PATH, the output of the exception thrown is weird: 

```
Exception: [31;1mexec: "ls": executable file not found in %PATH%[m
[tty], line 1: [1;4mls[m
```

Edit: it appears this happens not only on exceptions. I ran an example command from the Elvish website, `put lorem ipsum | count`, expecting the output to be: `2`. But instead this was my output: `▶ 27h[?2004l`. Even though the arrow shows up here on Github, it doesn't in Powershell or Command Prompt.

xiaq:
Finally took some time to fix this after I had spent several days developing on a Windows box :)

Enjoy the Elvish experience on Windows!

Xjs:
Something similar to this is still occuring for me on Windows using https://github.com/elves/elvish/commit/92cb80d89a6a4050033936dd8cce68a2debc811f . I used https://github.com/Xjs/elvish/commit/28fa32e00c7de50322e354b1f2327b2f1e8f4906 as a workaround for some time now, but that broke input in unbuffered programs. Any ideas on how to fix it?

Output:

```
Exception: ←[31;1mgit exited with 141←[m
[tty 7], line 1: ←[1;4mgit log←[m
```

Screenshot:

![image](https://user-images.githubusercontent.com/58270/81066074-d8d4de80-8edc-11ea-97e9-f7b4a00d08a8.png)

Screenshot with my fix:

![image](https://user-images.githubusercontent.com/58270/81066197-0883e680-8edd-11ea-8d45-09df27f707ba.png)

Issue #1165, opened 2020-10-04
On Windows running external command inside the prompt function causes problems


krader1961:
Running elvish on Windows+MSYS2 over a SSH connection results in an unusable interactive experience. This is what I see when I type "ls" then [enter]:

```
←[32m▶ ←[m'$echo~'l
echo is \msys64\usr\bin\echo.exe
←[6C←[?25h←[?25l                                                                                                                                                             WINDOWS10\krader@Windows10
←[6C←[?25h←[?25l                                                                                                                                                             WINDOWS10\krader@Windows10
←[6C←[Kl                                                                                                                                                                      ←[0;7mWINDOWS10\krader@Win
←[6C←[K←[0;31ml←[0;m                                                                                                                                                                      ←[0;7mWINDOWS1
←[6C←[K←[0;32mls←[0;m                                                                                                                                                                     ←[0;7mWINDOWS1
←[8C←[?25h←[?25l10←[0;m
cmd  CONTRIBUTING.md  cover  Dockerfile  elvish.exe  go.mod  go.sum  LICENSE  main.go  Makefile  NEXT-RELEASE.md  pkg  README.md  resources  started  tools  website
←[6C←[?25h5l                                                                                                                                                                 WINDOWS10\krader@Windows10
```

Text sequences such as `←[8C` literally appear on the terminal. I noticed this initially after switching from my custom prompt functions to module `github.com/zzamboni/elvish-themes/chain`. I eventually determined it is caused by running any external command within a function run while evaluating the prompt.

Running `elvish -norc` or otherwise ensuring your interactive config does not define a left or right prompt function that runs any external commands results in a usable interactive experience. If you force the running of an external command the terminal output is very confused because it makes some (all?) terminal escape sequences visible. For example, simply run this interactively after starting elvish:

```
edit:prompt = { e:echo -n "wtf > " }
```

Changing the `e:echo -n` to `print` results in a usable interactive shell without the unexpected output that garbles the display.

krader1961:
I just updated my *~/.elvish/rc.elv* script to explicitly assigns to `paths` rather than relying on the inherited `PATH` env var. Because I made a couple of typos I see these warnings:

```
←[31mWarning: directory C:\WINDOWS\system32\config\systemprofile\AppData\Local\Microsoft\WindowsApps in $paths does not exist.←[m
←[31mWarning: directory C:\msys64\Microsoft\WindowsApps in $paths does not exist.←[m
```

[...]

PR #1484, opened 2022-01-22
On Windows, implement sanitize in pkg/cli/term/setup_windows.

### LINKED ISSUE #1182: TTY modes left in inconsistent state when external command is killed
Consider this interactive Elvish scenario:

1) You run an external command (e.g., an editor such as `vim`) that modifies the TTY behavior.

1) That command dies in a manner which doesn't give it an opportunity to restore the TTY behavior that existed before it changed that behavior.

You are now left with an interactive Elvish session that may, or may not, be usable. I noticed this while working on issue #661. See also issue #232 which is essentially the same issue but was limited to what happens when Elvish launches a */bin/sh* rescue shell.

moko256:
This PR will fixes a part of (not all) #1182, and the last of #798.
I tested on the case in #1005.

I tested with vim in Git for Windows 2.34.1.windows.1 that changes stdout's console modes.
1. Open vim
2. Close vim
3. Cause a error in elvish

Before this PR, the indent in output after this will broken.

### DIFF
--- a/pkg/cli/term/setup.go
+++ b/pkg/cli/term/setup.go
@@ -16,8 +16,8 @@ func Setup(in, out *os.File) (func() error, error) {
 }
 
 // SetupGlobal sets up the terminal for the entire Elvish session.
-func SetupGlobal() func() {
-	return setupGlobal()
+func SetupGlobal(in, out *os.File) func() {
+	return setupGlobal(in, out)
 }
 
 // Sanitize sanitizes the terminal after an external command has executed.
--- a/pkg/cli/term/setup_unix.go
+++ b/pkg/cli/term/setup_unix.go
@@ -52,7 +52,7 @@ func setup(in, out *os.File) (func() error, error) {
 	return restore, errSetupVT
 }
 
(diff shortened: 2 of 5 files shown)


krader1961:
This is similar to the fix I started working on after opening issue #1182. It does appear to fix the problems I've observed on Windows. It does not fix issue #1182 on UNIX systems. Having said that, if the change otherwise looks good I recommend it be merged since it does fix a major annoyance with styled output on Windows. But it should not close issue #1182 since that is still unresolved.

moko256:
Thanks for comment. I updated the description not to close the issue.

krader1961:
@xiaq: I recommend merging this even though it is incomplete; i.e., does not fix #1182. It does fix the styled output on Windows; issue #1165. It is a stepping stone to a better interactive experience.

xiaq:
@moko256 Thanks for the contribution, especially the unit test!

The terminal setup/cleanup functions need a bit of reorganizing. I'll do that in a followup commit.

PR #1490, opened 2022-01-26
Use `src.elv.sh/pkg/rpc` instead of `net/rpc`.

moko256:
The `net/rpc` was removed (72fc2dbe4ac8f783f45f317326382880fcd5e431) but added again (8e117a287586bbce63ed132f21fafcff0c0e8344).
This PR removes it again to reduce binary size. (about 9,605 KB -> 8,391 KB, with `go build cmd\elvish\main.go`, on Windows)

### DIFF
--- a/pkg/daemon/server.go
+++ b/pkg/daemon/server.go
@@ -7,7 +7,6 @@ package daemon
 
 import (
 	"net"
-	"net/rpc"
 	"os"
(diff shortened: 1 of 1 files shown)


xiaq:
Thanks, it seems that `goimports` was being a bit too helpful here...

PR #1499, opened 2022-03-07
Improve searching PATH on Windows for completions

krader1961:
This was extracted from https://github.com/elves/elvish/pull/1326. It
augments that change by honoring $E:PATHEXT and some other
minor improvements.  This matches the behavior of the Go stdlib
os/exec.LookPath() function that is used to decide if what has been typed
(in a non-completion context) represents an external command.

### DIFF
--- a/pkg/fsutil/fsutil_unix.go
+++ b/pkg/fsutil/fsutil_unix.go
@@ -1,0 +1,11 @@
+//go:build !windows && !plan9
+// +build !windows,!plan9
+
+package fsutil
+
+import "os"
+
+// IsExecutable returns true if the stat object refers to an executable file on UNIX platforms.
+func IsExecutable(stat os.FileInfo) bool {
+	return !stat.IsDir() && (stat.Mode()&0o111 != 0)
+}
--- a/pkg/fsutil/fsutil_windows.go
+++ b/pkg/fsutil/fsutil_windows.go
@@ -1,0 +1,44 @@
+package fsutil
+
+import (
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"src.elv.sh/pkg/env"
+)
+
+// isExecutable determines whether pathStr refers to an executable file in PATH. It honors E:PATHEXT
+// but defaults to extensions ".com", ".exe", ".bat", ".cmd" if that env var isn't set.
+func isExecutable(pathStr string) bool {
+	var validExts []string
+	if pathext := os.Getenv(env.PATHEXT); pathext != "" {
+		for _, e := range strings.Split(strings.ToLower(pathext), string(filepath.ListSeparator)) {
+			if e == "" {
+				continue
+			}
+			if e[0] != '.' {
+				e = "." + e
+			}
+			validExts = append(validExts, e)
+		}
+	} else {
+		validExts = []string{".com", ".exe", ".bat", ".cmd"}
+	}
+
+	ext := strings.ToLower(path.Ext(pathStr))
+	for _, valid := range validExts {
+		if ext == valid {
+			return true
+		}
+	}
+	return false
+}
+
+// IsExecutable returns true if the stat object refers to a valid executable on Windows.
+// Note that on Windows file permissions are not checked but we do validate the file is not a
+// directory and has a recognized extension.
+func IsExecutable(stat os.FileInfo) bool {
+	return !stat.IsDir() && isExecutable(stat.Name())
+}
--- a/pkg/fsutil/search_windows_test.go
+++ b/pkg/fsutil/search_windows_test.go
@@ -1,6 +1,3 @@
-//go:build !windows && !plan9 && !js
-// +build !windows,!plan9,!js
-
 package fsutil
 
 import (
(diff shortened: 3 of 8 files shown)


pkg/fsutil/fsutil_unix.go
```diff
@@ -0,0 +1,11 @@
+//go:build !windows && !plan9
+// +build !windows,!plan9
+
+package fsutil
+
+import "os"
+
+// IsExecutable returns true if the stat object refers to an executable file on UNIX platforms.
```
xiaq, line 8: When exporting a function with platform-dependent implementations, I'd prefer putting the entry point in a shared file, even if it is just a call to the unexported version. This gives you a unified Godoc so you don't have to repeat things and don't have to skip between the platforms (see the platform dropdown on e.g. https://pkg.go.dev/golang.org/x/sys/unix#section-documentation) when reading the Godoc to get the whole picture.

xiaq, line 8: When exporting a function with platform-dependent implementations, I'd prefer putting the entry point in a shared file, even if it is just a call to the unexported version. This makes it the Godoc easier to read.

pkg/fsutil/fsutil_unix.go
```diff
@@ -0,0 +1,11 @@
+//go:build !windows && !plan9
```
xiaq, line 1: This file is a platform-specific part of search.go, so better name it `search_unix.go`.

xiaq, line 1: This file is a platform-specific part of search.go, so better name it `search_unix.go`.

pkg/fsutil/fsutil_windows.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"src.elv.sh/pkg/env"
+)
+
+// isExecutable determines whether pathStr refers to an executable file in PATH. It honors E:PATHEXT
+// but defaults to extensions ".com", ".exe", ".bat", ".cmd" if that env var isn't set.
+func isExecutable(pathStr string) bool {
+	var validExts []string
+	if pathext := os.Getenv(env.PATHEXT); pathext != "" {
+		for _, e := range strings.Split(strings.ToLower(pathext), string(filepath.ListSeparator)) {
```
xiaq, line 17: [`filepath.SplitList`](https://pkg.go.dev/path/filepath#SplitList) can make this slightly simpler.

xiaq, line 17: [`filepath.SplitList`](https://pkg.go.dev/path/filepath#SplitList) can make this slightly simpler.

pkg/fsutil/fsutil_windows.go
```diff
@@ -0,0 +1,44 @@
...
+	"path/filepath"
+	"strings"
+
+	"src.elv.sh/pkg/env"
+)
+
+// isExecutable determines whether pathStr refers to an executable file in PATH. It honors E:PATHEXT
+// but defaults to extensions ".com", ".exe", ".bat", ".cmd" if that env var isn't set.
+func isExecutable(pathStr string) bool {
+	var validExts []string
```
xiaq, line 15: Since the code is doing quite a bit of heavy lifting to build a slice anyway, better to build a set (`map[string]struct{}`) so that the check later can be done more easily.

xiaq, line 15: Since the code is doing quite a bit of heavy lifting to build a slice anyway, better to build a set (`map[string]struct{}`) so that the check later can be done more easily.

pkg/fsutil/search_windows_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	wantCmds := []string{"prog.bat", "prog.cmd", "PROG.EXE"}
+	gotCmds := []string{}
+	EachExternal(func(cmd string) { gotCmds = append(gotCmds, cmd) })
+
+	sort.Strings(wantCmds)
+	sort.Strings(gotCmds)
+	if !reflect.DeepEqual(wantCmds, gotCmds) {
+		t.Errorf("EachExternal want %q got %q", wantCmds, gotCmds)
+	}
```
xiaq, line 33: This can now use `cmp.Diff`.

xiaq, line 33: This can now use `cmp.Diff`.

xiaq (approved):

pkg/fsutil/fsutil_windows.go
```diff
@@ -0,0 +1,44 @@
...
+			if e[0] != '.' {
+				e = "." + e
+			}
+			validExts = append(validExts, e)
+		}
+	} else {
+		validExts = []string{".com", ".exe", ".bat", ".cmd"}
+	}
+
+	ext := strings.ToLower(path.Ext(pathStr))
```
xiaq, line 30: This should use `filepath.Ext` instead of `path.Ext`; `path` always assumes a path separator of `/`.

pkg/fsutil/fsutil_windows.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"src.elv.sh/pkg/env"
+)
+
+// isExecutable determines whether pathStr refers to an executable file in PATH. It honors E:PATHEXT
```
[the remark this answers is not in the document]
krader1961: Agreed. Good point. I'm curious what a careful review of the existing documentation would reveal regarding the ways in which env vars are referred to.

PR #1501, opened 2022-03-08
Replace ^ with math:pow in the documentation

krader1961:
A new user working through the documentation asked why `^ 2 10` resulted
in an error rather than `1024`.


xiaq:
This won't work without `use math` but it's a good opportunity to introduce how to import modules.

PR #1504, opened 2022-03-20
Fix typo in command.md.

nsf:
Source code uses $XDG_**STATE**_HOME for database file:
https://github.com/elves/elvish/blob/ab699bae8703af6fd7332aaa61e56072b1b0bb7b/pkg/shell/paths_unix.go#L46

Documentation incorrectly mentions $XDG_**DATA**_HOME:
https://github.com/elves/elvish/blob/ab699bae8703af6fd7332aaa61e56072b1b0bb7b/website/ref/command.md?plain=1#L49-L51

### DIFF
--- a/website/ref/command.md
+++ b/website/ref/command.md
@@ -46,8 +46,8 @@ history. Its path is determined as follows:
 
 -   Otherwise:
 
-    -   On UNIX (including macOS), `$XDG_DATA_HOME/elvish/db.bolt` is used,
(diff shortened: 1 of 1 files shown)


xiaq:
Thanks for the PR.

Also I just realized that you're the author of gocode - a huge thank-you for that too! People use gopls these days but gocode set a really high bar for Go's editor support.

PR #1505, opened 2022-03-21
Bump gccgo to 12

zhsj:
gccgo-12 implements go1.18

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -12,17 +12,16 @@ test_task:
     fingerprint_script: cat go.sum
     folder: $HOME/go/pkg/mod
   matrix:
-    # Re-enable gccgo when it supports Go 1.18.
-    #- name: Test on gccgo
-    #  container:
-    #    image: debian:unstable-slim
-    #  setup_script:
-    #    - apt-get -y update
-    #    - apt-get -y install ca-certificates gccgo-11
-    #    - ln -sf /usr/bin/go-11 /usr/local/bin/go
-    #  env:
-    #    # gccgo doesn't support race test
-    #    TEST_FLAG: ""
+    - name: Test on gccgo
+      container:
+        image: debian:unstable-slim
+      setup_script:
+        - apt-get -y update
+        - apt-get -y install ca-certificates git gccgo-12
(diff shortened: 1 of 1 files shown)


zhsj:
hmm, though gccgo-12 claims it support go1.18, but the generic seems broken.

xiaq:
Ha, seems that it uses a hacky way to implement the `any` alias by making it present in every module - which causes a symbol conflict when using dot imports. Hopefully this is easy enough to fix.

krader1961:
Personally, I'm not a fan of dot imports. However, package `evaltest` is a special case. Using qualified identifiers for that package, even if an alias such as `et` was used, would represent a significant decrease in readability.

I tried a lot of different google searches (e.g., `gccgo 1.18 import "redefinition" of any`) but couldn't find an open issue for this problem. Which is surprising, but perhaps it's really the case that no one has yet filed a bug report. @zhsj, you should probably open a bug report against gccgo. Or, if you managed to find an existing issue, perhaps add a comment with some relevant text that would make finding it easier.

krader1961:
@xiaq: FWIW, there is some inconsistency in the use of dot imports that seems to be purely a historical artifact of how the code evolved rather than for a good reason. For example, consider the output of `grep src.elv.sh/pkg/testutil **.go`. Most of the imports are qualified but there are some dot imports of that package. Any objection to cleaning up such instances to consistently use qualified imports?

krader1961:
Looking at the imports of `"src.elv.sh/pkg/tt"` we can see there is a slight majority using qualified imports:

```
> grep -h src.elv.sh/pkg/tt **.go | sort | uniq -c
  26    "src.elv.sh/pkg/tt"
  24    . "src.elv.sh/pkg/tt"
   1 import "src.elv.sh/pkg/tt"
```

I don't see any reason why all of those should not be qualified for consistency; i.e., the dot imports changed to qualified imports. Which then begs the question why `evaltest` also warrants an unqualified "dot" import. Notwithstanding my previous comment that packages like `evaltest` and `clitest` are somewhat special.

xiaq:
Tried gccgo 12 on a local debian installation, still doesn't work.

Please feel free to reopen this when gccgo 12 works.

PR #1508, opened 2022-03-27
Dot imports

krader1961:
TBD is what to do about packages `clitest`, `evaltest`, and `progtest`. All three have a small number of qualified imports and many more unqualified imports. My preference is to convert all of them to qualified imports with aliases `ct`, `et`, and `pt` respectively. But that may be a bridge too far.

### DIFF
--- a/pkg/cli/app_test.go
+++ b/pkg/cli/app_test.go
@@ -8,7 +8,7 @@ import (
 	"testing"
 	"time"
 
-	. "src.elv.sh/pkg/cli"
+	"src.elv.sh/pkg/cli"
 	. "src.elv.sh/pkg/cli/clitest"
 	"src.elv.sh/pkg/cli/term"
 	"src.elv.sh/pkg/cli/tk"
@@ -46,7 +46,7 @@ func TestReadCode_RestoresTTYBeforeReturning(t *testing.T) {
 }
 
 func TestReadCode_ResetsStateBeforeReturning(t *testing.T) {
-	f := Setup(WithSpec(func(spec *AppSpec) {
+	f := Setup(WithSpec(func(spec *cli.AppSpec) {
 		spec.CodeAreaState.Buffer.Content = "some code"
 	}))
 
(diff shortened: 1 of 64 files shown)


xiaq:
Most of the dot imports are test files importing the package under test.

This is a relatively common use case, even explicitly [allowed](https://github.com/golang/go/wiki/CodeReviewComments#import-dot) in the official Go codebase (and there are [many examples](https://cs.opensource.google/search?q=%5E%5Ct%5C.%5C%20%22.*%22$%20file:_test%5C.go$&ss=go) of it). I have the same stance here.

xiaq:
Most of the dot imports are test files importing the package under test.

This is a relatively common use case, even explicitly [allowed](https://github.com/golang/go/wiki/CodeReviewComments#import-dot) in the official Go codebase (and there are [many examples](https://cs.opensource.google/search?q=%5E%5Ct%5C.%5C%20%22.*%22$%20file:_test%5C.go$&ss=go) of it). I have the same stance here.

krader1961:
> This is a relatively common use case, even explicitly [allowed](https://github.com/golang/go/wiki/CodeReviewComments#import-dot) in the official Go codebase (and there are [many examples](https://cs.opensource.google/search?q=%5E%5Ct%5C.%5C%20%22.*%22$%20file:_test%5C.go$&ss=go) of it). I have the same stance here.

Okay, I can see I got carried away and changed many instances that don't strictly need a dot import but the use of a dot import is justified because the test file would be in the same package as the dot import if that didn't create a circular import. However, what got me started on this change were files like *pkg/eval/options_test.go* which do a dot import of *pkg/tt* even though there is no circular dependency and the test file is not testing that package. That was commit 44e2134b in this pull-request. Do you have an objection to eliminating those unnecessary, and inconsistent, dot imports?

xiaq:
Yes, the changes to remove dot imports of `testutil` are fine.

Re `clitest`, `evaltest` and `progtest` - I don't feel strongly about them, but if you want to eliminate the dot imports, my preference is

- Keep the current descriptive names;
- For imports of `evaltest`, make a local alias of `That` (`var That = evaltest.That`) - there is already a test doing this;
- For imports of `progtest`, make a local alias of `ThatElvish` (`var ThatElvish = progtest.ThatElvish`).

PR #1514, opened 2022-04-04
Small doc fix for set-env

iandol:
Replaces `set E:X = foo` with `set-env X foo` to be a bit more consistent with the others functions...

### DIFF
--- a/pkg/eval/builtin_fn_env.go
+++ b/pkg/eval/builtin_fn_env.go
@@ -34,7 +34,7 @@ var ErrNonExistentEnvVar = errors.New("non-existent environment variable")
 // Unset an environment variable. Example:
 //
 // ```elvish-transcript
-// ~> set E:X = foo
+// ~> set-env X foo
 // ~> unset-env X
 // ~> has-env X
 // ▶ $false


pkg/eval/builtin_fn_env.go
```diff
@@ -34,7 +34,7 @@ var ErrNonExistentEnvVar = errors.New("non-existent environment variable")
 // Unset an environment variable. Example:
 //
 // ```elvish-transcript
-// ~> set E:X = foo
+// ~> set-env X foo
```
krader1961, line 37: My preference is to keep the original statement, with a comment, and add the `set-env` statement. I think it's important to recognize that the `E:` namespace exists in these examples. Which obviously applies to closely related examples such as `del E:PATH` versus `unset-env PATH`. The `*-env` commands exist to support indirect env var names; e.g., `var y = X; set-env $y foo`.

iandol: The other `*-env` command docs don't reference the `E:` namespace. Do you suggest something like this:

```
~> set E:X = foo # could also use e.g. set-env X foo
```

xiaq:
I don't it necessary to avoid using the `E:` namespace in these commands.

I think a more important improvement to the docs of the `*-env` functions is to explicitly say what their corresponding operation using the `E:` namespace looks like.

xiaq:
a57275564b28aaf217d0e385d972cb878c6e05b3 now documents the relationship between the *-env commands and the E: namespace.

PR #1531, opened 2022-05-01
Correctly encode strings with invalid UTF-8

### LINKED ISSUE #1528: Completion behavior on filenames with non-printable characters
```
$ e:ls
å¤§è°·å??å¤« - ã?¨ã??ã??ã??ã??ã?¥ã??ã?¤ã?? O.S.T.torrent
$ e:ls "å¤§è°·å\x92\x8cå¤« - ã\x81¨ã\x81\x8dã\x82\x81ã\x81\x8dã\x83\x88ã\x82¥ã\x83\x8aã\x82¤ã\x83\x88 O.S.T.torrent" 
ls: å¤§è°·å??å¤« - ã?¨ã??ã??ã??ã??ã?¥ã??ã?¤ã?? O.S.T.torrent: No such file or directory
Exception: ls exited with 1
[tty 27], line 1: e:ls "å¤§è°·å\x92\x8cå¤« - ã\x81¨ã\x81\x8dã\x82\x81ã\x81\x8dã\x83\x88ã\x82¥ã\x83\x8aã\x82¤ã\x83\x88 O.S.T.torrent" 
$ e:ls -B
å¤§è°·å\302\222\302\214å¤« - ã\302\201¨ã\302\201\302\215ã\302\202\302\201ã\302\201\302\215ã\302\203\302\210ã\302\202¥ã\302\203\302\212ã\302\202¤ã\302\203\302\210 O.S.T.torrent
$ e:ls "å¤§è°·å\302\222\302\214å¤« - ã\302\201¨ã\302\201\302\215ã\302\202\302\201ã\302\201\302\215ã\302\203\302\210ã\302\202¥ã\302\203
                      \302\212ã\302\202¤ã\302\203\302\210 O.S.T.torrent"
å¤§è°·å??å¤« - ã?¨ã??ã??ã??ã??ã?¥ã??ã?¤ã?? O.S.T.torrent
```

The filename is wrongly decoded thus contains unprintable chars. The second line is completed by default tab completion, and not recognized by `ls` (and other programs). However the filename printed with `ls -B` can be recognized.

 `-B` is

> Force printing of non-printable characters (as defined by ctype(3) and current locale settings) in file names as \xxx, where
> xxx is the numeric value of the character in octal.  This option is not defined in IEEE Std 1003.1-2008 (“POSIX.1”).

Right now I'm not going to request changes on the default completion behavior, because Windows might have different behavior on handling pathname with unprintable chars, and this is really an edge case. However, I'd like to know is there a way to config the default filename completion to handle unprintable chars in the same fashion of `ls -B`?

krader1961:
In addition to fixing the encoding bugs standardize on \xNN notation for
bytes rather than \uNNNN; e.g., \x00 rather than \u0000.

Also, update the documentation for double-quoted strings to include the
\e sequence.

Fixes #1528

### DIFF
--- a/pkg/eval/compile_value_test.go
+++ b/pkg/eval/compile_value_test.go
@@ -73,8 +73,9 @@ func TestMapLiteral(t *testing.T) {
 func TestStringLiteral(t *testing.T) {
 	Test(t,
 		That(`put 'such \"''literal'`).Puts(`such \"'literal`),
-		That(`put "much \n\033[31;1m$cool\033[m"`).
-			Puts("much \n\033[31;1m$cool\033[m"),
+		That(`put "much \n\033[31;1m$cool\033[m"`).Puts("much \n\033[31;1m$cool\033[m"),
+		That(`put "x\302\222\302\214y"`).Puts("x\u0092\u008cy"),
+		That(`put "x\u0092\u008cy"`).Puts("x\u0092\u008cy"),
 	)
 }
 
--- a/pkg/eval/value_test.go
+++ b/pkg/eval/value_test.go
@@ -16,6 +16,7 @@ var reprTests = []struct {
 	{"a\nb", `"a\nb"`},
 	{"foo bar", "'foo bar'"},
 	{"a\x00b", `"a\x00b"`},
+	{"a\033b", `"a\eb"`},
 	{true, "$true"},
 	{false, "$false"},
 	{vals.EmptyList, "[]"},
--- a/pkg/parse/quote.go
+++ b/pkg/parse/quote.go
@@ -3,6 +3,7 @@ package parse
 import (
 	"bytes"
 	"unicode"
+	"unicode/utf8"
 )
 
 // Quote returns a valid Elvish expression that evaluates to the given string.
@@ -22,8 +23,9 @@ func QuoteVariableName(s string) string {
 
 	// Keep track of whether it is a valid (unquoted) variable name.
 	bare := true
-	for _, r := range s {
-		if !unicode.IsPrint(r) {
+	for s2 := s; len(s2) > 0; {
+		r, w := utf8.DecodeRuneInString(s2)
+		if r == utf8.RuneError || !unicode.IsPrint(r) {
 			// Contains unprintable character; force double quote.
 			return quoteDouble(s)
 		}
@@ -31,6 +33,7 @@ func QuoteVariableName(s string) string {
 			bare = false
 			break
 		}
+		s2 = s2[w:]
 	}
 
 	if bare {
@@ -53,14 +56,16 @@ func QuoteAs(s string, q PrimaryType) (string, PrimaryType) {
 
 	// Keep track of whether it is a valid bareword.
 	bare := s[0] != '~'
-	for _, r := range s {
-		if !unicode.IsPrint(r) {
+	for s2 := s; len(s2) > 0; {
+		r, w := utf8.DecodeRuneInString(s2)
+		if r == utf8.RuneError || !unicode.IsPrint(r) {
 			// Contains unprintable character; force double quote.
 			return quoteDouble(s), DoubleQuoted
 		}
 		if !allowedInBareword(r, strictExpr) {
 			bare = false
 		}
+		s2 = s2[w:]
 	}
 
 	if q == Bareword && bare {
@@ -82,6 +87,8 @@ func quoteSingle(s string) string {
 	return buf.String()
 }
 
+// rtohex is optimized for the common cases encountered when encoding Elvish strings and should be
+// more efficient than using fmt.Sprintf("%x").
 func rtohex(r rune, w int) []byte {
 	bytes := make([]byte, w)
 	for i := w - 1; i >= 0; i-- {
@@ -99,26 +106,39 @@ func rtohex(r rune, w int) []byte {
 func quoteDouble(s string) string {
 	var buf bytes.Buffer
 	buf.WriteByte('"')
-	for _, r := range s {
+	for len(s) > 0 {
+		r, w := utf8.DecodeRuneInString(s)
+		if r == utf8.RuneError {
+			// An invalid UTF-8 sequence was seen -- encode first byte as a hex literal.
+			buf.WriteByte('\\')
+			buf.WriteByte('x')
+			buf.Write(rtohex(rune(s[0]), 2))
+			s = s[w:]
+			continue
+		}
+
+		// The common case -- a valid UTF-8 sequence was decoded so we have a rune.
 		if e, ok := doubleUnescape[r]; ok {
-			// Takes care of " and \ as well.
+			// Handle " and \ as well as things like \n.
 			buf.WriteByte('\\')
 			buf.WriteRune(e)
-		} else if !unicode.IsPrint(r) {
-			buf.WriteByte('\\')
-			if r <= 0xff {
-				buf.WriteByte('x')
-				buf.Write(rtohex(r, 2))
-			} else if r <= 0xffff {
-				buf.WriteByte('u')
-				buf.Write(rtohex(r, 4))
-			} else {
-				buf.WriteByte('U')
-				buf.Write(rtohex(r, 8))
-			}
-		} else {
+		} else if unicode.IsPrint(r) {
 			buf.WriteRune(r)
+		} else if r <= ' ' {
+			// Control char not handled by doubleUnescape.
+			buf.WriteByte('\\')
+			buf.WriteByte('x')
+			buf.Write(rtohex(r, 2))
+		} else if r <= 0xffff {
+			buf.WriteByte('\\')
+			buf.WriteByte('u')
+			buf.Write(rtohex(r, 4))
+		} else {
+			buf.WriteByte('\\')
+			buf.WriteByte('U')
+			buf.Write(rtohex(r, 8))
 		}
+		s = s[w:]
 	}
 	buf.WriteByte('"')
 	return buf.String()
--- a/pkg/parse/quote_test.go
+++ b/pkg/parse/quote_test.go
@@ -24,8 +24,7 @@ func TestQuote(t *testing.T) {
 
 		// Double quote when there is unprintable char.
 		Args("a\nb").Rets(`"a\nb"`),
-		Args("\x1b\"\\").Rets(`"\e\"\\"`),
-		Args("\x00").Rets(`"\x00"`),
+		Args("\000\x1b\"\\").Rets(`"\x00\e\"\\"`),
 		Args("\u0600").Rets(`"\u0600"`),         // Arabic number sign
 		Args("\U000110BD").Rets(`"\U000110bd"`), // Kathi number sign
 
@@ -47,6 +46,9 @@ func TestQuoteAs(t *testing.T) {
 		Args("a", SingleQuoted).Rets(`'a'`, SingleQuoted),
 		Args("\n", SingleQuoted).Rets(`"\n"`, DoubleQuoted),
 
+		// Verify bareword invalid UTF-8 case.
+		Args("bad\xffUTF-8", Bareword).Rets(`"bad\xffUTF-8"`, DoubleQuoted),
+
 		// Bareword tested above in TestQuote.
 	})
 }
@@ -57,5 +59,6 @@ func TestQuoteVariableName(t *testing.T) {
 		Args("foo").Rets("foo"),
 		Args("a/b").Rets("'a/b'"),
 		Args("\x1b").Rets(`"\e"`),
+		Args("bad\xffUTF-8").Rets(`"bad\xffUTF-8"`),
 	})
 }
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -118,6 +118,8 @@ represents Unicode codepoints in hexadecimal):
 
     -   `\r` is U+000D CR (carriage return).
 
+    -   `\e` is U+001B ESC (escape).
+
     -   `\"` is U+0022, the double quote `"` itself.
 
     -   `\\` is U+005C, the backslash `\` itself.
@@ -135,8 +137,8 @@ represents Unicode codepoints in hexadecimal):
     be used to write arbitrary byte sequences that are not necessary valid UTF-8
     sequences.
 
-    **Note**: `\0`, while supported by C, is invalid in Elvish; write `\000`
-    instead.
+    **Note**: `\0`, while supported by C, is invalid in Elvish; write `\x00` or
+    `\000` instead.
 
 -   The following escape sequences encode any Unicode codepoint using their
     numeric values:


krader1961:
I have no idea why `make -C website check-rellinks` failed. It passes on my local system and nothing in this P.R. introduces a relative link in the docs.

krader1961:
I just pushed the Elvish master branch to my Github fork and got the same check-rellinks failure: https://github.com/krader1961/elvish/runs/6250621681?check_suite_focus=true

So that failure isn't due to anything contained in this P.R.

pkg/eval/value_test.go
```diff
@@ -16,6 +16,11 @@ var reprTests = []struct {
 	{"a\nb", `"a\nb"`},
 	{"foo bar", "'foo bar'"},
 	{"a\x00b", `"a\x00b"`},
+	{"c\u0000d", `"c\x00d"`},
+	{"a\033b", `"a\eb"`},
+	{"c\u001bd", `"c\ed"`},
+	{"x\302\222\302\214y", `"x\u0092\u008cy"`},
+	{"x\u0092\u008cy", `"x\u0092\u008cy"`},
```
xiaq, line 23: This test is not intended to be an exhaustive test against the quoting mechanism; that is the responsibility of the test in the `parse` package. If you think the `parse` package test could use more test cases, add them there instead.

pkg/parse/quote.go
```diff
@@ -20,19 +21,19 @@ func QuoteVariableName(s string) string {
...
 	}
 
-	// Keep track of whether it is a valid (unquoted) variable name.
-	bare := true
-	for _, r := range s {
-		if !unicode.IsPrint(r) {
+	bare := true // keep track of whether it is a valid (unquoted) variable name
+	for s2 := s; len(s2) > 0; {
+		r, w := utf8.DecodeRuneInString(s2)
+		if r == utf8.RuneError || !unicode.IsPrint(r) {
```
xiaq, line 27: Only this line seems to be the necessary change in this function; iterating a string with invalid UTF-8 sequences produces utf8.RuneError too.

Please also refrain from making unnecessary stylistic changes; I prefer the existing comment format on line 23 and the empty line on line 35.

pkg/parse/quote.go
```diff
@@ -53,14 +54,16 @@ func QuoteAs(s string, q PrimaryType) (string, PrimaryType) {
 
 	// Keep track of whether it is a valid bareword.
 	bare := s[0] != '~'
-	for _, r := range s {
-		if !unicode.IsPrint(r) {
+	for s2 := s; len(s2) > 0; {
+		r, w := utf8.DecodeRuneInString(s2)
+		if r == utf8.RuneError || !unicode.IsPrint(r) {
```
xiaq, line 59: Same, this seems to be the only necessary change in this function.

pkg/parse/quote.go
```diff
@@ -99,26 +104,33 @@ func rtohex(r rune, w int) []byte {
...
-		if e, ok := doubleUnescape[r]; ok {
-			// Takes care of " and \ as well.
-			buf.WriteByte('\\')
-			buf.WriteRune(e)
-		} else if !unicode.IsPrint(r) {
-			buf.WriteByte('\\')
-			if r <= 0xff {
+	for len(s) > 0 {
+		r, w := utf8.DecodeRuneInString(s)
+		if r != utf8.RuneError { // the common case -- a valid UTF-8 sequence was decoded
```
xiaq, line 109: Handle the case of `r == utf8.RuneError` first, and you won't need an extra level of indentation.

Also please revert the change to the comment style; avoiding line-end comments is a conscious stylistic choice of mine.

pkg/parse/quote.go
```diff
@@ -99,26 +104,33 @@ func rtohex(r rune, w int) []byte {
 func quoteDouble(s string) string {
 	var buf bytes.Buffer
 	buf.WriteByte('"')
-	for _, r := range s {
```
xiaq, line 102: It's a bit cleaner to keep the `range` iteration, and you can still access the underlying byte using the index:

```go
for i, r := range s {
  if r == utf8.RuneError{
    // s[i] contains the invalid byte
  }
}
```

pkg/parse/quote_test.go
```diff
@@ -24,8 +24,8 @@ func TestQuote(t *testing.T) {
 
 		// Double quote when there is unprintable char.
 		Args("a\nb").Rets(`"a\nb"`),
-		Args("\x1b\"\\").Rets(`"\e\"\\"`),
-		Args("\x00").Rets(`"\x00"`),
+		Args("\000\x1b\"\\").Rets(`"\x00\e\"\\"`),
```
xiaq, line 27: The NUL byte is already tested in the test case below, why introduce it to this test case too?

pkg/parse/quote_test.go
```diff
@@ -24,8 +24,8 @@ func TestQuote(t *testing.T) {
...
+		Args("\u0000").Rets(`"\x00"`),
```
xiaq, line 28: The change from `\x00` to `\u0000` seems unnecessary; these two Go literals are equivalent, and the change doesn't have any effect on which Elvish code is hit.

pkg/parse/quote_test.go
```diff
@@ -47,7 +47,8 @@ func TestQuoteAs(t *testing.T) {
 		Args("a", SingleQuoted).Rets(`'a'`, SingleQuoted),
 		Args("\n", SingleQuoted).Rets(`"\n"`, DoubleQuoted),
 
-		// Bareword tested above in TestQuote.
+		// Bareword also tested above in TestQuote but verify invalid UTF-8 case here as well.
```
xiaq, line 50: These two things are unrelated, so don't write them in the same comment.

It's also better if you move this new test case above the original comment; these "foo tested elsewhere" comments should come last in a list of test cases.

```go
// Invalid UTF-8 sequences are quoted with \x
Args("bad\xffUTF-8", Bareword).Rets(`"bad\xffUTF-8"`, DoubleQuoted),

// Bareword tested above in TestQuote.
```

website/ref/language.md
```diff
@@ -109,15 +109,17 @@ represents Unicode codepoints in hexadecimal):
 
     -   `\t` is U+0009 HT (horizontal tabulation).
 
-    -   `\n` is U+000A LF (line feed), the standard line termination character
-        on Unix.
+    -   `\n` is U+000A LF (line feed, the standard line termination character on
+        Unix).
```
xiaq, line 113: The original formatting is fine; the "line feed" part serves to document what "LF" means.

pkg/eval/value_test.go
krader1961: I see your point. My first two tests are not a test of the Elvish quoting mechanism. They are a lower level test that the repr functionality produces the expected output.  You're correct that one of the two tests of the `\e` encoding should be removed since it is redundant. And the UTF-8 sequence, and Unicode codepoint, to Unicode codepoint tests belong in the parse package unit test so I'll move them there.

pkg/parse/quote.go
krader1961: > iterating a string with invalid UTF-8 sequences produces utf8.RuneError too 

Yes, although it technically returns `unicode.ReplacementChar` which happens (albeit presumably by design) to have the same value as `utf8.RuneError`. More importantly this formulation is, AFAICT, the preferred idiom for handling strings (and byte slices) that contain invalid UTF-8 sequences. I've read a lot of blog posts and Go issues that talk about the current behavior and the consensus opinion seems to be to use the `unicode/utf8` package in these situations rather than ranging over the string. Since we have to use the `utf8.DecodeRuneInString` pattern where we actually need to capture the invalid UTF-8 byte it's preferable to use the same pattern in places like this where we don't to ease the cognitive load on readers of the code.

> Please also refrain from making unnecessary stylistic changes; I prefer the existing comment format on line 23 and the empty line on line 35.

Understood and I'll revert that part of the change. But note that I disagree, at least in the case of the comment on line 23, because it only applies to the `bare` var; not the entire block of code and thus should be, IMHO, attached to the var declaration rather than a block comment.

pkg/parse/quote.go
krader1961: Agreed. In my defense it seemed sensible to handle the common case first but I agree that it makes more sense, on looking at the code again, to handle the error case first since it is very short and we can simply add a `continue` to that block and thus avoid indenting the common case.

pkg/parse/quote.go
krader1961: Yes, but as I mentioned in a prior comment it is better to use the utf8 package decode rune pattern consistently. That makes it easier for readers of the code to reason about the behavior. It also appears to be the current consensus in the Go community as a best practice for code that needs to iterate over strings/byte-slices that might contain invalid UTF-8. I've also seen comments by people like Ian Taylor (i.e., Go core developers) expressing regret about the current `range string` behavior.

pkg/parse/quote_test.go
krader1961: Good point. The intent was to verify that both forms (`"\000"` and `"\u0000"`) where handled identically by Elvish. But, obviously in hindsight, this is actually verifying that Go translates both sequences to a NUL byte. I'll remove the second, redundant, test. This does, however, raise the question of how to verify Elvish, rather than Go, interprets such sequences. Presumably there is Elvish test code that does so but I couldn't find it.

pkg/parse/quote_test.go
krader1961: Yes, per my reply to to your previous observation of this same point this is simply testing that Go translates both sequences to a NUL byte rather than testing that Elvish translates both sequences to a NUL byte. TBD is how to do the latter. I've eliminated the redundant test.

website/ref/language.md
krader1961: Okay, reverted. I made that change to move the "UNIX" text into the parenthetical because it only applies to UNIX like systems. But that is clearly outside the scope of this pull-request.

pkg/parse/quote_test.go
krader1961: Done

pkg/parse/quote.go
xiaq: > I've read a lot of blog posts and Go issues that talk about the current behavior and the consensus opinion seems to be to use the unicode/utf8 package in these situations rather than ranging over the string.

Can you provide some reference links for that, in particular the opinion from Ian Taylor (which you mentioned in another thread)?

I haven't been able to find any GitHub issues that discuss the behavior of the `range` loop over strings that contain invalid UTF-8 sequences in general, but GitHub's search functionality is pretty bad.

> Since we have to use the utf8.DecodeRuneInString pattern where we actually need to capture the invalid UTF-8 byte it's preferable to use the same pattern in places like this where we don't to ease the cognitive load on readers of the code.

You can always index the string to get the actual byte.

pkg/eval/compile_value_test.go
```diff
@@ -73,8 +73,9 @@ func TestMapLiteral(t *testing.T) {
 func TestStringLiteral(t *testing.T) {
 	Test(t,
 		That(`put 'such \"''literal'`).Puts(`such \"'literal`),
-		That(`put "much \n\033[31;1m$cool\033[m"`).
-			Puts("much \n\033[31;1m$cool\033[m"),
+		That(`put "much \n\033[31;1m$cool\033[m"`).Puts("much \n\033[31;1m$cool\033[m"),
+		That(`put "x\302\222\302\214y"`).Puts("x\u0092\u008cy"),
+		That(`put "x\u0092\u008cy"`).Puts("x\u0092\u008cy"),
```
xiaq, line 78: I still don't get why these two extra tests belong here; the interpretation of escape sequences is done in the parser, not in this package.

The existing test on line 76 does test for escape sequences, but that's intended to be a sanity check of whether double-quoted strings work as expected, and is not intended to be an in-depth test of escape sequences in general.

pkg/parse/quote.go
krader1961: Sadly, I can't find, using a google search, the message I know I read recently (but not necessarily published recently) on this topic wherein Ian Taylor advocated for using `utf8.DecodeRuneInString` rather than ranging over the string. However, see https://github.com/golang/go/issues/28599 which proposed augmenting ranging over a string to make it easier to handle invalid UTF-8.  See also the `Map` function in *src/strings/strings.go* in the Go source for how awkward it is to deal with invalid UTF-8 using `range string`. See also all the numerous uses of `utf8.DecodeRuneInString` in the Go source when iterating over, or otherwise handling, strings.

The bottom line is that I believe a pattern using `utf8.DecodeRuneInString` is easier to understand than one based on `range string`. Which approach is more efficient is probably the more salient point. I'll admit I did not do any performance tests.

pkg/eval/compile_value_test.go
krader1961: Yes, and no. I agree the second test isn't meaningful but believe the first test, of a sequence of valid UTF-8 bytes, is useful. I thought I had verified that both tests failed without the other changes in this PR but that is not true. Nonetheless, I think the first test is useful to help protect against future regressions.

pkg/parse/quote.go
```diff
@@ -99,26 +106,39 @@ func rtohex(r rune, w int) []byte {
...
+			buf.WriteByte('x')
+			buf.Write(rtohex(rune(s[0]), 2))
+			s = s[w:]
+			continue
+		}
+
+		// The common case -- a valid UTF-8 sequence was decoded so we have a rune.
 		if e, ok := doubleUnescape[r]; ok {
-			// Takes care of " and \ as well.
+			// Handle " and \ as well as things like \n.
```
xiaq, line 122: The new comment is confusing since the code does **not** specifically handle `"` or `\\`. The intention of the old comment to clarify that `"` and `\\` are handled as part of this branch since they belong to `doubleEscape`.

pkg/parse/quote.go
```diff
@@ -99,26 +106,39 @@ func rtohex(r rune, w int) []byte {
 func quoteDouble(s string) string {
 	var buf bytes.Buffer
 	buf.WriteByte('"')
-	for _, r := range s {
+	for len(s) > 0 {
+		r, w := utf8.DecodeRuneInString(s)
+		if r == utf8.RuneError {
```
xiaq, line 111: `&& w == 1`

pkg/parse/quote.go
```diff
@@ -99,26 +106,39 @@ func rtohex(r rune, w int) []byte {
...
+			// An invalid UTF-8 sequence was seen -- encode first byte as a hex literal.
+			buf.WriteByte('\\')
+			buf.WriteByte('x')
+			buf.Write(rtohex(rune(s[0]), 2))
+			s = s[w:]
+			continue
+		}
+
+		// The common case -- a valid UTF-8 sequence was decoded so we have a rune.
 		if e, ok := doubleUnescape[r]; ok {
```
xiaq, line 121: Make this `else if`, and it's unnecessary to duplicate the `s = s[:w]` part in the `RuneError` branch.

pkg/parse/quote.go
```diff
@@ -99,26 +106,39 @@ func rtohex(r rune, w int) []byte {
...
-			// Takes care of " and \ as well.
+			// Handle " and \ as well as things like \n.
 			buf.WriteByte('\\')
 			buf.WriteRune(e)
-		} else if !unicode.IsPrint(r) {
+		} else if unicode.IsPrint(r) {
+			buf.WriteRune(r)
+		} else if r <= ' ' {
```
xiaq, line 127: `r <= 0x7F` will take care of DEL too.

pkg/eval/compile_value_test.go
xiaq: The code these test cases exercise are already covered by unit tests in the `parse` package. Not sure which regressions you are intending to guard against.

I am not against having more test cases, but test cases need to have clear intentions on which part of the code they are supposed to guard.

pkg/eval/value_test.go
```diff
@@ -16,6 +16,7 @@ var reprTests = []struct {
 	{"a\nb", `"a\nb"`},
 	{"foo bar", "'foo bar'"},
 	{"a\x00b", `"a\x00b"`},
+	{"a\033b", `"a\eb"`},
```
xiaq, line 19: Not sure what this additional test case is supposed to do either.

pkg/parse/quote.go
xiaq: OK, reading that issue made me understand why it's desirable to use `utf8.DecodeRuneInString` to correctly handle strings with invalid UTF-8 sequences. It is to distinguish invalid sequences from actual occurrences of U+FFFD. When an actual U+FFFD is encountered, the second return value will be 3 instead of 1.

However, the code you wrote did not check this condition, meaning that it will still fail to distinguish an invalid sequence from an actual U+FFFD, and there really is no benefit over a `range` loop if you aren't checking that.

I'll fix that and add a test case.

There may be some merit in avoiding `range` loop over strings altogether and stick to `utf8.DecodeRuneInString` when iterating strings just for consistency's sake, but I find it too cumbersome to do so.

I'll revert the places where it is not necessary to use `utf8.DecodeRuneString` to the previous `range` loop code.

pkg/eval/value_test.go
krader1961: It reflects the fact that in the past the `\e` sequence wasn't recognized. I complained that even though Go doesn't recognize it the `\e` sequence is useful in programs written in Elvish and POSIX shells. Somewhere along the line you introduced support for `\e` but didn't document it or, AFAICT, include a test that it is emitted when encoding an Esc char in a string. This test addresses the second point.

PR #1537, opened 2022-05-11
Fix undefined temporary environment restoring issue

### LINKED ISSUE #1448: Unset environment variable after temporary assignment
After temporary assignment to an environment variable, that variable should be unset after the function exit, _provided_ it was unset before. Consider this (using the new `tmp` syntax):

```
⬥ date; has-env TZ; { tmp E:TZ = UTC; date }; date; has-env TZ; put $E:TZ
Sat Dec 11 10:07:40 CET 2021
⮕ $false
Sat Dec 11 09:07:40 UTC 2021
Sat Dec 11 09:07:40 UTC 2021
⮕ $true
⮕ ''
```

As can be seen from the various `date` outputs, it does make a difference whether an environment variable is unset or merely empty.

### LINKED ISSUE #1536: Setting temporary environment variable will leave a blank environment behind
```sh
~> env | grep TEST
Exception: grep exited with 1
[tty 120], line 1: env | grep TEST
~> { tmp E:TEST = 1; put $E:TEST }
▶ 1
~> env | grep TEST
TEST=
```

tw4452852:
Close #1448

### DIFF
--- a/pkg/eval/builtin_special_test.go
+++ b/pkg/eval/builtin_special_test.go
@@ -147,6 +147,8 @@ func TestTmp(t *testing.T) {
 
 		That("var x; tmp x = y").DoesNotCompile(),
 		That("{ tmp x = y }").DoesNotCompile(),
+		That("{ tmp E:X = y; put $E:X }; has-env X; put $E:X").
+			Puts("y", false, nil),
 	)
 }
 
--- a/pkg/eval/vars/env.go
+++ b/pkg/eval/vars/env.go
@@ -12,6 +12,10 @@ type envVariable struct {
 }
 
 func (ev envVariable) Set(val any) error {
+	if val == nil {
+		return os.Unsetenv(ev.name)
+	}
+
 	if s, ok := val.(string); ok {
 		os.Setenv(ev.name, s)
 		return nil
@@ -20,7 +24,11 @@ func (ev envVariable) Set(val any) error {
 }
 
 func (ev envVariable) Get() any {
-	return os.Getenv(ev.name)
+	if v, exist := os.LookupEnv(ev.name); exist {
+		return v
+	}
+
+	return nil
 }
 
 // FromEnv returns a Var corresponding to the named environment variable.
--- a/pkg/eval/vars/env_test.go
+++ b/pkg/eval/vars/env_test.go
@@ -8,6 +8,11 @@ import (
 func TestFromEnv(t *testing.T) {
 	name := "elvish_test"
 	v := FromEnv(name)
+
+	if v.Get() != nil {
+		t.Errorf("get non-exist envVariable doesn't return nil")
+	}
+
 	os.Setenv(name, "foo")
 	if v.Get() != "foo" {
 		t.Errorf("envVariable.Get doesn't return env value")
@@ -21,5 +26,13 @@ func TestFromEnv(t *testing.T) {
 	err = v.Set(true)
 	if err != errEnvMustBeString {
 		t.Errorf("envVariable.Set to a non-string value didn't return an error")
+	}
+
+	err = v.Set(nil)
+	if err != nil {
+		t.Errorf("envVariable.Set(nil) failed")
+	}
+	if _, exist := os.LookupEnv(name); exist {
+		t.Errorf("envVariable.Set(nil) doesn't unset env value")
 	}
 }


tw4452852:
The reason for the failure below
```
--- FAIL: TestEPM (0.03s)
    --- FAIL: TestEPM/use_epm (0.03s)
        epm_test.go:13: unexpected exception
        epm_test.go:13: got: vals.cannotConcat: cannot concatenate nil and string
        epm_test.go:13: stack trace: []string{"$E:XDG_DATA_HOME/elvish/lib", "use epm"}
        epm_test.go:13: want: <nil>
```

is that `$E:XDG_DATA_HOME` doesn't exist, It returned `""` before, but now `nil`.

This is what I want to ask: do we need to make this compatible or introduce a breaking change when retrieving a non-exist environment variable?

krader1961:
> This is what I want to ask: do we need to make this compatible or introduce a breaking change when retrieving a non-exist environment variable?

I can't speak for the project owner, @xiaq, but I believe the current behavior is incorrect and your change introduces correct behavior. At least in so far as expanding a non-existing environment variable should be a fatal error. The current behavior (expanding to an empty string) is a source of security bugs, or at least surprising behavior. The current behavior is probably inherited from POSIX shells where expanding a variable that doesn't exist produces an empty string rather than an error unless the "`-u`" (or "no_unset") option is enabled.

Your change identifies a bug: the bundled `epm` module assumes that `$E:XDG_DATA_HOME` exists. That module should instead use `has-env XDG_DATA_HOME` rather than `not-eq $E:XDG_DATA_HOME ''`.

tw4452852:
Agree with you @krader1961 . Mind I opening a PR to fix it?

krader1961:
> Thanks for the review, I've added evaler test as well.

Fantastic. LGTM. Note that the project owner is sometimes slow to respond to third-party pull-requests, and I believe is currently dealing with personal issues that will add additional delay, so be patient.

krader1961:
I'm going to have to retract my earlier LGTM. That's because this change causes expansion of a non-existent env var to return `$nil` rather than an empty string. I believe it should instead raise the same exception that `get-env` does when given a non-existent env var name. In other words, the two cases should behave identically and both should raise an exception. The current behavior for `$E:` expansion of non-existent env vars substituting an empty string is a source of unexpected behavior (aka bugs). Replacing the empty string with `$nil` as this P.R. does is an improvement but inconsistent with `get-env` and how every other undefined var dereference is handled.

xiaq:
Thanks for the PR, it does fix the bug, but I don't really like the idea of getting a `$nil` for non-existent environment variables. It should either evaluate to more reasonable useful default value (the current behavior) or throw an exception (what @krader1961 proposed).

Moreover, this will break code that tests whether an environment variable is unset or empty silently; `eq $E:VAR ''` currently outputs `$true` if `$E:VAR` is unset, but with the change in this PR it will output `$false` instead.

Getting back to whether `E:NONEXISTENT_VAR` should throw an exception, I am in favor of the current behavior of evaluating to an empty string. It's true this perpetuates a sloppy behavior inherited from earlier shells. But this sloppy behavior is now ingrained in the ecosystem - most scripts and programs that handle environment variables do not distinguish unset vs empty [1]. So it makes more sense for Elvish to play along than to try to fix the problem. There is always `get-env` anyway if one wishes to make that distinction.

[1] If someone has evidence that shows otherwise I'd be happy to be proven wrong, and consider making `$E:NONEXISTENT_VAR` throw an exception instead.

krader1961:
> most scripts and programs that handle environment variables do not distinguish unset vs empty

That's certainly true for POSIX scripts but is it true for Elvish scripts? And to the extent that Elvish scripts rely on the current behavior I would argue that is only due to the different behavior between `$E:NONEXISTENT_VAR` and `get-env NONEXISTENT_VAR`. If both forms threw an exception we would, obviously, see every Elvish script use `has-env` if they wanted to protect against an env var not being set. Having said all that I'm inclined to agree the `$E:...` form should retain its current behavior of expanding to an empty string if not set but that behavior should be documented (it is for `get-env` but not `$E:...`). I'll open a documentation P.R. in a few minutes.

@tw4452852, Hopefully you'll rework your fix to retain returning an empty string for an unset env var and open a new P.R.

tw4452852:
Hi @xiaq 
Yes, I agree with you about the current decision on `$E:NONEXIST` due to the compatibility. Maybe I will work out another approach to fix the original issue. 

@krader1961 
Of couse.

PR #1539, opened 2022-05-13
use has-env to detect E:XDG_DATA_HOME properly

tw4452852:

### DIFF
--- a/pkg/mods/epm/epm.elv
+++ b/pkg/mods/epm/epm.elv
@@ -31,7 +31,7 @@ var -default-domain-config = [
 var managed-dir = (
   if $platform:is-windows {
     put $E:LocalAppData/elvish/lib
-  } elif (not-eq $E:XDG_DATA_HOME '') {
+  } elif (has-env XDG_DATA_HOME) {
     put $E:XDG_DATA_HOME/elvish/lib
   } else {
     put ~/.local/share/elvish/lib


krader1961:
LGTM. See my [comment](https://github.com/elves/elvish/pull/1537#issuecomment-1124505434) on a related pull-request.

xiaq:
The existing code is working as intended; the [spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) explicitly documents that XDG_DATA_HOME being empty and non-existent are to be treated as the same:

> $XDG_DATA_HOME defines the base directory relative to which user-specific data files should be stored. If $XDG_DATA_HOME is either not set or empty, a default equal to $HOME/.local/share should be used.

krader1961:
> the [spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) explicitly documents that XDG_DATA_HOME being empty and non-existent are to be treated as the same

I would argue the spec shouldn't be "protecting" users from doing stupid things like exporting an XDG env var with an empty string as its value. But since the spec dictates that behavior we should, of course, be compliant. However, note that the pending P.R. that fixes the bug where an env var modified using `tmp` is incorrectly reverted to an empty string if the var didn't already exist was the trigger for this P.R. The question is whether we want to

a) require people to write conditions like `if (and (has-env XDG_DATA_HOME) (eq $E: XDG_DATA_HOME ""))`, or

b) continue to expand non-existing env vars to an empty string (something that is a notorious source of bugs)?

tw4452852:
Actually, I am just curious about the scenario in which the user exports `XDG_DATA_HOME` with an empty string...

krader1961:
> Actually, I am just curious about the scenario in which the user exports XDG_DATA_HOME with an empty string

It's something that happens far too often using POSIX compatible shells (e.g., `bash` and `ksh`) due to the many problems with the POSIX shell standard. The POSIX shell standard pretty much guarantees someone will inadvertently export an env var with an empty string as its value. Which is why standards like the XDG spec explicitly state that the empty string should be treated as if the env var didn't exist.

tw4452852:
Make sense, thanks @krader1961 .

xiaq:
See my comment in https://github.com/elves/elvish/pull/1537#issuecomment-1140482929.

PR #1548, opened 2022-05-25
improve consistency and readability by eliminating unnecessary dot imports 

krader1961:
This is based on my earlier P.R. to eliminate unnecessary dot imports and improve consistency. It does not change the dot imports of packages being tested (even if the dot import isn't strictly necessary).

### DIFF
--- a/pkg/cli/modes/mode_test.go
+++ b/pkg/cli/modes/mode_test.go
@@ -11,6 +11,8 @@ import (
 	"src.elv.sh/pkg/ui"
 )
 
+var Args = tt.Args
+
 func TestModeLine(t *testing.T) {
 	testModeLine(t, tt.Fn("Line", modeLine))
 }
@@ -22,9 +24,9 @@ func TestModePrompt(t *testing.T) {
 
 func testModeLine(t *testing.T, fn *tt.FnToTest) {
 	tt.Test(t, fn, tt.Table{
-		tt.Args("TEST", false).Rets(
+		Args("TEST", false).Rets(
 			ui.T("TEST", ui.Bold, ui.FgWhite, ui.BgMagenta)),
-		tt.Args("TEST", true).Rets(
+		Args("TEST", true).Rets(
 			ui.Concat(
 				ui.T("TEST", ui.Bold, ui.FgWhite, ui.BgMagenta),
 				ui.T(" "))),
--- a/pkg/cli/term/event.go
+++ b/pkg/cli/term/event.go
@@ -1,6 +1,8 @@
 package term
 
-import "src.elv.sh/pkg/ui"
+import (
+	"src.elv.sh/pkg/ui"
+)
 
 // Event represents an event that can be read from the terminal.
 type Event interface {
--- a/pkg/cli/tk/codearea_test.go
+++ b/pkg/cli/tk/codearea_test.go
@@ -10,6 +10,8 @@ import (
 	"src.elv.sh/pkg/ui"
 )
 
+var Args = tt.Args
+
 var bb = term.NewBufferBuilder
 
 func p(t ui.Text) func() ui.Text { return func() ui.Text { return t } }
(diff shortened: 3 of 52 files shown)


xiaq:
This still changes a lot of tests that import the package under test (for example the tests in `pkg/cli`).

When importing `tt`, please also alias `tt.Args` locally with `var Args = tt.Args` as it is used a lot.

krader1961:
> This still changes a lot of tests that import the package under test (for example the tests in pkg/cli).

Hmmm, the `src.elv.sh/pkg/cli` example is interesting. The dot imports of that package in *pkg/cli/app_test.go* and *pkg/cli/tty_unix_test.go* aren't actually needed and atypical compared to every other import of that package. But I'll grant you that those two unit test modules are technically testing pkg/cli and therefore a dot import is justified even if not necessary. I'll revert that particular commit.

> When importing tt, please also alias tt.Args locally with var Args = tt.Args as it is used a lot.

Yes, it is used a lot. In fact, `grep 'tt.Args' **.go | wc -l` returns 215 lines as I type this on the Elvish master branch. Which is why I felt free to use the explicit, unaliased, form in my commits. So I assume you want me to introduce aliases for those existing uses of `tt.Args`?

krader1961:
> When importing tt, please also alias tt.Args locally with var Args = tt.Args as it is used a lot.

Consider *pkg/cli/term/reader_test.go* and *pkg/cli/term/reader_windows_test.go*. The latter uses the `Args` alias created by the former package. That is non-obvious and extremely confusing if not using an IDE. This type of aliasing is less confusing if it is an idiom used everywhere in the code. However, that wasn't true before my pull-request and I would argue is an anti-pattern. I think that needless aliasing of `tt.Args` should be eliminated. Eliminating three characters from the use of a function name seems like a weak argument in this case.

krader1961:
> When importing tt, please also alias tt.Args locally with var Args = tt.Args as it is used a lot.

Regarding `tt.Args` versus `Args` as an alias consider the vast number of existing uses of the former. Such as in *pkg/cli/tk/codearea_test.go* and *pkg/cli/tk/colview_test.go*  and *pkg/edit/builtins_test.go*.

krader1961:
And, unsurprisingly, there is one test module that uses unaliased `tt.Args`: *pkg/eval/closure_test.go*. Its dot import of evaltest makes it impossible to insert `var Args = tt.Args` in that module because that creates a conflict with the declaration in *pkg/eval/options_test.go*. Yet, for reasons I don't understand *closure_test.go* can't use that alias declaration even though the compiler complains the declaration already exists if we attempt to create it:

```
pkg/eval/closure_test.go:14:2: Args already declared through dot-import of package eval ("src.elv.sh/pkg/eval")
        pkg/eval/options_test.go:10:5: other declaration of Args
```

I haven't figured out what is going on but this sort of thing reinforces my antipathy to dot imports except where absolutely needed.

xiaq:
The reason I insist on aliasing `tt.Args` as `Args` is because a test using the `tt` package will call it many times, whereas the other symbols (`tt.Test`, `tt.Fn`, `tt.Table`) will only be used once per table. You'd be right to call this inconsistent, but as I've stated elsewhere, I try very hard to keep test cases easy to write, and that sometimes means favoring writability over readability. The consistent way to do this aliasing is to always put the `var Args = tt.Args` in a `testutils_test.go` file.

I'm not a fan of the idea of import-aliasing `evaltest` to `et`; it can break `goimports`. This codebase doesn't use import aliasing anywhere and I intend to keep it this way.

krader1961:
> The consistent way to do this aliasing is to always put the var Args = tt.Args in a testutils_test.go file.

Is there some reason that change hasn't been made? I am also not convinced by the "keep test cases easy to write" argument. In my experience adding test cases tend to be copy/paste affairs and readability is more important. I can type `tt.Args` (or copy/paste it) just as fast as `Args`. The former makes it crystal clear which function is being used. Whereas the latter is ambiguous; especially in light of the `var Args = tt.Args` declaration that can appear in a non-obvious related package file.

> I'm not a fan of the idea of import-aliasing evaltest to et; it can break goimports.

If import aliasing breaks `goimports` doesn't that imply that tool has a significant bug? Why should that bug limit our use of an otherwise useful mechanism? Especially in light our extensive use of dot imports which are also aliases?

xiaq:
Aliasing `Args` is the style I prefer; obviously I can't convince you and obviously you don't have to agree with me.

`goimports` is heuristics anyway.

PR #1550, opened 2022-05-30
Improve $E:... documentation

krader1961:
Recent discussions around the fix for b.elv.sh/1448 caused me to notice
that the documentation for $E:... expansion does not explicitly explain
the current behavior. This addresses that shortcoming.

### DIFF
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -672,10 +672,11 @@ or part of an expression.
 
 **Note**: In most other shells, variables can map directly to environmental
 variables: `$PATH` is the same as the `PATH` environment variable. This is not
-the case in Elvish. Instead, environment variables are put in a dedicated `E:`
-namespace; the environment variable `PATH` is known as `$E:PATH`. The `$PATH`
-variable, on the other hand, does not exist initially, and if you have defined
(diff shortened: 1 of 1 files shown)


website/ref/language.md
```diff
@@ -2358,10 +2359,16 @@ The following namespaces have special meanings to the language:
...
-    environment variable `USER`.
-
-    This **is** always needed, because unlike command resolution, variable
-    resolution does not fall back onto environment variables.
+    environment variable `USER`. If the environment variable does not exist it
+    expands to an empty string. If you care that the env var is not set use
+    [`has-env`](./builtin.html#has-env) or [`get-env`](./builtin.html#get-env).
+    This behavior makes it easy to deal with the common case where you don't
+    care if the env var is unset, or set to an empty string, by doing this:
+    `eq $E:UNSET_VAR ''`.
```
xiaq, line 2367: > If you care that the env var is not set ...
>
> ... where you don't care if the env var is set

This is a bit hard to read, and it isn't quite clear what "caring" exactly entails. This would be more explicit and less ambiguous:

> If you want to distinguish between environment variables that are unset and those that are set but empty ...
> 
> ... where you don't need to distinguish unset and empty environment variables

PR #1551, opened 2022-05-30
Fix undefined temporary environment restoring issue

### LINKED ISSUE #1448: Unset environment variable after temporary assignment

### LINKED ISSUE #1536: Setting temporary environment variable will leave a blank environment behind

tw4452852:
This is the 2nd attempt to fix this issue. The previous one is #1537 

Close #1448 

### DIFF
--- a/pkg/eval/builtin_special_test.go
+++ b/pkg/eval/builtin_special_test.go
@@ -128,6 +128,10 @@ func TestSet(t *testing.T) {
 
 		// = is required.
 		That("var x; set x").DoesNotCompile(),
+
+		// set a non-exist environment
+		That("has-env X; set E:X = x; get-env X; unset-env X").
+			Puts(false, "x"),
 	)
 }
 
@@ -147,6 +151,11 @@ func TestTmp(t *testing.T) {
 
 		That("var x; tmp x = y").DoesNotCompile(),
 		That("{ tmp x = y }").DoesNotCompile(),
+
+		That("has-env X; { tmp E:X = y; put $E:X }; has-env X; put $E:X").
+			Puts(false, "y", false, ""),
+		That("set-env X x; { tmp E:X = y; put $E:X }; get-env X; put $E:X; unset-env X").
+			Puts("y", "x", "x"),
 	)
 }
 
--- a/pkg/eval/vars/env.go
+++ b/pkg/eval/vars/env.go
@@ -7,11 +7,11 @@ import (
 
 var errEnvMustBeString = errors.New("environment variable can only be set string values")
 
-type envVariable struct {
+type EnvVariable struct {
 	name string
 }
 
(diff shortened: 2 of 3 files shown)


pkg/eval/builtin_special_test.go
```diff
@@ -147,6 +147,11 @@ func TestTmp(t *testing.T) {
 
 		That("var x; tmp x = y").DoesNotCompile(),
 		That("{ tmp x = y }").DoesNotCompile(),
+
+		That("{ tmp E:X = y; put $E:X }; has-env X").
+			Puts("y", false),
```
krader1961, line 152: AFAICT there is no test that expanding an undefined env var returns an empty string. This is a good time to add such a test and this test seems like a natural place to do it. Simple add `; put $E:X` to the end of the test and `""` as the third expected value.

krader1961, line 152: Also, there doesn't appear to be an equivalent test of `set` so this is a good time to add one. For example:

```
That("{ set E:X = y; put $E:X }; has-env X; put $E:X").
    Puts("y", true, "y"),
```

pkg/eval/vars/env.go
```diff
@@ -7,23 +7,32 @@ import (
...
-func (ev envVariable) Get() any {
+func (ev EnvVariable) Get() any {
 	return os.Getenv(ev.name)
 }
 
+func (ev EnvVariable) Unset() error {
+	return os.Unsetenv(ev.name)
+}
+
+func (ev EnvVariable) IsExist() bool {
```
krader1961, line 30: This method name is awkward. Maybe change it to simply `Exists` or `IsSet`.

krader1961:
LGTM from a behavior perspective, and modulo my comments about improving testing of this behavior and the naming of one new method.

pkg/eval/builtin_special_test.go
tw4452852: I think `TestSet` is a better place, so I add this case there.

xiaq (approved):

PR #1555, opened 2022-06-09
Unless

lispnik:
I liked `unless` from Common Lisp and Perl and thought it might be handy in Elvish. Of course it's redundant because `if (not <cond> ...` is the same thing, but I think it reads better and is less cumbersome to look at and type (which is probably why it's in Lisp and Perl).

I also have the beginnings of a `when` branch, but I'll wait on feedback for doing anymore with that

### DIFF
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -1984,8 +1984,8 @@ unless <condition> {
 }
 ```
 
-The `unless` special command evaluates the condition to a booleanly true 
-value and skips its body.
+The `unless` special command evaluates the condition to a booleanly true value
+and skips its body.
 
 The condition part is an expression, not a command like in other shells.
 Example:
@@ -2028,10 +2028,10 @@ unless ?(test -d ~/.asdf) {
 However, for Elvish's builtin predicates that output values instead of throw
 exceptions, the output capture construct `()` should be used.
 
-**Note**: Similarly to the `if` command, the `unless` command itself doesn't 
-introduce a new scope. For example, `unless (var x = foo; put $x) { }` will leave
-the variable `$x` defined. However, the body block introduces a new scope because 
-it is a [lambdas](#function).
+**Note**: Similarly to the `if` command, the `unless` command itself doesn't
+introduce a new scope. For example, `unless (var x = foo; put $x) { }` will
+leave the variable `$x` defined. However, the body block introduces a new scope
+because it is a [lambdas](#function).
 
 ## Conditional loop: `while` {#while}
 
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -1981,11 +1981,14 @@ Syntax:
 ```elvish-transcript
 unless <condition> {
     <body>
+} else {
+    <else-body>
 }
 ```
 
-The `unless` special command evaluates the condition to a booleanly true value
-and skips its body.
+The `unless` special command evaluates the condition to a booleanly true value,
+skips its body and executes the else body, if the else body is present. It is
+the negation of the `if` special command except no `elif` clauses are allowed.
 
 The condition part is an expression, not a command like in other shells.
 Example:
@@ -2010,6 +2013,8 @@ evaluate to multiple values, in which case they are and'ed:
 ```elvish
 unless (put $true $false) {
     echo "will be executed"
+} else {
+    echo "will not be executed"
 }
 ```
 
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -1972,6 +1972,67 @@ exceptions, the output capture construct `()` should be used.
 `if (var x = foo; put $x) { }` will leave the variable `$x` defined. However,
 the body blocks introduce new scopes because they are [lambdas](#function).
 
+## Condition: `unless` {#unless}
+
+**TODO**: Document the syntax notation, and add more examples.
+
+Syntax:
+
+```elvish-transcript
+unless <condition> {
+    <body>
+}
+```
+
+The `unless` special command evaluates the condition to a booleanly true 
+value and skips its body.
+
+The condition part is an expression, not a command like in other shells.
+Example:
+
+```elvish
+unless (path:is-dir ~/.asdf) {
+    echo "no asdf install found"
+}
+```
+
+Is equivalent to:
+
+```elvish
+if (not (path:is-dir ~/.asdf) {
+   echo "no asdf install found"
+}
+```
+
+The condition part must be syntactically a single expression, but it can
+evaluate to multiple values, in which case they are and'ed:
+
+```elvish
+unless (put $true $false) {
+    echo "will be executed"
+}
+```
+
+If the expression evaluates to 0 values, it is considered true, consistent with
+how `and` works.
+
+Tip: a combination of `unless` and `?()` gives you a semantics close to other
+shells:
+
+```elvish
+unless ?(test -d ~/.asdf) {
+    mkdir ~/.asdf
+}
+```
+
+However, for Elvish's builtin predicates that output values instead of throw
+exceptions, the output capture construct `()` should be used.
+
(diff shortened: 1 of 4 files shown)


krader1961:
I agree with @xiaq that `if-not` is preferable to `unless` but am ambivalent about introducing a new syntactical keyword to avoid typing `if (not expression)`. In the Elvish source there are currently seven instances of this pattern:

```
pkg/mods/epm/epm.elv:      if (not (is-installed $pkg)) {
pkg/mods/epm/epm.elv:  if (not (is-installed $pkg)) {
pkg/mods/epm/epm.elv:    if (not (has-value $special-keys $key)) {
pkg/mods/epm/epm.elv:      if (not $silent-if-installed) {
pkg/mods/epm/epm.elv:    if (not (is-installed $pkg)) {
pkg/persistent/add-slowdown:    if (not (eq $fields[-1] ns/op)) {
pkg/persistent/add-slowdown:        if (not (has-key $native $native-name)) {
```

I have six in my personal Elvish modules:

```
lib/env.elv:if (not (has-env ELVISH_TRACE)) {
lib/env.elv:if (not (has-env SSH_AUTH_SOCK)) {
lib/login.elv:  if (not ?(ssh-add -l >/dev/null 2>&1)) {
lib/login.elv:  if (not ?(ps waux | grep -q '[p]ostfix/master')) {
lib/prompt.elv:    if (not (eq "" $E:CONDA_DEFAULT_ENV)) {
lib/util.elv:    if (not ?(test -d $dir)) {
```

I'm not convinced those few instances would be easier to read (or write) by an `unless` or `if-not` keyword. Too, it seems hard to justify a new keyword given how infrequently that pattern is used in practice. It also requires users of the language to learn the behavior of yet another keyword. On the other hand adding `if-not` does have a nice symmetry with `if` and has an obvious interpretation for people for whom English is not their native language.

I can't speak to why Lisp supports it (since I've only written trivial amounts of Lisp in the context of the Emacs editor) but I used to be a Perl guru and consider its `unless` keyword to be an example of the kitchen sink phenomenon that affects many languages such as `ksh`, `bash`, and `zsh`.

krader1961:
> I also have the beginnings of a when branch, but I'll wait on feedback for doing anymore with that

That cryptic comment and this P.R. is why I prefer to see such proposals discussed in an issue before being asked to review code. I know @xiaq is more flexible on this point but it seems premature, to me, to be asked to review a major change to the language before the merits of such a change have been discussed. In the case of a `when` keyword it's not at all obvious what you're proposing. Are you suggesting a switch/case mechanism such as described by https://perldoc.perl.org/5.14.3/functions/when? A general switch/case mechanism should have significant discussion before a particular implementation is put forth for review.

lispnik:
> I'm not sure that when will be useful for Elvish - care to elaborate?

Regarding this, I feel it was half baked and should be ignored! It was from my earlier experimentation.

lispnik:
I've pushed some changes to implement an unless that works a bit more like `if`: 

```
unless <cond> { 
    <body> 
} else { 
    <else-body> 
}
```

The `else` branch is optional, as in `if`.

This was mostly inspired by Ruby while also has an `unless` control flow statement https://ruby-doc.org/core-3.1.2/doc/syntax/control_expressions_rdoc.html#label-unless+Expression

website/ref/language.md
```diff
@@ -1981,11 +1981,14 @@ Syntax:
...
+} else {
+    <else-body>
 }
 ```
 
-The `unless` special command evaluates the condition to a booleanly true value
-and skips its body.
+The `unless` special command evaluates the condition to a booleanly true value,
+skips its body and executes the else body, if the else body is present. It is
+the negation of the `if` special command except no `elif` clauses are allowed.
```
krader1961, line 1991: Ugh! This is another reason, IMHO, why this should not be introduced. At least in its current form. I can't see any rational reason why `elif` clauses should not be allowed since `unless` is simply syntactic sugar for `if (not *expression*)`. It seems like the restriction is because the syntactic sugar is not really an obvious, or equivalent, variant of `if (not *expression*)`. Which argues against its addition, IMHO.

lispnik:
> I find if-not a slightly better name than unless since it's more apparent what the relationship with if is.

While not a linguist, I still think `unless foo` is easier than `if (not foo` and the kind-of similar `if-not foo` and kind of has a precedent in other languages (Ruby, Perl, Lisp, maybe more) so less surprise for people perhaps. 

Also, if it's really similar to `if`, people might want `elsif`s in there, like in Perl, which seems pointless. Ruby explicitly disallows `elsif` in its `unless` syntax perhaps for the reason that if you've got _several branches_ you might have well used `if` in the first place. The Ruby syntax is like the one being proposed here.

`if-not` works though and I'd be fine renaming it. Should it support `elsif` clauses?

krader1961:
> if-not works though and I'd be fine renaming it. Should it support elsif clauses?

See my previous comments. I don't speak for the project owner. I am ambivalent about a new keyword; whether `unless` or `if-not`. Personally I don't think either keyword adds enough benefit to outweigh the costs of a new language keyword. Syntactic sugar such as `unless` rather than `if (not *expression*)` requires significant benefits or justification. It's not clear that is true for this proposed change.

lispnik:
I tried implementing this myself in user code to see if I can avoid this altogether.

```
fn unless {
  # unless <cond> { body } [ else { else-body } ]
  | cond body @rest |
  if (not $cond) {
    $body
  } else {
    if (== (count $rest) 0) {
      nop
    } elif (and (== (count $rest) 2) (eq $rest[0] "else")) {
      $rest[1]
    } else {
      fail "invalid unless else syntax"
    }
  }
}
```

It gets me 80% of what I want and I can happy with that! I don't think I can implement `elsif` branches because of how eval will probably eval all the conditions before I get to conditionally eval them. I can probably put the cond and elif-conds in lambdas eval them lazy-style. It just looks a bit unlike other special commands though: `unless { <cond1> } { <body1> } elsif { cond2 } ... else { ... }`. Also, I haven't found a way to make it "parse" like other special commands – e.g.. that cool underlining of the problem Elvish does.

krader1961:
I'm always amazed at how the Elvish language makes it possible to introduce mechanisms such as your `fn unless` definition, @lispnik. It did cause me to consider another objection to your proposal to introduce an `unless` keyword. If added it argues for an `elunless` or `elif-not` keyword for symmetry (depending on whether you prefer `unless` or `if-not`). I think you can see why I'm even less of a fan of an `unless` (or `if-not`) keyword after making that observation. Again, it adds a lot of cognitive overhead in terms of the language specification in order to avoid using the obvious, and reasonably short `(not (some-expression))`. Obviously, the else forms could be named in a way that is easier to read (e.g., `else-unless`) but I don't think that materially alters my argument and conclusion.

lispnik:
I would just go with `unless <cond> { <body> } [ else { <body> } ]` and call it a day.

xiaq:
Coming back to this, I tried changing some uses of `if (not ...)` to `if-not ...` and `unless ...`, and I don't feel there is a clear advantage. So this is going to a reject, sorry.

FWIW, I feel that `unless` with `else` or `elif` is going to be quite confusing, and is better disallowed. [Raku's `unless`](https://docs.raku.org/language/control#unless) does not support `else`, and neither do [Racket's unless](https://docs.racket-lang.org/reference/when_unless.html), which is probably just inherited from [R6RS's `unless`](http://www.r6rs.org/final/html/r6rs-lib/r6rs-lib-Z-H-6.html). I do like Clojure's naming of [its `if-not`](https://clojuredocs.org/clojure.core/if-not) but I consider its support for an else clause is a mistake.

As you have observed, once you don't need support for `elif`, this command can be implemented just as a function. So you can do that yourself if you find it compelling enough; in future if the balance tilts in favor of having this functionality, it can be introduced as a builtin function too.

PR #1559, opened 2022-06-22
Expose special paths used by Elvish via $platform:

### LINKED ISSUE #1385: Feature request: command to find location of rc.elv file
Now that https://github.com/elves/elvish/issues/383 has been fixed, the location of `rc.elv` is less obvious. It would be nice if there was an Elvish command to determine the location of `rc.elv`. Other shells that vary the location of their config files based on OS or environment variables already provide a way to do this:
- PowerShell: `echo $profile`
- Nushell: `config path`

krader1961:
There are several paths used by Elvish internally that should also be visible
to Elvish scripts. This change exposes two of those paths via the `platform`
module. Thus making it easier for Elvish scripts to determine where various
files live and, more importantly, should be written. The latter is especially
important for transient state files that should be placed in the same,
or a subdirectory of, the directory used for the Elvish database file.

Resolves #1385


krader1961:
Argh! I just realized I failed to include unit tests for the new vars. Updated P.R. on the way.

krader1961:
FYI, I've added relevant unit tests and force pushed an update to this pull-request. I think it is ready for review.

xiaq:
- These paths are Elvish-specific, and I'd rather this be in a new `runtime` module rather than in `platform`. The `platform` module is supposed to be about the platform that Elvish is running on top of, not about the Elvish runtime itself. 

- The concept of state home, config home and data home are defined in XDG, and these paths do **not** contain last application-specific component. Having `$state-home` point to `~/.local/state/elvish` rather than `~/.local/state` (which is the default state home according to XDG) is confusing at best. It's better to just expose the paths Elvish actually uses, such as the database directory (`~/.local/state/elvish/db.bolt`) and the RC file (`~/.config/elvish/rc.elv`).

- The various paths used by Elvish are computed at startup and do not change afterwards. As such, the implementation should use the paths that are already computed, rather than computing them anew every time; for example, if the `XDG_STATE_HOME` env variable gets changed, the path to Elvish's database should **not** change.

  The path to the database file is extra tricky, because even the value of `XDG_STATE_HOME` at startup may not matter if Elvish is connecting to an existing daemon, in which case it is the value of `XDG_STATE_HOME` of the Elvish session that spawned the daemon that matters. Getting the path to the database file accurately requires asking the daemon, and is better put in the `store:` module instead.

Implementing the design and implementation I have in mind is unlikely to reuse much of the code in this PR, so I'll close this. The new `runtime` module has been introduced in 9794eecd6869d798f43e7d2abf13d3996bcc49d1.

krader1961:
As i've noted many times before I think the daemon should be removed as soon as possible. So I don't think there is a good reason to change the daemon code to report its current database path. Too, I also think it is preferable to have vars that reflect higher level concepts than the literal path to the *rc.elv* or *db.bolt* files. But I do agree that this information belongs in a `runtime` module rather than `platform`.

xiaq:
Having variables for high-level concepts is fine, but they shouldn't use the confusing "foo home" naming. "Data dir", "state dir" and "config dir" would be fine.

PR #1565, opened 2022-06-26
Document how to conditionally `use` a module

### LINKED ISSUE #1116: How to conditionally `use` a module interactively
I like to automatically `use` all the modules I'm likely to need in my interactive Elvish shells. For example, from my *~/.elvish/rc.elv*:

```
# Standard elvish modules we always want to be readily available.
use builtin
use epm
use math
use platform
use re
use readline-binding
use str
use unix
```

The problem is the `use unix` fails on Windows. If I instead do this then the unix namespace isn't available until I type `use unix` interactively:

```
if $platform:is-unix {
  use unix
  unix:umask = 0o027
}
```

It would be nice if I could do something like

a) `use &ignore-missing unix`, or

b) `use &up unix` to force the import to occur in the outer scope.

krader1961:
Resolves #1116

### DIFF
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -2432,6 +2432,32 @@ use a/b/c # imports the "a/b/c" module as "c:"
 use a/b/c foo # imports the "a/b/c" module as "foo:"
 ```
 
+### Conditionally importing modules with `use`
+
+You may want to import a module in an interactive session only if a specific
+condition is true. You might think the solution is straightforward and put
+something like this in your [_rc.elv_](./command.html#rc-file):
+
+```elvish
+if (a-condition) {
+    use a-module
+}
+```
+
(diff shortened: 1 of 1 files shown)


website/ref/language.md
```diff
@@ -2432,6 +2432,32 @@ use a/b/c # imports the "a/b/c" module as "c:"
 use a/b/c foo # imports the "a/b/c" module as "foo:"
 ```
 
+### Conditionally importing modules with `use`
```
xiaq, line 2435: This example already exists in https://elv.sh/ref/edit.html#edit:add-var.

It also doesn't belong to the reference doc - it's not part of the language, and the entire section is non-normative.

It's better to expand the tip in https://elv.sh/ref/edit.html#edit:add-var a bit and add a link here.

PR #1566, opened 2022-06-26
Add `platform:elvish-path`

### LINKED ISSUE #1423: Set the SHELL env variable to be the full path of the Elvish binary

krader1961:
Resolves #1423


krader1961:
In light of @xiaq's response to my solution for issue #1385, wherein they introduced a `runtime` module, I'm going to assume they would prefer this variable to exist in that namespace. Updated PR forthcoming...

pkg/mods/runtime/runtime.go
```diff
@@ -38,13 +40,32 @@ import (
...
 
+//elvdoc:var elvish-path
+//
+// The path to the elvish binary. If that could not be determined the value will be the empty
+// string.
+//
+// This is read-only.
+
+func elvishPath() any {
+	// TODO: Decide what to do if os.Executable returns an error. It appears that all platforms
```
xiaq, line 51: The pattern established here is to use `$nil` when there is any error.

PR #1567, opened 2022-06-27
Fix glob expansion that includes broken symlinks

### LINKED ISSUE #1240: Broken symlink causes wildcard matching to fail

krader1961:
Glob expansion should use os.Lstat rather than os.Stat so that broken
symlinks don't stop glob expansion. It's not the place of glob expansion
to decide if a particular path is good.

The symlink test cases and logic were borrowed from
pkg/mods/path/path_test.go.

Fixes #1240

### DIFF
--- a/0.19.0-release-notes.md
+++ b/0.19.0-release-notes.md
@@ -21,6 +21,9 @@ and compiled, even if it is not executed:
 -   Temporary assignment on an unset environment variables no longer leave it
     set to an empty string ([#1448](https://b.elv.sh/1448)).
 
+-   Glob expansion that includes broken symlinks no longer short-circuits the
+    expansion ([#1240](https://b.elv.sh/1240)).
+
 # Notable new features
 
 -   A new `inexact-num` converts its argument to an inexact number.
--- a/pkg/glob/glob_test.go
+++ b/pkg/glob/glob_test.go
@@ -20,6 +20,15 @@ var (
 		"lorem", "ipsum",
 		"d1/e/f/g/X", "d2/e/f/g/X"}
 	createDots = []string{".x", ".el/x"}
+	symlinks   = []struct {
+		path   string
+		target string
+	}{
+		{"d1/s-f", "f"},
+		{"s-d", "d2"},
+		{"s-d-f", "d1/f"},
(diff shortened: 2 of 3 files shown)


0.19.0-release-notes.md
```diff
@@ -21,6 +21,9 @@ and compiled, even if it is not executed:
 -   Temporary assignment on an unset environment variables no longer leave it
     set to an empty string ([#1448](https://b.elv.sh/1448)).
 
+-   Glob expansion that includes broken symlinks no longer short-circuits the
+    expansion ([#1240](https://b.elv.sh/1240)).
```
xiaq, line 25: This is a fix but it's also a breaking change - symlinks to directories will now be treated as files, not directories. This breaking change is fine for me - among other things it avoids infinite recursions when expanding `**` symlinks introduce loops in the FS tree. But it should be documented as such.

pkg/glob/glob_test.go
```diff
@@ -80,6 +88,15 @@ func testGlob(t *testing.T, abs bool) {
 		}
 		f.Close()
 	}
+	for _, link := range symlinks {
+		err := os.Symlink(link.target, link.path)
+		if err != nil {
+			// Creating symlinks requires a special permission on Windows. If
+			// the user doesn't have that permission, just skip the whole test.
+			t.Skip(err)
```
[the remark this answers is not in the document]
krader1961: Good point. I would argue that both sets of tests should fail if unable to create a symlink on Windows. The CI environment should always have that capability enabled. And anyone running these tests on their local system should also have the capability enabled. The question is whether, and how, to test the behavior on Windows when the symlink capability isn't enabled.

0.19.0-release-notes.md
krader1961: Another good point but the existing behavior is flat out wrong and it's unclear how to document the change in behavior.

Issue #1453, opened 2021-12-23
Add command to remove consecutive duplications (like the UNIX `uniq` command)


krader1961:
I was creating a pipeline involving `grep` to find all the mentions of `rc.elv` that weren't Go imports and noticed it would have been useful if `builtin:order` had a `&uniq` (or `&unique`) option to eliminate duplicates. Also, for that particular scenario it would have been sufficient if a `builtin:uniq` was available that could filter the value stream since I didn't really need to order (sort) the values since the duplicates would always appear adjacent to each other.

xiaq:
I don't see any substantial advantage in adding an option to `order`. Adding another command after it is simple enough - `order &uniq` isn't any shorter than `order | uniq`.

Also regarding the name - it's better called something different from `uniq` to avoid shadowing the external command. This is why the sorting command is `order`, not `sort`.

I also never liked the name `uniq` - it can mislead people into assuming that it will remove all duplications, whereas it only removes consecutive duplications. Go'e experimental slice package calls the same functionality "compact" (https://pkg.go.dev/golang.org/x/exp@v0.0.0-20220218215828-6cf2b201936e/slices#Compact). This name is a bit vague but at least it's not misleading.

krader1961:
The advantage of a "unique" option to the `order` builtin is better performance, but I'll admit that is a weak argument since it is a micro optimization given the overall performance of a shell like Elvish. Hence why I opened this with a proposal for a new command that could operate on a stream of values to eliminate identical consecutive values similar to how the POSIX `uniq` command handles lines in byte streams. I agree about using a builtin name other than `uniq`; both to avoid clashing with the external command and because the traditional name is misleading (albeit only mildly). I don't like "compact" but I don't have a better suggestion at this time that isn't excessively verbose.

Does a new command to eliminate successive duplicate values have to be in the "builtin" module? Should we consider introducing a new module for such operations that would minimize the likelihood of clashing with an external command of the same name? What would that mean for other, relatively recently introduced, commands added to the "builtin" namespace such as `order`? I'm not really a fan of this idea but think it warrants some discussion since it would minimize future clashes with external commands.

xiaq:
It's fine living in the builtin namespace for now. Most of the commands that operate on streams are in the builtin namespace.

krader1961:
[...]

PR #1568, opened 2022-06-27
Add an `order &dedupe` option

krader1961:
Related #1453

### DIFF
--- a/pkg/eval/builtin_fn_container_test.go
+++ b/pkg/eval/builtin_fn_container_test.go
@@ -350,6 +350,18 @@ func TestOrder(t *testing.T) {
 		That("put l x o x r x e x m | order &less-than={|a b| eq $a x }").
 			Puts("x", "x", "x", "x", "l", "o", "r", "e", "m"),
 
+		// &dedupe eliminates consecutive identical values
+		That("put a b c a b c | order &dedupe").Puts("a", "b", "c"),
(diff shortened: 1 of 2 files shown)


SolitudeSF:
why is that an option instead of separate command?

krader1961:
> why is that an option instead of separate command?

For performance. It's more efficient to do the dedupe when ordering the values than it is to pipe them into a separate command. I'm also working on a separate command. See issue #1453.

krader1961:
@SolitudeSF, I just created P.R. #1569 that implements a `dedupe` command. So now we can efficiently sort lists with duplicate elimination and dedupe arbitrary inputs.

xiaq:
Sorry, but no.

Optimizing `order | dedupe` is indeed desirable, eventually. But it can and should be done transparently - the compiler can recognizing the pattern `order | dedupe` in a pipeline and use a more efficient implementation under the hood.

krader1961:
> Optimizing order | dedupe is indeed desirable, eventually. But it can and should be done transparently - the compiler can recognizing the pattern order | dedupe in a pipeline and use a more efficient implementation under the hood.

Yes, but I suspect I'll be dead before that happens. 😄

It's not clear to me why an obviously trivial optimization that works today, and the immediate future, isn't acceptable. Even if the optimization you envision is implemented in the distant future I don't see any reason to not implement the much simpler optimization today since it is trivial.

PR #1569, opened 2022-06-28
Implement a `dedupe` command

### LINKED ISSUE #1453: Add command to remove consecutive duplications (like the UNIX `uniq` command)

krader1961:
Resolves #1453

### DIFF
--- a/0.19.0-release-notes.md
+++ b/0.19.0-release-notes.md
@@ -30,5 +30,8 @@ and compiled, even if it is not executed:
     behavior may change in future if there are more underlying types for inexact
     numbers.
 
+-   A new `dedupe` command to eliminate consecutive duplicate values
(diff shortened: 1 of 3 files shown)


xiaq:
This new command does very different things depending on the option, which should just be different commands.

I suppose this design is based on what the `uniq` command supports in most Unix systems. It's a bad design.

See also my comment in #1453 re the name.

PR #1573, opened 2022-07-05
A couple of minor documentation markup fixes

krader1961:


pkg/eval/builtin_fn_num.go
```diff
@@ -614,19 +614,22 @@ func div(rawNums ...vals.Num) (vals.Num, error) {
 // % $x $y
 // ```
 //
-// Output the remainder after dividing `$x` by `$y`. The result has the same
-// sign as `$x`. Both must be integers that can represented in a machine word
-// (this limit may be lifted in future).
+// Output the remainder after dividing `$x` by `$y`. The result has the same sign as `$x`. Both must
+// be integers that can be represented by a native machine word (this limit may be lifted in the
+// future). This means that integers otherwise supported by Elvish, but that are outside the range
+// of the system `int` can result in an exception.
```
xiaq, line 620: I chose the term "native machine word" because the size of `int` is not really well defined across language, so I'd rather stick to that wording here. It's also confusing that the same concept appears under two different terms here.

krader1961: Fair enough.

PR #1578, opened 2022-07-13
Add to-yaml, from-yaml builtins

### LINKED ISSUE #1155: Allow parsing YAML documents (and streams) like we do with the "from-json" function
Today it's possible to parse JSON in Elvish using [the builtin `from-json` function](https://elv.sh/ref/builtin.html#from-json). It would be helpful if we could also load YAML documents with an analogous `from-yaml` function. As people often write multiple [YAML documents](https://yaml.org/spec/1.2/spec.html#id2800132) together in a single [YAML stream](https://yaml.org/spec/1.2/spec.html#id2801681), it would also be helpful to be able to load a YAML stream that would yield any number of documents (perhaps with a function called `from-yaml-stream`).

I understand that doing so would require introducing a dependency on a YAML parsing library. Whether we should provide functions to _create_ a YAML document like [`to-json`](https://elv.sh/ref/builtin.html#to-json) is debatable. [Jsonnet's `manifestYamlDoc` function](https://jsonnet.org/ref/stdlib.html#manifestYamlDoc) in its standard library has been controversial for the way it quotes strings in mapping keys.

aca:
When writing a script to handle yaml, It's quite annoying as no language supports yaml format natively and have to install dependancies. If it's json, jq works fine, but cli tools for yaml like `yq` is quite annoying as there's quite many different versions of it. It would be quite useful if elvish can natively support various formats just like json.

fixes #1155

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/sourcegraph/jsonrpc2 v0.1.0
 	go.etcd.io/bbolt v1.3.6
 	golang.org/x/sys v0.0.0-20220227234510-4e6760a101f9
+	gopkg.in/yaml.v3 v3.0.1
 )
 
 require golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 // indirect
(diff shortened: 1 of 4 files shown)


aca:
test for `thatOutputErrorIsBubbled` is commented out as I just can't get an easy fix. If implementation is accepted, I wish I can get some help.

xiaq:
This is a fine feature by itself. I'm not a fan of YAML, but supporting YAML in Elvish is a good idea since YAML is already used in quite a few places.

However, the additional dependency is quite chunky and adds 516,016 bytes to the Elvish binary in my local environment (darwin/amd64). Elvish's binary is already pretty large and I'm not quite keen to take on more dependency at the moment. So sorry, I'm going to close this PR.

Right now what I can offer as a workaround is to use a command to convert YAML from and to JSON in combination of `from-json` and `to-json` to implement `from-yaml` and `to-yaml`:

```elvish
fn from-yaml { yj | from-json }
fn to-yaml { to-json | yj -r }
```

The main disadvantage is that the user needs to install `yj` themself, and there is no uniform way to do that across the platforms Elvish supports. For that I've written down some ideas in #1607.

PR #1581, opened 2022-07-14
Remove redundant math: var usage text

### LINKED ISSUE #1580: Variables in the `builtin:` namespace do not include a usage while those in other namespaces do
Compare https://elv.sh/ref/builtin.html#after-chdir to https://elv.sh/ref/math.html#math:e. I noticed this inconsistency when implementing a `help` command. Personally, I don't think there is any value in a "usage" block for variables and all of the non-builtin modules should omit such blocks since the redundant usage blocks simply repeat that the variable is dereferenced by a dollar-sign prefix. Thus, they don't add any information. Alternatively, the `builtin:` variable definitions should include a "usage" block for consistency with the variables in the non-`builtin:` modules.

krader1961:
Resolves #1580

### DIFF
--- a/pkg/mods/math/math.go
+++ b/pkg/mods/math/math.go
@@ -50,19 +50,11 @@ var Ns = eval.BuildNsNamed("math").
 
 //elvdoc:var e
 //
-// ```elvish
-// $math:e
-// ```
-//
 // Approximate value of
 // [`e`](https://en.wikipedia.org/wiki/E_(mathematical_constant)):
 // 2.718281.... This variable is read-only.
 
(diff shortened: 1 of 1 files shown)


krader1961:
This change passes on my local `FreeBSD` VM. It's also not clear why it failed on the `FreeBSD` CI environment since the "details" link doesn't provide any text that explains why it failed. Since this change only removes eight lines of documentation that is not tested it appears to be a spurious CI failure.

xiaq:
The FreeBSD CI runner on CirrusCI is somewhat more flaky than GitHub CI.

Issue #882, opened 2019-12-28
Support generating API doc from Elvish sources


xiaq:
We have an `elvdoc` tool that can extract comments like the following from Go sources:

```go
//elvdoc:fn name-of-function
// Doc body

//elvdoc:var name-of-variable
// Doc body
```

And it is being used to generate parts of https://elv.sh/ref/edit.html. However, we do not have such a mechanism for Elvish sources yet, and as a result https://elv.sh/ref/epm.html is still written "by hand".

A similar tool that extracts doc comments from .elv files can be useful. Different from Go, since in .elv files you are always defining an Elvish API, we don't need the explicit `//elvdoc:` style comment; something similar to Go doc suffices. For example, in the following code:

```elvish
# Does something
fn f { }
```

The tool can extract the comment above `fn f` as the API doc for the function.

PR #1583, opened 2022-07-19
Implement a `help` command

### LINKED ISSUE #1432: A help system for the built-in commands
As rasied on the user group, I asked about the possibility of having the documentation for each command available via the command line, so we could type e.g. `> help randint` and get the documentation that is already present in the go source code, e.g. https://github.com/elves/elvish/blob/master/pkg/eval/builtin_fn_num.go#L576 

Currently this documentation is in markdown. The simplest solution would be to convert from markdown to plain text, resulting in something like (for this I used `pandoc -t plain`):

```
elvdoc:fn randint

    randint $low $high

Output a pseudo-random integer in the half-open interval [$low, $high).
Example:

    ~> # Emulate dice
    randint 1 7
```

This would already be a big advance. Of course for cherries on top it would be nice if links could be preserved and code could get syntax highlighting. I do not know anything about go and how feasible this would be, I assume there are many markdown packages available...

krader1961:
Resolves #1432


xiaq:
This is a fine implementation by itself, but there are two flaws based on which I'll reject this PR:

- This introduces a build-time dependency on `glow` for building Elvish, without which Elvish won't build at all. Go's build system does not support build hooks, so this requires a custom build script, as your Makefile change shows.

  This means that every distro packager will need to use the custom build script, instead of using their standard recipe for building Go binaries. But even worse, this breaks the usability of the Elvish repo as a Go module, so no other Go module can depend on `src.elv.sh`. I'd rather keep Elvish a good citizen of the Go ecosystem unless absolutely necessary.

  This problem can be partially mitigated by having a default `pkg/help/documentation.go` that declares the required variables initialized to empty maps, so that Elvish will still build, albeit without the builtin documentation, if the custom build script is not used. With this mitigation, this flaw is less critical, but still very much not ideal - the distro packagers will still need to use the custom build script, and there is also no way for other Go modules to depend on a version of Elvish with the builtin documentation.

- Another, arguably bigger, issue is that the implementation is hard to extend to support documentation found in Elvish sources (#882), barring the builtin modules implemented in Elvish (`epm` and `readline-binding`), since the entire rendering process happens during compilation.

  It is probably possible to adapt the elvdoc rendering process to a script that the user can run against their Elvish sources, but that will require the user to install the `glow` command. Installing a Go dependency is not too hard; but if documentation is really important enough to be made accessible directly from the Elvish binary, it's hard to justify the sudden jump in friction when going from viewing documentation of builtin modules to viewing documentation of user-defined modules.

Taking a step back, instead of rendering the elvdoc in builtin modules and generalizing that to Elvish sources later (the first step of which is evidently what this PR is trying to do), I'd rather solve it the other way around:

1. Support rendering elvdoc in any Elvish source (#882);

2. Generalize that to builtin modules implemented in Go.

To be able to do 1, the hard part is teaching Elvish to render Markdown in the terminal. This PR sidesteps that problem by using `glow` during compilation, but if Elvish needs to do that on its own, it needs to have a Markdown parser. The obvious choice is to use an existing Markdown parser like https://github.com/yuin/goldmark, but it's a very heavy dependency (including the parser alone adds more than 1MB to the binary). So the alternative to write a new Markdown parser tailored to the requirements of Elvish.

Once 1 is done, step 2 can be done by moving the elvdoc of modules implemented in Go into `.elv` files. For example,

```go
//elvdoc:fn some-fn
// Some function
func SomeFunc(x string, y ...string) { }
```

becomes

```elv
# Some function
fn some-func {|x @y| }
```

The implementation will still live in the Go files; the `.elv` file will just provide the elvdoc and the function signature, functioning as a sort of "header" file, and they can be embedded into the Elvish binary with `//go:embed`. The function signatures in those `.elv` files can also be used, for example, for providing code completions in future.

The hardest part of the plan is to parse Markdown, so I tried to build one against the [CommonMark spec](https://spec.commonmark.org/), and the https://github.com/elves/elvish/tree/builtin-elvdoc branch now has a parser that implements a substantial subset of the spec. Now that this plan seems quite feasible, I'll close this PR.

PR #1584, opened 2022-07-19
Documentation fixups

krader1961:
I noticed these functions didn't have a "usage" while working on a `help` command.


pkg/edit/builtins.go
```diff
@@ -17,16 +17,24 @@ import (
...
+// edit:-dump-buf
+// ```
+//
 // Dumps the current UI buffer as HTML. This command is used to generate
 // "ttyshots" on the [website](https://elv.sh).
 //
 // Example:
 //
-// ```elvish
+// ```elvish-transcript
```
xiaq, line 37: This is not a transcript; it probably should be changed to be one, but then you'd need to add prompts to the content.

xiaq: Actually, I'll just remove this command altogether now that your `ttyshot` change has been integrated.

PR #1587, opened 2022-07-24
vi append command binding

aca:
This adds https://vimhelp.org/insert.txt.html#a like key binding in command:binding.
It can be just tested with this on rc.elv. But I tought it would be better to make it as a default binding.

```
set edit:command:binding[a] = { edit:move-dot-right; edit:close-mode }
```

### DIFF
--- a/pkg/edit/init.elv
+++ b/pkg/edit/init.elv
@@ -57,6 +57,7 @@ set command:binding = (binding-table [
   &b=   $move-dot-left-word~
   &h=   $move-dot-left~
   &i=   $close-mode~
+  &a=   { $move-dot-right~; $close-mode~ }
   &j=   $move-dot-down~
   &k=   $move-dot-up~
   &l=   $move-dot-right~


krader1961:
More context in the commit comment would be helpful. What does this do? What problem does it solve? Why isn't there a unit test?

aca:
I added some context in PR. There's already test for `move-dot-right` and `close-mode`. And I didn't thought there this requires additional test. Correct me if I'm wrong.

krader1961:
> There's already test for move-dot-right and close-mode. And I didn't thought there this requires additional test. Correct me if I'm wrong.

You're introducing a definition for a new key binding. The fact that your implementation depends on the behavior of existing functions is not relevant.  A unit test is intended to verify the observable behavior regardless of the implementation. In other words, if your implementation used something other than `edit:move-dot-right; edit:close-mode` it should still pass the (missing) unit test. A unit test should not assume anything about the implementation other than documented by its API.

krader1961:
@aca You wrote "I added some context in PR." However, this pull-request wasn't updated. Did you neglect to do a `git push -f` Also, your change still needs to include an explicit unit test of the new key binding rather than assuming it works given that it is based on existing behaviors.

xiaq:
This is fine - the vi mode is kind of "experimental" anyway and the other keybindings in it aren't tested either.

PR #1588, opened 2022-07-25
Support comparing booleans

### LINKED ISSUE #1585: boolean values are not comparable
While working of #1570 I tried this to verify that a constant key results in no change to the order of the inputs:

```
> order [1 3 2 5 4] &key={|a| put $true }
Exception: bad value: inputs to "compare" or "order" must be comparable values, but is uncomparable values
```

I thought I had a bug in my `&key` implementation but it turns out that the current Elvish implementation does not support comparing booleans:

```
> order [$true $false]
Exception: bad value: inputs to "compare" or "order" must be comparable values, but is uncomparable values
```

Yes, this is an odd thing to do but it should be possible to compare booleans. Note that this isn't a problem with the `order` command -- it's a problem with the `compare` command (whose implementation `order` uses for comparisons):

```
> compare $true $false
Exception: bad value: inputs to "compare" or "order" must be comparable values, but is uncomparable values
[tty 25], line 1: compare $true $false
```

krader1961:
Fixes #1585


pkg/eval/builtin_fn_pred.go
```diff
@@ -305,6 +305,17 @@ func cmp(a, b any) ordering {
 				return more
 			}
 		}
+	case bool:
+		if b, ok := b.(bool); ok {
+			switch {
+			case a == b:
+				return equal
+			case !a: // `&& b` is implicit
```
xiaq, line 313: I find this to be a rare case where `a == false` is preferable to `!a`, as `a` is being tested for its value rather than being used as a condition.

0.19.0-release-notes.md
```diff
@@ -32,3 +32,7 @@ and compiled, even if it is not executed:
 
 -   A new type of interactive abbreviation: `edit:command-abbr`
     ([#1472](https://b.elv.sh/1472)).
+
+-   The `compare` command now supports comparing booleans
+    ([#1585](https://b.elv.sh/1585)). This also means that the `order` command
+    can now sort booleans.
```
xiaq, line 38: Better to just say that both `compare` and `order` both support boolean values now.

PR #1591, opened 2022-07-30
Add new benchmark command

### LINKED ISSUE #1586: Add a `&benchmark` option to the `time` command
While working on issue #1570 I wrote a simple script that used `time` to test the speed of various sorting options. It would be extremely helpful if the `time` command had a `&benchmark` option that would run the function *n* times (perhaps defaulting to three or five but allowing the user to specify the number of runs) and outputting the best time.

krader1961:
Resolves #1586

### DIFF
--- a/pkg/eval/options.go
+++ b/pkg/eval/options.go
@@ -37,6 +37,16 @@ func scanOptions(rawOpts RawOptions, ptr any) error {
 		if !ok {
 			return UnknownOption{k}
 		}
+
+		// An option with no value (e.g., `&a-opt`) has `$true` as its default value. However, if
+		// the option struct member is a string we want an empty string as the default value.
+		switch b := v.(type) {
+		case bool:
+			if b && structValue.Field(fieldIdx).Type().Name() == "string" {
+				v = ""
+			}
+		}
+
 		err := vals.ScanToGo(v, structValue.Field(fieldIdx).Addr().Interface())
 		if err != nil {
 			return err
--- a/0.19.0-release-notes.md
+++ b/0.19.0-release-notes.md
@@ -56,3 +56,5 @@ and compiled, even if it is not executed:
 
 -   A new `compact` command that replaces consecutive runs of equal values with
     a single copy, similar to the Unix `uniq` command.
+
+-   A new `benchmark` command has been added ([#1586](https://b.elv.sh/1586)).
--- a/pkg/eval/builtin_fn_misc.go
+++ b/pkg/eval/builtin_fn_misc.go
@@ -16,8 +16,10 @@ import (
 )
 
 var (
-	ErrNegativeSleepDuration = errors.New("sleep duration must be >= zero")
-	ErrInvalidSleepDuration  = errors.New("invalid sleep duration")
+	ErrNegativeSleepDuration    = errors.New("sleep duration must be >= zero")
+	ErrInvalidSleepDuration     = errors.New("invalid sleep duration")
+	ErrInvalidBenchmarkDuration = errors.New("benchmark duration must be > zero")
+	ErrInvalidBenchmarkIter     = errors.New("benchmark iterations must be > zero")
 )
 
 // Builtins that have not been put into their own groups go here.
@@ -36,8 +38,9 @@ func init() {
 		"deprecate": deprecate,
 
 		// Time
-		"sleep": sleep,
-		"time":  timeCmd,
+		"benchmark": benchmark,
+		"sleep":     sleep,
+		"time":      timeCmd,
 
 		"-ifaddrs": _ifaddrs,
 	})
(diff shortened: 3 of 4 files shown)


xiaq:
This should be a separate command. No need to cram as much functionality as possible to a single command.

krader1961:
> This should be a separate command. No need to cram as much functionality as possible to a single command.

Doing so certainly makes it much easier to provide more options, in an easier to document and use format, for controlling the benchmark execution. I've already refactored my code into a separate `benchmark` command (leaving the `time` command unchanged). I've replaced the `&benchmark` option with `&min-iters` and `&min-time`. Updated PR forthcoming.

krader1961:
@xiaq: Please take another look. I've introduced a `benchmark` command and also made the associated unit tests more robust.

pkg/eval/options.go
```diff
@@ -37,6 +37,16 @@ func scanOptions(rawOpts RawOptions, ptr any) error {
...
+
+		// An option with no value (e.g., `&a-opt`) has `$true` as its default value. However, if
+		// the option struct member is a string we want an empty string as the default value.
+		switch b := v.(type) {
+		case bool:
+			if b && structValue.Field(fieldIdx).Type().Name() == "string" {
+				v = ""
+			}
+		}
+
```
xiaq, line 49: Do not change the semantics of `&opt`. This introduces a disparity with Elvish functions, whose options do not have an intrinsic type.

xiaq:
I have not reviewed the rest of the PR. The change to the semantics of `&opt` to suit the API of this one command is not going to fly for me.

pkg/eval/options.go
krader1961: Options do have an intrinsic type but I grok your point. I added that code in the early stages of implementing this change because I was surprised by this behavior:

```
> > benchmark &min-time { }
Exception: wrong type: need string, got bool
[tty 1], line 1: benchmark &min-time { }
```

It seemed more useful to me that use case should behave as if an empty string was assigned to the string option; i.e., as if the user had typed `benchmark &min-time='' { }`. I now agree with you that an option without a value in that case should be an error if the option is not a boolean. Updated PR in a few minutes.

krader1961: And, unsurprisingly, removing that code does not affect any of the unit tests I wrote since none of them use a string option without providing a value. So removing that ill-advised change is the right thing to do. Forced push update with that change in a few minutes.

krader1961: FWIW, my original change that introduced a `time &benchmark` option relied on the behavior of this particular aspect of the PR so that someone could run `time &benchmark { code }` to enable benchmark behavior without explicitly specifying the benchmark options (thus get the benchmark defaults). It still seems useful to me, as a general principal, to allow string and numeric options to be used without a value; thus defaulting to the empty string or zero respectively. But that discussion is no longer pertinent to this change so I'm happy to punt it down the road when there is a stronger case for introducing such behavior.

xiaq: > Options do have an intrinsic type but I grok your point.

Values have intrinsic types.

In a function like `{|&opt=foo| }`, the `&opt` has no intrinsic type.

xiaq:
A few more issues when integrating this to the main branch:

- I was surprised by the behavior of only keeping the minimal execution time. This might make sense for certain use cases, but it's a surprising default. If `benchmark` reports a single number, it should be the average.

  Other metrics (like the minimum) can still be interesting. One approach I considered was passing the full list of durations to `&on-end`, so that it can make the decision to report whatever metric is interesting. But this could lead to runaway RAM usage if `$callback` is fast enough to get called many times.

  I eventually decided to record several interesting metrics that can be easily calculated in a streaming algorithm - average, standard deviation, min and max, and pass all of them to `&on-end`. The default reporting now also reports all of these.

- The default `&min-time` of 10s seems way too high, especially for microbenchmarks. I changed it to 1s.

- The `&min-time` shouldn't support unitless values, so I removed that.

  The `sleep` command supports that because people may be familiar with the Unix's `sleep` command, which doesn't apply here.

- The terminology around what each run is called is confused: `&on-run` suggests "run" while `&min-iters` suggests "iter[ation]".  I standardized on the more plain "run" since there isn't any compelling reason to introduce the term "iteration" here.

- The name `&on-run` doesn't say *when* during a run it's called, so I changed it to `&on-run-end`. It was also undocumented, despite appearing in the signature and the example.

krader1961:
> If benchmark reports a single number, it should be the average.

I disagree. Especially if the single number isn't more sophisticated than a simple average that doesn't exclude outliers. When performing a quick-and-dirty benchmark you only care about the minimum 99% of the time. That's why I included an option to collect all the run times. That allows the user to perform a more sophisticated analysis of the benchmark. I agree with your other comments.

PR #1596, opened 2022-08-05
Document `edit:listing:start-custom` example

brunoroque06:
Is the example too "complex"? Not in terms of complexity per se, but because it uses several functions: `put`/`each`/`styled`.

### DIFF
--- a/pkg/edit/listing_custom.go
+++ b/pkg/edit/listing_custom.go
@@ -28,6 +28,11 @@ func (*customListingOpts) SetDefaultOptions() {}
 //elvdoc:fn listing:start-custom
 //
 // Starts a custom listing addon.
+//
+// ```elvish
+// var opts = (put 0 1 | each {|n| put [&to-accept=$n &to-filter=$n &to-show=(styled $n green)] } | put [(all)])
+// edit:listing:start-custom &caption='Pick One' &accept={|o| echo $o } $opts
+// ```
 
 func listingStartCustom(ed *Editor, fm *eval.Frame, opts customListingOpts, items any) {
 	var bindings tk.Bindings


pkg/edit/listing_custom.go
```diff
@@ -28,6 +28,11 @@ func (*customListingOpts) SetDefaultOptions() {}
 //elvdoc:fn listing:start-custom
 //
 // Starts a custom listing addon.
+//
+// ```elvish
```
krader1961, line 32: This should have a sentence or two explaining what the example does. I also copy/pasted the statements into my Elvish shell and was perplexed by the results since it resulted in output such as `[bytes out] 0`. I don't think this is a realistic example of how to use `edit:listing:start-custom`.

brunoroque06: I agree that it is not a realistic example. I did not want to create an example with too many "dependencies", and I was more interested in showing the signature/arguments.

Actually when I starting coming up with an example, I was doing something along the lines: `git branch --format='%(refname:short)' | from-lines | ... | edit:listing:start-custom [(all)] &accept={ |b| git checkout -b $b }`. But then I thought it was a better idea to just show how to use it.

I use the following:

```elvish
az account list ^
  | from-json ^
  | all (one) ^
  | each { |s| put [&to-filter=$s[name] &to-accept=$s[id] &to-show=(if (put $s[isDefault]) { styled $s[name] green } else { put $s[name] })] } ^
  | edit:listing:start-custom [(all)] &caption='Azure Subscription' &accept={ |s| az account set --subscription $s }
```

Or:

```elvish
store:cmds 0 -1 ^
  | each { |c| put [&to-filter=$c[text] &to-accept=$c[text] &to-show=$c[text]] } ^
  | edit:listing:start-custom [(all)] &caption='Yank Command' &accept={ |c| printf $c | pbcopy }
```

Or:

```elvish
pulumi stack ls --json ^
   | from-json ^
   | all (one) ^
   | each { |s| put [&to-filter=$s[name] &to-accept=$s[name] &to-show=(if (put $s[current]) { put (styled $s[name] green) } else { put $s[name] })] } ^
   | edit:listing:start-custom [(all)] &caption='Pulumi Stack' &accept={ |s| pulumi stack select $s }
```

But of course these are bad examples to document.

What do you think? Would the `git checkout -b` example be a good one? I was also thinking that then it asks the question: why not simply use autocompletion instead? I could think of the following advantages when using `edit:listing:start-custom`:

- it is easier to remember a single elvish function, than a command with subcommands and flags;
- does not pollute the history with variations of a command (not always an advantage of course).

Are there more?

krader1961: Despite using Elvish for four years and contributing PRs for over two years I haven't felt the need to use the `edit:listing:start-custom` function. Too, there are only two uses in the third-party Elvish modules I have installed, which I didn't notice until I looked for its use as a result of this PR. Those modules are: github.com/zzamboni/elvish-modules/bang-bang.elv and github.com/zzamboni/elvish-modules/dir.elv. So I don't have a strong opinion about how best to document the behavior of this function but perhaps @zzamboni does.

PR #1598, opened 2022-08-08
Implement file:is-tty

### LINKED ISSUE #1263: A `file:is-tty` command would be useful
The equivalent of the POSIX `test -t $fd` command would be useful. It should default to the Elvish equivalent of `os.Stdout` to allow for LHS of a pipe and redirections. It should optionally accept an explicit `*os.File` or `vals.File` argument with any other type (e.g., a pathname string, including something like `/dev/tty`) being an error.

Related #1262

krader1961:
Resolves #1263

### DIFF
--- a/pkg/mods/file/file.go
+++ b/pkg/mods/file/file.go
@@ -8,15 +8,63 @@ import (
 	"src.elv.sh/pkg/eval"
 	"src.elv.sh/pkg/eval/errs"
 	"src.elv.sh/pkg/eval/vals"
+	"src.elv.sh/pkg/sys"
 )
 
 var Ns = eval.BuildNsNamed("file").
 	AddGoFns(map[string]any{
 		"close":    close,
+		"is-tty":   isTty,
 		"open":     open,
 		"pipe":     pipe,
 		"truncate": truncate,
 	}).Ns()
+
+//elvdoc:fn is-tty
+//
+// ```elvish
+// file:is-tty $file-obj?
+// ```
+//
+// Outputs `$true` if `$file-obj` is open on a tty (i.e., a terminal device);
+// otherwise, outputs `$false`. If the `$file-obj` argument is omitted the
+// default output byte stream is tested.
+//
+// ```elvish-transcript
+// ~> file:is-tty
+// ▶ $true
+// ~> var fh = (file:open /dev/tty)
+// ~> file:is-tty $fh
+// ▶ $true
+// ~> if (file:is-tty) { echo no } else { echo yes }
+// yes
+// ~> var fh = (file:pipe)
+// ~> file:is-tty $fh
+// ▶ $false
+// ~> var fh = (file:open /dev/null)
+// ~> file:is-tty $fh
+// ▶ $false
+// ```
+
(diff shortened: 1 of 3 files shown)


pkg/mods/file/file.go
```diff
@@ -1,23 +1,75 @@
...
+
+//elvdoc:fn is-tty
+//
+// ```elvish
+// file:is-tty $file-obj?
+// ```
+//
+// Does nothing if `$file-obj` is open on a tty (i.e., a terminal device); otherwise, raises the
+// "not a tty" exception. If the `$file-obj` argument is omitted the default output byte stream is
+// tested.
```
xiaq, line 34: The name of the function (`is-tty`) implies that it's a predicate. Like other predicates in Elvish, it should output a single boolean value instead of raising an exception.

krader1961: /face-palm Yes, of course. I have no idea why I decided it made more sense to raise an error.

pkg/mods/file/file.go
```diff
@@ -8,16 +8,64 @@ import (
...
 
+//elvdoc:fn is-tty
+//
+// ```elvish
+// file:is-tty $file-obj?
+// ```
+//
+// Outputs `$true` if `$file-obj` is open on a tty (i.e., a terminal device);
+// otherwise, outputs `$false`. If the `$file-obj` argument is omitted the
+// default output byte stream is tested.
```
xiaq, line 31: Only supporting testing stdout is an unnecessary limitation; it can be useful to test stdin and stderr too.

`is-tty` should accept either a file object or a number representing a FD. So testing stdin, stdout and stderr can be `file:is-tty 0`, `file:is-tty 1` and `file:is-tty 2`. A FD can be converted to a os.File with (os.NewFile)[https://pkg.go.dev/os#NewFile].

This will also simplify the implementation:

- `is-tty` now only needs to deal with a single argument

- ByteOutput no longer needs to expose the underlying os.File.

pkg/mods/file/file.go
```diff
@@ -8,16 +8,64 @@ import (
...
+// yes
+// ~> var fh = (file:pipe)
+// ~> file:is-tty $fh
+// ▶ $false
+// ~> var fh = (file:open /dev/null)
+// ~> file:is-tty $fh
+// ▶ $false
+// ```
+
+func isTty(fm *eval.Frame, fileObj ...any) (bool, error) {
```
xiaq, line 49: The Go function should be called isTTY.

pkg/mods/file/file_test.go
```diff
@@ -68,6 +68,12 @@ func TestFile(t *testing.T) {
 			What:  "size argument to file:truncate",
 			Valid: "integer", Actual: "non-integer",
 		}),
+
+		// TODO: Improve these tests if, and when, https://b.elv.sh/1595 is resolved.
```
xiaq, line 72: It is both shorter and more informative to just say what needs to be improved:

```go
// TODO: Test with PTY when https://b.elv.sh/1595 is resolved
```

PR #1604, opened 2022-09-01
Add support for completing invocations of local executables

exactly-one-kas:


pkg/edit/complete/complete_test.go
```diff
@@ -363,8 +363,35 @@ func TestComplete(t *testing.T) {
...
 
-	// Symlinks and executable bits are not available on Windows.
-	if runtime.GOOS != "windows" {
+	if runtime.GOOS == "windows" {
+		// Symlinks require admin permissions on Windows, so we won't test them
+
+		// Check completions with forward slash
+		allLocalCommandItems := []modes.CompletionItem{
+			fci("./a.exe", " "), fci("./d\\", ""),
+		}
```
xiaq, line 372: Since this is only used once below, it's better to just inline it. This also avoids reusing the same name for different test cases.

krader1961: While each instance of this var is used once I would argue that the separate variable improves readability. Presumably the compiler does something intelligent vis-a-vis avoiding unnecessary heap allocations, and if it doesn't that is less important than the improved readability given this is a unit test.

pkg/edit/complete/complete_test.go
```diff
@@ -363,8 +363,35 @@ func TestComplete(t *testing.T) {
...
+		tt.Test(t, tt.Fn("Complete", Complete), tt.Table{
+			// Complete local external commands.
+			Args(cb("./"), ev, cfg).Rets(
+				&Result{
+					Name: "command", Replace: r(0, 2),
+					Items: allLocalCommandItems},
+				nil),
+		})
+
+		// Check completions with bckward slash
```
krader1961, line 382: Typo: `bckward` => `backward`

Issue #1377, opened 2021-07-22
exit builtin does not trigger program's defer statements


matttproud:
I was prototyping with elvish today as a part of an experiment, and I noticed that `exit` builtin in the interactive shell as opposed to `^D` bypasses `defer` statements in `prog.Run`, like [this one for pprof](https://github.com/elves/elvish/blob/659bd5fa68d4b61a6c916f2cafca360c9e37482b/pkg/prog/prog.go#L122).  This is because, `^D` appears to [cause `Program.Run` to return](https://github.com/elves/elvish/blob/659bd5fa68d4b61a6c916f2cafca360c9e37482b/pkg/prog/prog.go#L148), presumably due to `io.EOF`; whereas [`exit`](https://github.com/elves/elvish/blob/9a576529a3029a29d6ca68607f5088e90a2e60eb/pkg/eval/builtin_fn_cmd.go#L110) calls `os.Exit`, which [bypasses any defers](https://changelog.com/gotime/165).

Is this behavior intentional, or is it a bug?  If this is a bug, I would propose solving it by propagating a structured error value from `eval.exit` that parent callers could handle, presumably [around here](https://github.com/elves/elvish/blob/9a576529a3029a29d6ca68607f5088e90a2e60eb/pkg/prog/prog.go#L146) in `prog.Run`, assuming one does not already exist:

```
type ExitError struct {
  Code int
}

func (ExitError) Error() string { ... }
```

Would doing that make sense?  If so, I might be happy to look at fixing this.  I see some evidence that there is a RPC subsystem in here, so I presume that the error type might need to be serializable for that as well.

First time user of elvish, so apologize if I am misunderstanding something here.  From what I've seen, this is a very nicely written Go program and intriguing project.  By the way, is this the same @krader1961 from Platforms at Google?  If so, hello again!  I learned a lot working with your team.

krader1961:
[...]

PR #1606, opened 2022-09-11
Move profiling back into the standard binary

### LINKED ISSUE #1605: The `-cpuprofile` flag no longer works
I was working on a fix for issue #1377 and was surprised that

1) the default `elvish` binary did not support the `-cpuprofile` flag, and

2) that building with `go build -o ./elvish ./cmd/withpprof/elvish` did not result in an Elvish binary that correctly handled the `-cpuprofile` flag. With, or without, my fix for handling the `exit` command.

The reason for both issues is commit 33a04f8dc, merged seven months ago, that refactored the code in response to the desire to minimize the size of the `elvish` binary for unusual situations. Such as using `elvish` as a shell in an embedded environment where every byte in the size of the binary matters. 

Issue #1 is because we now have to do something like `go build ./cmd/withpprof/elvish` rather than simply running `make get` (which, FWIW, I have never liked as a `make` target name).

Issue #2 is because the [defer](https://github.com/elves/elvish/blob/2986d29f34113823f39243715fe0fd70b6925681/pkg/pprof/pprof.go#L29) in the cpuprofile setup stops profiling when the [pprof `Run()`](https://github.com/elves/elvish/blob/2986d29f34113823f39243715fe0fd70b6925681/pkg/pprof/pprof.go#L21) function returns. Which is before anything useful to be profiled is executed.

On my macOS ARM system the `elvish` binary without pprof support is 9.5 MB, and with it the size is 9.8 MB. That is a 3.2% increase in size for profiling support. Performance metrics are extremely useful; especially in a relatively new shell like Elvish that is undergoing rapid development. A borderline insignificant increase in the size of the binary is worthwhile, IMHO, to make it easy to capture performance metrics that will help us figure out where it is most advantageous to make performance improvements.

krader1961:
This reintroduces the `-cpuprofile` flag into the standard Elvish binary. It also adds a `-memprofile` flag. Since Elvish programs are likely to cause lots of heap allocations that profiling option is likely to be helpful. Adding that support has a barely measurable effect on the size of the Elvish binary compared to just CPU profiling alone.

I measured the size of the Elvish binary with and without this change on the platforms I use to validate the behavior of Elvish. They are all 64-bit platforms with the exception of ubuntu32 (specifically, Ubuntu 16.04 32-bit). Adding profiling increased the size of the Elvish binary between 3.1% and 3.4%:

    macOS    3.4%
    osuse    3.1%
    fedora   3.1%
    ubuntu   3.4%
    ubuntu32 3.4%
    freebsd  3.1%
    windows  3.4%

That is obviously not a trivial increase but seems to me to be acceptable given the probable value it adds to debugging Elvish performance issues in the future.

Fixes #1605
Related #1377


xiaq:
This was moved out of the standard build exactly because of the binary size issue. It is not useful to most end users of Elvish.

krader1961:
> This was moved out of the standard build exactly because of the binary size issue. It is not useful to most end users of Elvish.

I understand your first point but disagree with your second point. While most end users are not going to ever use the profiling feature it is likely that we will want to ask users to provide that information when they report performance related problems. That may not be today, or even in the immediate future. But it will inevitably occur. Is it reasonable to ask the user to build a custom Elvish binary to help us resolve a performance issue? I am certainly ambivalent since the increase in size if greater than one percent but it is still small enough to be shrugged off given the potential usefulness.

I'm willing to try and fix the `cmd/withpprof/elvish` build target. I didn't do so because it wasn't immediately obvious how to fix the problem and reinstating the capability in the primary Elvish binary was trivial. Also, if we're going to support alternative builds as first-class citizens there really should be `make` targets and they should be documented.

xiaq:
> I understand your first point but disagree with your second point. While most end users are not going to ever use the profiling feature it is likely that we will want to ask users to provide that information when they report performance related problems. That may not be today, or even in the immediate future. But it will inevitably occur. Is it reasonable to ask the user to build a custom Elvish binary to help us resolve a performance issue? I am certainly ambivalent since the increase in size if greater than one percent but it is still small enough to be shrugged off given the potential usefulness.

That's a valid point, but that's assuming a simplistic situation:

1. End user experiences performance issues
2. End user is requested to run Elvish with profiling on and send a profile
3. Developer inspects profile and makes improvement
4. End user confirms performance improvement

While this workflow may be sufficient in some cases, in the majority of performance optimization work there needs to be multiple iterations of profiling and optimizing, and ideally some sort of performance regression test should be added. To do this, either:

- The developer needs to reproduce the environment the end user's code runs in and do the iterations themself, and the developer should know how to build the `withpprof` variant; or

- The end user needs to test the developer's build and run profiles multiple times, in which case the end user needs to know how to download or build a patch anyway, and the initial friction of building the `withpprof` variant is perhaps negligible.

I'd like to contrast this with crash reporting: debug symbols take up a lot of space, yet I'm not considering stripping them from the Elvish binary. This is because unlike performance optimization, it *is* very easy for a developer to look at a stack trace from a user and find the correct way to fix a crash. Performance optimization hardly admits this sort of simplistic workflow. Sure, it might be possible *initially* when there are a lot of low-hanging fruits, but such low-hanging fruits are not that hard to find without end user's profiling results anyway.

krader1961:
> While this workflow may be sufficient in some cases, in the majority of performance optimization work there needs to be multiple iterations of profiling and optimizing, and ideally some sort of performance regression test should be added. To do this, either:

This counter argument is not convincing to me. Specifically, you say that the end user reporting a performance problem simply needs to download a patched version of Elvish and build the `withpprof` variant to verify a proposed fix addresses their performance problem. However, there is no `make withpprof` target. Nor should there be, in my opinion. Asking an end user to jump through hoops to build a non-standard Elvish binary, rather than just doing `make get`, seems problematic to me. Having said all that I would be satisfied to see the hypothetical `withpprof` target documented and fixed so that it works; i.e. produces profiling data.

xiaq:
d7c5e35d39304f302371b12827cc5e22001f99e9 documented how to build a variant.

krader1961:
Okay, I'll close this P.R. since it is clear you are, not unreasonably, opposed to including profiling support in the default Elvish binary. That, obviously, leaves issue #1605 (`-cpuprofile` doesn't work) unresolved.

PR #1613, opened 2022-09-21
Change `make lint` to be comprehensive

krader1961:
I (almost) always run `make lint` on a change I plan on turning into a pull-request. However, the usefulness of that is limited because it only performs two Go source code lint checks. Which is a small fraction of the actual lint checks run in the continuous integration environment. Which means that I am regularly surprised by CI failures due to lint checks I didn't think to run because I expected `make lint` to run all the checks run by the CI environment.

This change modifies the top-level "lint" target to run all the lint checks. Also, with this change `make lint` produces no output if there is no lint.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ GNU Make is required.
 
 The [`tools`](tools) directory contains scripts too complex to fit in the
 `Makefile`. Among them, [`tools/pre-push`](tools/pre-push) can be used as a Git
-hook, and covers a large subset (but not all) of CI checks.
+hook, and covers all of the CI checks.
 
 ## Testing changes
 
@@ -145,26 +145,14 @@ Some of the generation rules depend on the `stringer` tool. Install with
 
 ## Code hygiene
 
-Some basic aspects of code hygiene are checked in the CI.
+Some basic aspects of code hygiene are checked in the continuous-integration
+environment. You are encouraged to install the lint tools and run them before
+opening a pull-request.
 
-### Formatting
+### Installing Lint Tools Used By This Project
 
-Install [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) to
-format Go files, and [prettier](https://prettier.io/) to format Markdown files.
-
-```sh
-go install golang.org/x/tools/cmd/goimports@latest
-npm install --global prettier@2.7.1
-```
-
-Once you have installed the tools, use `make style` to format Go and Markdown
-files. If you prefer, you can also configure your editor to run these commands
-automatically when saving Go or Markdown sources.
-
-Use `make checkstyle` to check if all Go and Markdown files are properly
-formatted.
-
-### Linting
+You'll need several tools to check that your changes do not introduce "lint";
+that is, style or correctness issues.
 
 Install [staticcheck](https://staticcheck.io):
 
@@ -176,9 +164,13 @@ The other linter Elvish uses is the standard `go vet` command. Elvish doesn't
 use golint since it is
 [deprecated and frozen](https://github.com/golang/go/issues/38968).
 
-Use `make lint` to run `staticcheck` and `go vet`.
+Install [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) to
+format Go files, and [prettier](https://prettier.io/) to format Markdown files.
 
-### Spell checking
+```sh
+go install golang.org/x/tools/cmd/goimports@latest
+npm install --global prettier@2.7.1
+```
 
 Install [codespell](https://github.com/codespell-project/codespell) to check
 spelling:
@@ -187,18 +179,26 @@ spelling:
 pip install --user codespell==2.1.0
 ```
 
-Use `make codespell` to run it.
+### Checking For Lint Problems
 
-### Running all checks
+Run `make lint` to execute all lint checks.
 
-Use this command to run all checks:
+The lint tests are run by the continuous-integration environment. It is
+recommended that you use `tools/pre-push`. That script will run `make lint`,
+`make test`, and also perform some other sanity checks such as a
+cross-compilation succeeding. Thus helping ensure you aren't surprised by CI
+failures due to style, spelling, and other lint mistakes as well as more
+fundamental problems (e.g., unit tests failing). See the comment at the top of
+`tools/pre-push` for how to use it.
 
-```sh
-make test checkstyle lint codespell
-```
+There are also `make` targets for each individual lint tool if you want to run a
+specific lint check. See the `Makefile` for details.
 
-You can put this in `.git/hooks/pre-push` to ensure that your published commits
-pass all the checks.
+### Fixing Style Problems
+
+Run `make style` to format Go and Markdown files. If you prefer, you can also
+configure your editor to run that command automatically when saving Go or
+Markdown files.
 
 ## Licensing
 
--- a/tools/check-content.sh
+++ b/tools/check-content.sh
@@ -1,17 +1,33 @@
-#!/bin/sh -e
+#!/bin/sh
 
 # Check Go source files for disallowed content.
+status=0
 
-echo 'Disallowed import of net/rpc:'
-if find . -name '*.go' | xargs grep '"net/rpc"'; then
-    exit 1
-else
-    echo '  None!'
+x="$(find . -name '*.go' | xargs grep '"net/rpc"')"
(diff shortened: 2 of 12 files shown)


krader1961:
On the `master` branch I attempted to `git push` to my Github repo, using the `./tools/pre-push` hook (as modified by this PR) and saw the following output (ignoring the output of `make test` run by that script):

```
Go files need these changes:
  None!
Markdown files that need changes:
  None!
codespell --skip .git
./website/learn/effective-elvish.html:105: bu ==> by
./website/learn/unique-semantics.html:172: bu ==> by
./website/learn/unique-semantics.html:174: bu ==> by
./website/learn/unique-semantics.html:182: bu ==> by
./website/learn/unique-semantics.html:231: bu ==> by
./website/learn/unique-semantics.html:236: bu ==> by
./website/learn/unique-semantics.html:237: bu ==> by
./website/learn/unique-semantics.html:242: bu ==> by
./website/learn/unique-semantics.html:327: bu ==> by
./website/learn/tour.html:361: bu ==> by
./website/blog/0.11-release-notes.md:133: reliabily ==> reliably, reliability
./website/_dst/learn/effective-elvish.html:801: bu ==> by
./website/_dst/learn/unique-semantics.html:868: bu ==> by
./website/_dst/learn/unique-semantics.html:870: bu ==> by
./website/_dst/learn/unique-semantics.html:878: bu ==> by
./website/_dst/learn/unique-semantics.html:927: bu ==> by
./website/_dst/learn/unique-semantics.html:932: bu ==> by
./website/_dst/learn/unique-semantics.html:933: bu ==> by
./website/_dst/learn/unique-semantics.html:938: bu ==> by
./website/_dst/learn/unique-semantics.html:1023: bu ==> by
./website/_dst/learn/tour.html:1060: bu ==> by
./pkg/cli/term/reader_unix.go:281: beging ==> begging, beginning, being
./pkg/cli/lscolors/feature.go:40: RO ==> TO, ROW, ROB, ROD, ROE, ROT
make: *** [codespell] Error 65
error: failed to push some refs to 'https://github.com/krader1961/elvish'
Exception: git exited with 1
[tty 17], line 1: git push
```

I didn't expect any of that output and the `bu` spelling errors are because the `codespell` check is looking at the generated HTML which should be ignored.

krader1961:
> On the master branch I attempted to git push to my Github repo, using the ./tools/pre-push hook (as modified by this PR) and saw the following output (ignoring the output of make test run by that script):

That is incorrect because I created a symlink to *./tools/pre-push* while implementing this change. So the errors in my previous comment do not include the changes included in this PR. They reflect the behavior of the pre-existing pre-push script. Which is another reason for merging this change.

krader1961:
Ping? I'd like to get this merged because I'm getting codespell errors after upgrading the version I have installed and this PR silences them; either by ignoring the website generated .html files or explicitly whitelisting a couple more words. I also much prefer having no output from these linters when no problems are found.

CONTRIBUTING.md
```diff
@@ -19,7 +19,7 @@ GNU Make is required.
 
 The [`tools`](tools) directory contains scripts too complex to fit in the
 `Makefile`. Among them, [`tools/pre-push`](tools/pre-push) can be used as a Git
-hook, and covers a large subset (but not all) of CI checks.
+hook, and covers all of the CI checks.
```
xiaq, line 22: Revert; this is not true. CI checks run on multiple platforms and can't be easily simulated locally.

tools/check-content.sh
```diff
@@ -1,17 +1,33 @@
-#!/bin/sh -e
```
xiaq, line 1: For all the shell script changes:

- Not sure why it's desirable to have different exit statuses depending on whether the script is running in CI or not.

- The cosmetic changes are good per se but they make the script more complex. I'd rather keep these shell scripts as simple as possible.

tools/pre-push
```diff
@@ -5,21 +5,21 @@
 #   cd .git/hooks # from repo root
 #   ln -s ../../tools/pre-push .
 
-if ! git diff HEAD --quiet; then
```
xiaq, line 8: Don't take away the functionality to be able to push with a dirty staging area. I value this functionality over seeing the exact style changes required; on a local environment, it's easy enough to just run `make style` and make styling issues go away.

xiaq (commented):
I'm not going to merge this PR since it does a lot more than just what the title advertises (changing `make lint` to be comprehensive), and I don't agree with most of the additional changes.

I also don't feel it's necessary to make `make lint` comprehensive. If you like, feel free to introduce another (say) `allchecks` target that runs all the existing checks.

I've manually cherry-picked the codespell changes and committed them as 89063ab095605d144e314e692d870cb2e4054a2d.

CONTRIBUTING.md
krader1961: Hmmm, okay, but it seems like a distinction that isn't particularly important as it runs all the CI checks relevant for the developer's platform. Too, the fact that CI environments typically employ multiple combinations of architectures and OS's is taken as a given by any developer who is not a complete noob. I would argue that saying "and covers all CI checks relevant to your platform" is clearer and more informative.

tools/check-content.sh
krader1961: > Not sure why it's desirable to have different exit statuses depending on whether the script is running in CI or not.

This allows `make lint` when run interactively to execute all checks rather than short-circuiting when a particular check fails. In an interactive context the developer is going to notice if there is a problem by observing the error messages. Especially with the other changes to only produce output from a test if it detects a problem. This means the developer will need to run `make lint` fewer times to resolve all problems. In the CI environment you want the detection of a problem to be noticed via the exit status. This refinement isn't, strictly speaking, necessary so I'll just remove that logic.

> The cosmetic changes are good per se but they make the script more complex. I'd rather keep these shell scripts as simple as possible.

I would argue they do not increase the complexity (especially if the exit status refinement is removed). Those changes are also not merely cosmetic. They greatly improve the usability by only producing output if a problem is detected. This makes it significantly easier to look at the `make lint` output and determine, at a glance, whether or not a problem was found.

tools/pre-push
krader1961: I get the sense this is your primary objection to this pull-request. If the other changes in this P.R. to broaden `make lint` to be comprehensive are merged then I have no objection to keeping the current pre-push hook behavior; although, I admit I don't understand why you want to allow pushing with a dirty staging area. But that's no doubt because you and I employ different Git workflows. If my staging area is dirty, but I still want to push, I expect to manually perform a `git stash` and would be surprised by the pre-push hook doing that automatically. But that's just me. 😄

krader1961:
> I also don't feel it's necessary to make make lint comprehensive. If you like, feel free to introduce another (say) allchecks target that runs all the existing checks.

I strongly disagree. In fact, that was explicit in this change by renaming the `lint` target to `codelint`. Then introducing a `lint` target that performs all the checks performed by the CI environment. I argue that this is what 99% of developers expect. That was my primary motivation in creating this pull-request. In other words, I have been surprised many times that running `make lint` on my system reported no problems while the CI environment reported problems I expected `make lint` to warn me about. This change makes it effectively impossible for `make lint` to tell me everything is okay yet the CI environment says there are problems (e.g., spelling or formatting errors).

krader1961:
> I strongly disagree. In fact, that was explicit in this change by renaming the lint target to codelint.

Also, historically, and in most projects, `make lint` is intended to warn a developer that their change has problems that will cause the change to be rejected. Not just narrow problems detected by tools like `go vet`. Most developers will expect `make lint` to report any problem detected by the CI environment. At least to the extent possible given the vagaries of various architectures, operating systems, and Go versions.

xiaq:
OK, I agree that there's value in making it easier to see the result of all the checks and minimizing noise in the output; I'll do that.

I don't like using the word "lint" to mean "all checks before push" though. Maybe it is used in such a way in some places at some time, but that's not my experience. Google internally uses the term "presubmit" (which I'm sure you're aware of), so does [Chromium](https://www.chromium.org/developers/how-tos/depottools/presubmit-scripts/). I think "pre-push checks" is also a good candidate because "submit" refers to an operation in Google's VCS, the closest equivalent in Git being push.

krader1961:
Please hold off on changing the behavior of `make lint`. I've been slowly reworking that aspect of my original P.R. to address some of your earlier points (e.g., the exit code of some scripts depending on whether they are being run in the CI environment) and improve the wording of some aspects of the change. I'll probably have something I'm willing to have reviewed in another day or two.

Issue #1222, opened 2021-01-14
Remove the daemon


xiaq:
In #1216 @krader1961 suggested using [this sqlite package](https://godoc.org/modernc.org/sqlite) instead of boltdb so that the database can be accessed from multiple processes without needing a daemon.

The package is quite mysterious though, as the [sqlite project ](https://sqlite.org/index.html) is implemented in C and that package appears to contain Go sources translated from sqlite's original C source, but I couldn't find that translation process documented anywhere.

xiaq:
I filed an issue: https://gitlab.com/cznic/sqlite/-/issues/41

It's also unclear how mature the project is. In an issue from 4 months ago (https://gitlab.com/cznic/sqlite/-/issues/27) the author said that the priority is "completeness and correctness", meaning that there are likely bugs and missing features.

krader1961:
The project update on 2020-07-26 (v1.4.0-beta1) reports:

> 630 errors out of 200177 tests on  Linux 64-bit little-endian
> Unfreed memory: 698816 bytes in 322 allocations

There are two questions:

1) What is the current state of those test failures and memory leaks five months later?

1) Are the unresolved issues likely to be problematic for Elvish or are they in niche areas of the code that Elvish wouldn't exercise?

Of the open issues the only one that would be a problem for Elvish is "[install issue on windows 10 with go 1.15](https://gitlab.com/cznic/sqlite/-/issues/23)" and it looks like there is someone actively working on making that package work on Windows 10.

**Note:** I do not care if a Go port of SQLite3 is used. I'd be equally happy to see a key/value DB written from scratch in idiomatic Go used if it supported access to the DB by multiple processes so the daemon could be eliminated. I say that because I have also experienced problems starting the daemon; even with elvish built from the master branch. Specifically, I initialize a bunch of tmux sessions in parallel to different machines. That frequently results in daemon initialization failures. I had to resort to adding a one second sleep after launching the first elvish shell on each machine to eliminate those daemon failures.

Sadly, the searches I've performed did not yield any promising alternatives to BoltDB. For example, it appears that all the key-value stores documented at https://github.com/gostor/awesome-go-storage only allow a single process to access the database. Which is why I'm still inclined to propose using a pair of text files (for commands and location mode data respectively) and using the algorithm from bash, fish, or another shell for accessing and mutating the contents of each file.

krader1961:
[...]

PR #1615, opened 2022-09-27
Remove support for shared vars

krader1961:
Proactively remove support for shared vars per the discussions in issue https://b.elv.sh/1222 and the IM channels. This feature is not used by Elvish itself and is not believed to be used by any Elvish users. Removing this (presumptively) unused feature makes it simpler to replace the existing BoltDB store of interactive command/location history with a simpler flat file implementation (such as used by Bash and Fish). Also, eliminating shared var support makes it explicit that solving issue #1222 doesn't need to deal with that feature.

Related #1222


0.19.0-release-notes.md
```diff
@@ -9,6 +9,8 @@ This is the draft release notes for 0.19.0, scheduled to be released around
 -   Symbolic links are now always treated as ordinary files by the global
     modifiers `type:dir` and `type:regular` in wildcard expansions.
 
+-   Support for shared vars has been removed; e.g., `store:shared-var`.
```
xiaq, line 12: I know "e.g.,` is the preferred style in US English, but Elvish's documentation never uses a comma after "e.g." or "i.e.". I'll remove that.

xiaq: It's probably best just to avoid *i.e.* and *e.g.*, as suggested by https://developers.google.com/style/abbreviations#dont-use suggests.

krader1961: > It's probably best just to avoid _i.e._ and _e.g._, as suggested by https://developers.google.com/style/abbreviations#dont-use suggests.

Good point. Those abbreviations are strongly English (Anglo Saxon/Latin) specific and may not be recognized by  individuals for whom English is a second language. However, note that there are a huge number of such uses in this project -- 257 lines by my count. What do you want to do about them? I'm happy to make a "janitor" change that replaces them and also augment the *CONTRIBUTING.md* file to make note of this project's preference.

xiaq:
In response to a message you sent to the user group on 2022-11-25:

> Kurtis Rader: @xiaqqaix: I am confused. Do you want to allow uses of i.e. and e.g. or not? I'm asking because you recently committed changes that introduced uses of those abbreviations. Yet you chastised me for using them in https://github.com/elves/elvish/pull/1615#discussion_r1027311627. I've been working on a change that replaces i.e. with that is and e.g. with for example. The change also includes a check for those abbreviations by the CI and tools/pre-push. Do you want to allow those abbreviations or not?

Look, I appreciate the feedback that I'm not applying the suggestion of avoiding "i.e." and "e.g." consistently, but:

- First, I said "it's probably best to avoid". It's not a commitment to any blanket policy.

- Second, you initially replied "good point" to my comment, but later described the interaction as me "chastising" you. I was expressing my opinion and you seemed to agree. If that was "chastising", you were picturing me as some sort of authority that I  never purported to be.

- Finally, I don't appreciate the tone here. I've said this before: Elvish is my project. To use a somewhat tired metaphor, this place is like my garden that I voluntarily opened up. I appreciate contributions and try to be hospitable, because I like knowing that other people enjoy my garden too. But at the end of the day, it's my garden, not a public park that I get paid to manage. I decide how to curate and trim the plants, and I will re-arrange and re-trim the plants contributed by guests. I usually explain my rationale, but I do that out of courtesy, not because I owe anybody that.

  The fact that I made the source code of Elvish available under a free license and put it on GitHub doesn't mean I signed up to any specific model of running a project. Sure, some successful projects are run under an "open governance" model that tries to hold all contributors to the same standard. But the F/OSS community is a diverse place, and many other successful projects are run essentially as personal projects. I develop Elvish mainly because it's a good outlet for my creativity, and would like to enjoy that quietly. Please respect that. Insisting that other people should run projects in a specific way is common in the F/OSS community than other places, but that doesn't make it a nice thing to do.

Getting back to the issue itself, I only raised such small stylistic issues as the use of "e.g." and "i.e." since your PRs often contain small stylistic changes, so I feel a need to explain why I don't agree with them and/or how I would word things differently. I'd recommend that you just don't bother with stylistic changes, and I won't bother you with my opinions either. These days I tend to merge PRs as is and do fixups myself anyway.

PR #1617, opened 2022-09-30
Improve some compilation errors

### LINKED ISSUE #1560: Misleading error message with an empty variable name
This error message is rather misleading:

```
⬥ var '' = foo                                                                                              ~
compilation error: 4-6 in [tty]: cannot create variable $; new variables can only be created in the local sco
pe
```

Now there are good reasons not to allow empty variable names. But the error message could be improved upon.

krader1961:
Quote variable names appearing in some compilation errors to improve the clarity of those errors.

Fixes #1560


xiaq:
This makes sense but it doesn't fix #1560.

PR #1618, opened 2022-10-06
Make testing compilation errors more deterministic

krader1961:
A year ago I submitted a change to replace AnyError with tests for specific errors (see https://github.com/elves/elvish/commit/87656c99).  This does something similar for DoesNotCompile. This ensures the test does what is implied and makes correlating specific unit tests with compilation errors easier.

This includes a couple of changes to compilation error messages to improve readability (IMHO) but those are not the primary purpose of this change.

Related #1560

### DIFF
--- a/pkg/eval/builtin_special.go
+++ b/pkg/eval/builtin_special.go
@@ -133,7 +133,7 @@ func compileDel(cp *compiler, fn *parse.Form) effectOp {
 		}
 		head, indices := cn.Indexings[0].Head, cn.Indexings[0].Indices
 		if head.Type == parse.Variable {
-			cp.errorpf(cn, "arguments to del must drop $")
+			cp.errorpf(cn, "arguments to del must omit the dollar-sign")
 		} else if !parse.ValidLHSVariable(head, false) {
 			cp.errorpf(cn, delArgMsg)
 		}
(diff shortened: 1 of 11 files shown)


pkg/eval/builtin_special.go
```diff
@@ -133,7 +133,7 @@ func compileDel(cp *compiler, fn *parse.Form) effectOp {
 		}
 		head, indices := cn.Indexings[0].Head, cn.Indexings[0].Indices
 		if head.Type == parse.Variable {
-			cp.errorpf(cn, "arguments to del must drop $")
+			cp.errorpf(cn, "arguments to del must omit the dollar-sign")
```
xiaq, line 136: > dollar-sign

Remove the hyphen

pkg/eval/compile_value_test.go
```diff
@@ -129,7 +129,7 @@ func TestTilde_ErrorForCurrentUser(t *testing.T) {
 
 func TestWildcard(t *testing.T) {
 	Test(t,
-		That("put ***").DoesNotCompile(),
+		That("put ***").DoesNotCompile("bad wildcard: \"***\""),
```
xiaq, line 132: Use a raw string if the string content contains `"`.

pkg/eval/evaltest/evaltest.go
```diff
@@ -122,9 +123,16 @@ func (c Case) Throws(reason error, stacks ...string) Case {
...
-// compilation.
-func (c Case) DoesNotCompile() Case {
-	c.want.CompilationError = anyError{}
+// compilation with a specific error. It accepts a string or a MatchingRegexp{}.
+func (c Case) DoesNotCompile(ce any) Case {
+	switch v := ce.(type) {
+	case string:
+		c.want.CompilationError = stringError{isRegexp: false, str: v}
+	case MatchingRegexp:
+		c.want.CompilationError = stringError{isRegexp: true, str: v.Pattern}
```
xiaq, line 132: Regular expressions are used in exactly one test case. Best to just stick to matching the full message.

Issue #570, opened 2018-01-20
Do not abuse panic for errors


xiaq:
Currently the Elvish codebase abuses panics for returning errors in the runtime (the `eval` package). This has turned out to cause more problems than it solves. Let's switch to the typical Go style of returning an `error` to indicate an error.

PR #1619, opened 2022-10-19
Change how `exit` works so defers are run

### LINKED ISSUE #1377: exit builtin does not trigger program's defer statements

krader1961:
This replaces the direct use of os.Exit() in the `exit` implementation with a panic so that deferred functions are run. For example, those involved in saving profiling data.

This also adds a sanity check to the value of the `exit` command. Every UNIX like system requires the value to be in the range [0..255]. So use that limit for now. It is true that Windows supports the range [0..2^32) but I don't see a good reason to support that at this time. If we find it useful to support a different range on non-UNIX systems the check can always be abstracted out into platform specific validation.

Fixes #1377

### DIFF
--- a/pkg/eval/builtin_fn_cmd.go
+++ b/pkg/eval/builtin_fn_cmd.go
@@ -3,8 +3,10 @@ package eval
 import (
 	"os"
 	"os/exec"
+	"strconv"
 
 	"src.elv.sh/pkg/eval/errs"
+	"src.elv.sh/pkg/prog"
 )
 
 // Command and process control.
@@ -93,7 +95,8 @@ func searchExternal(cmd string) (string, error) {
 // exit $status?
 // ```
 //
-// Exit the Elvish process with `$status` (defaulting to 0).
+// Exit the Elvish process with `$status` (defaulting to 0). The status must be
+// in the range 0 to 255 inclusive.
 
 // Can be overridden in tests.
 var osExit = os.Exit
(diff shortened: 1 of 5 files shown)


xiaq:
Abusing panics is a bad idea. The codebase used to do this a lot, but it created so many more problems that I decided to get rid of them (#570).

My memory is a bit hazy regarding the specific problems, but on the high level, having panics that are supposed to be caught means that you will have to assume *every* function can have such panics, and think about whether the code will misbehave when that happens. Sometimes these panics just cause resources to not get released properly (which is not a problem for `exit`), but a lot of other times they result in correctness issues, especially in code that needs to coordinate multiple goroutines.

So no, I'm not going to allow panic abuses back in the codebase.

There is a much cleaner approach - there is already a lot of code that propagates exceptions, so `exit` can throw a special exception value that can't be caught by `try`.

Finally re restricting the exit code: also no. Elvish shouldn't prevent users from writing unportable code. If Elvish must only provide portable facilities, the `exec` builtin needs to be removed because it can't be used on Windows, and Elvish probably shouldn't allow code to use external commands at all.

krader1961:
Okay, are you saying that use of the `exit` builtin explicitly means that deferred functions, such as for profiilng, will never be supported? Because I don't see a straightforward way to support profiling using your proposed approach:

> There is a much cleaner approach - there is already a lot of code that propagates exceptions, so exit can throw a special exception value that can't be caught by try.

I'm sure that can be made to work but am doubtful it is simpler, cleaner, easier to understand for this particular case. The `exit` command is very much a special case where the issues you allude to for other error propagation does not apply.

PR #1628, opened 2022-12-03
Small fixes on homepage

dunsany:
Besides the obvious: capitalizing headers seems unnecessary to me, but I don't call the shots here.

### DIFF
--- a/website/home.md
+++ b/website/home.md
@@ -15,7 +15,7 @@ Windows.
 <div id="demo-window"> <div id="demo-container">
   <div class="demo-wrapper"> <div class="demo">
     <div class="demo-col demo-description">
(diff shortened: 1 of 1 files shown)


xiaq:
There's indeed no standard on whether title case applies to headings.

I don't title-case section headings, but I think of the headings on the homepage more as headlines than section headings. https://www.python.org, https://racket-lang.org, https://lisp-lang.org, https://clojure.org all use this style. (Maybe I was just unconsciously influenced by homepages of programming languages.)

I've manually cherry-picked all the other changes in https://github.com/elves/elvish/commit/fe4ca3cb05709448ca30d353f40fa7bdc3c76e1c so I'll close this now. Much thanks.

Issue #1265, opened 2021-03-15
edit:del-var would be useful


zzamboni:
I think it would be useful go have an `edit:del-var`, analogous to `edit:add-var`, to remove a variable from the interactive scope.

For me, a specific use case would be to undefine [aliases](https://github.com/zzamboni/elvish-modules/blob/master/alias.org) in the current session. At the moment this can be done by redefining the alias variable to a function that calls the corresponding external command (see https://github.com/zzamboni/elvish-modules/blob/master/alias.elv#L95), but it seems just being able to remove the variable would be cleaner.

supreetsingh10:
Can I work on this one? I think I can implement this.

krader1961:
@supreetsingh10: No need to ask permission. In general, for open source projects like Elvish it is sufficient to simply say "I'm going to work on this issue." And a lot of the time even that isn't needed. Having said that, I think you are underestimating the difficulty of this problem relative to your knowledge of Elvish and Go. Take a look at any of my [pending pull requests](https://github.com/elves/elvish/pulls) such as PR #1270. All of those address an issue that is no more complicated than this issue. If you understand those changes you might be successful at implementing this feature.

krader1961:
@supreetsingh10: Take a look at PR #1221 which has been abandoned by its author. Given the feedback in that PR you should be able to create a new PR that correctly fixes the problem.

supreetsingh10:
Thank you for the guidance. That is true, I need to get more skilled before I tackle these problems. Thank you Mr. @krader1961.

supreetsingh10:
Mr. @krader1961 I think I have got this problem figured out. I have written some code, it is not enough to be pushed, so I am posting my doubts here, you can reply to them when you have time. 

Since I am working on ```edit:del-var```, I am currently in the working in ```vars.go``` and I have written and the function ```delVars()```, it is linking with the shell and running. I am using the ```addVar()``` to understand how this thing works. I want to ask that am I supposed to add ```eval.DeleteVarFromName()``` thing in ```closure,go``` as there is a function called ```MakeVarFromName()```. Is there a ```free()``` function which I have to use to ````delete a variable```.

krader1961:
[...]

Issue #1521, opened 2022-04-11
edit:add-var supplements


gergelyk:
Commands complementary to `edit:add-var` would be nice to have. E.g.:
```
edit:get-var - get value of variable specified by name
edit:list-vars - list names of available variables
```
It would help to integrate interactive shell with external tools.

krader1961:
@gergelyk: It would help evaluate, and prioritize, your proposal if you could provide some concrete use cases. What do you mean by "it would help to integrate interactive shell with external tools"? For example, presumably you mean that a hypothetical `edit:get-var` would only return a value for an interactive var created by `edit:add-var` and otherwise raise an exception. But how do you envision that being used? In other words, it is unclear to me exactly how the proposed feature would be useful; i.e., what problems it solves.

gergelyk:
Regarding `edit:get-var`... I was working on python integration layer. I ended up with:

```sh
var x = (num 123)
var y = (num 2)
var res: = (ns (pyeval [&x=$x &y=$y] 'z = x / y'))
echo $res:z  # 61.5
```

I was considering improving it towards implementation that imports/exports selected variables automatically:

```sh
var x = (num 123)
var y = (num 2)
pyeval &export=[x y] &import[z] 'z = x / y'
echo $z  # 61.5
```

Or even towards this variant that imports/exports all variables automatically:
```sh
var x = (num 123)
var y = (num 2)
pyeval 'z = x / y'
echo $z  # 61.5
```

Anyway, this would be limited to the interactive prompt only. In this case I consider staying with current implementation. Also note that this is an experimental "feature" only. I currently don't have any serious business behind.

Regarding `edit:del-var` - I actually have no good justification.

Regarding `edit:list-vars` - I believe that it would be a handful utility during development in the interactive shell. An equivalent of `dir()` from Python. This would tell me the answer whenever I wonder "is this or that symbol defined by rc.elv already?" or "how did I called this variable two minutes ago?".

gergelyk:
One more use case: I would like to have a handy command that shows definition of a function available in the interactive session. I ended up with this, which is not exactly accurate (works only with functions defined in `rc.elv`):
```
fn def {|name| { tmp pwd = ~/.elvish; eval 'use ./rc; echo $rc:'$name'~[def]' }}
```
hypothetical `edit:get-var` would improve it as follows:
```
fn def {|name| { echo (edit:get-var $name'~')[def] }
```

krader1961:
[...]

PR #1629, opened 2022-12-05
edit: Add del-var and del-vars

mmlb:
Fixes: https://github.com/elves/elvish/issues/1265
Issue: https://github.com/elves/elvish/issues/1521

### DIFF
--- a/pkg/edit/vars.go
+++ b/pkg/edit/vars.go
@@ -15,6 +15,7 @@ func initVarsAPI(ed *Editor, nb eval.NsBuilder) {
 		"add-var":  addVar,
 		"add-vars": addVars,
 		"del-var":  delVar,
+		"del-vars": delVars,
 	})
 }
 
@@ -72,6 +73,32 @@ func addVars(fm *eval.Frame, m vals.Map) error {
 		nb.AddVar(name, variable)
 	}
 	fm.Evaler.ExtendGlobal(nb)
+	return nil
+}
+
+func delVars(fm *eval.Frame, m vals.List) error {
+	for it := m.Iterator(); it.HasElem(); it.Next() {
+		n := it.Elem()
+		name, ok := n.(string)
+		if !ok {
+			return errs.BadValue{
+				What:  "name of argument to edit:del-vars",
+				Valid: "string", Actual: vals.Kind(n)}
+		}
+		if !isUnqualified(name) {
+			return errs.BadValue{
+				What:  "name of argument to edit:del-vars",
+				Valid: "unqualified variable name", Actual: name}
+		}
+
+		variable := fm.Evaler.Global().IndexString(name)
+		if variable == nil {
+			return errors.New("no variable $" + name)
+		}
+
+		fm.Evaler.DeleteGlobal(name)
+	}
+
 	return nil
 }
 
--- a/pkg/edit/vars.go
+++ b/pkg/edit/vars.go
@@ -1,6 +1,7 @@
 package edit
 
 import (
+	"errors"
 	"strings"
 
 	"src.elv.sh/pkg/eval"
@@ -13,6 +14,7 @@ func initVarsAPI(ed *Editor, nb eval.NsBuilder) {
 	nb.AddGoFns(map[string]any{
 		"add-var":  addVar,
 		"add-vars": addVars,
+		"del-var":  delVar,
 	})
 }
 
@@ -28,6 +30,22 @@ func addVar(fm *eval.Frame, name string, val any) error {
 		return err
 	}
 	fm.Evaler.ExtendGlobal(eval.BuildNs().AddVar(name, vars.FromInit(val)))
+	return nil
+}
+
+func delVar(fm *eval.Frame, name string) error {
+	if !isUnqualified(name) {
+		return errs.BadValue{
+			What:  "name argument to edit:del-var",
+			Valid: "unqualified variable name", Actual: name}
+	}
+
+	variable := fm.Evaler.Global().IndexString(name)
+	if variable == nil {
+		return errors.New("no variable $" + name)
+	}
+
+	fm.Evaler.DeleteGlobal(name)
 	return nil
 }
 
--- a/pkg/edit/vars_test.go
+++ b/pkg/edit/vars_test.go
@@ -67,3 +67,25 @@ func TestAddVars(t *testing.T) {
 		That("add-vars [&a~='']").Throws(ErrorWithType(vals.WrongType{})),
 	)
 }
+
+func TestDelVars(t *testing.T) {
+	TestWithSetup(t, func(ev *eval.Evaler) {
+		ev.ExtendGlobal(eval.BuildNs().AddGoFn("add-vars", addVars))
+		ev.ExtendGlobal(eval.BuildNs().AddGoFn("del-vars", delVars))
+	},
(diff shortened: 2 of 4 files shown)


mmlb:
Hi @xiaq / @krader1961. This is my first working stab at getting del-var support. I also added `del-vars` for completeness. I'm not really a fan of the way I went about with the `*Deleted` methods in `eval` but that was the easiest/less disruptive change I could think of, let me know if you would like to see it done another way and I'd be happy to make the changes.

mmlb:
My main use case for `del-var` is to combine it with direnv to avoid direnv's hook that needs to iterate through a json object calling `set-env` and `unset-env`. Also for use with nix-community/home-manager (possibly thru direnv actually).

mmlb:
hmm although now that I look at this again, I notice that maybe `edit:add-var` and `set-env` are not equivalent, right?

xiaq:
> I'm not really a fan of the way I went about with the `*Deleted` methods in `eval` but that was the easiest/less disruptive change I could think of, let me know if you would like to see it done another way and I'd be happy to make the changes.

I am not sure the change to `Ns` is necessary. You can just add a specialized `DeleteGlobals` method like this:

```go
func (ev *Evaler) DeleteGlobal(name string) {
  ev.mu.Lock()
  defer ev.mu.Unlock()
  g := ev.global.clone()
  for i := range g.infos {
    if g.infos[i].name == name {
      g.infos[i].deleted = true
    }
  }
  ev.global = g
}
```

And similarly for deleting multiple names.

> hmm although now that I look at this again, I notice that maybe `edit:add-var` and `set-env` are not equivalent, right?

`edit:add-var` manipulates the REPL's global namespace (variables like `$foo`) while `set-env` manipulates environment variables (like `$E:FOO`). These are completely disjoint in Elvish.

mmlb:
> > I'm not really a fan of the way I went about with the `*Deleted` methods in `eval` but that was the easiest/less disruptive change I could think of, let me know if you would like to see it done another way and I'd be happy to make the changes.
> 
> I am not sure the change to `Ns` is necessary. You can just add a specialized `DeleteGlobals` method like this:
> 
> ```go
> func (ev *Evaler) DeleteGlobal(name string) {
>   ev.mu.Lock()
>   defer ev.mu.Unlock()
>   g := ev.global.clone()
>   for i := range g.infos {
>     if g.infos[i].name == name {
>       g.infos[i].deleted = true
>     }
>   }
>   ev.global = g
> }
> ```
> 
> And similarly for deleting multiple names.

I'll give it try with this route.

> 
> > hmm although now that I look at this again, I notice that maybe `edit:add-var` and `set-env` are not equivalent, right?
> 
> `edit:add-var` manipulates the REPL's global namespace (variables like `$foo`) while `set-env` manipulates environment variables (like `$E:FOO`). These are completely disjoint in Elvish.

Right, I was just typing-out-loud that my original reason for the PR isn't addressed by the PR :man_facepalming: . Would you be open to `edit:add-env` `edit:del-env`?

mmlb:
@xiaq I've updated the PR with your suggestion, my edit tests passed locally so lets see if CI agrees :D.

xiaq:
> Right, I was just typing-out-loud that my original reason for the PR isn't addressed by the PR 🤦‍♂️ . Would you be open to `edit:add-env` `edit:del-env`?

Environment variables are shared within the entire process, so you can just use `set E:ENV_NAME = ...` and `del E:ENV_NAME`.

You can also use `set-env` and `unset-env` but they are only needed when the name is dynamic.

mmlb:
@xiaq wanted to ping you on this, even though I'm really mostly interested on `set E:ENV_NAME` and `del E:ENV_NAME`, I think its good to have the set/del pairs for this functionality.

pkg/edit/vars.go
```diff
@@ -31,6 +34,22 @@ func addVar(fm *eval.Frame, name string, val any) error {
...
+		return errs.BadValue{
+			What:  "name argument to edit:del-var",
+			Valid: "unqualified variable name", Actual: name}
+	}
+
+	variable := fm.Evaler.Global().IndexString(name)
+	if variable == nil {
+		return errors.New("no variable $" + name)
+	}
+
```
xiaq, line 48: There's a race condition here when the global namespace is swapped out after the check but before the `.DeleteGlobal` call.

For dynamic deletion calls, I feel it's best to just ignore keys that don't exist. This is how `dissoc` and `del E:FOO` behave today.

pkg/edit/vars.go
```diff
@@ -57,6 +76,32 @@ func addVars(fm *eval.Frame, m vals.Map) error {
 	return nil
 }
 
+func delVars(fm *eval.Frame, m vals.List) error {
+	for it := m.Iterator(); it.HasElem(); it.Next() {
+		n := it.Elem()
+		name, ok := n.(string)
+		if !ok {
+			return errs.BadValue{
+				What:  "name of argument to edit:del-vars",
```
xiaq, line 85: `name of argument to edit:del-vars` doesn't seem to be an accurate description of what the element is; `element of the argument to edit:del-vars` is better.

I think it's better to first convert the argument into a `[]string` and then delete all the arguments. Deleting them in a loop is prone to race conditions.

pkg/edit/vars_test.go
```diff
@@ -47,3 +67,25 @@ func TestAddVars(t *testing.T) {
 		That("add-vars [&a~='']").Throws(ErrorWithType(vals.WrongType{})),
 	)
 }
+
+func TestDelVars(t *testing.T) {
+	TestWithSetup(t, func(ev *eval.Evaler) {
+		ev.ExtendGlobal(eval.BuildNs().AddGoFn("add-vars", addVars))
+		ev.ExtendGlobal(eval.BuildNs().AddGoFn("del-vars", delVars))
```
xiaq, line 74: These two `.ExtendGlobal` calls should be combined:

```go
ev.ExtendGlobal(eval.BuildNs().AddGoFns(map[string]any{"add-vars": addVars, "del-vars": delVars}))
```

pkg/edit/vars_test.go
```diff
@@ -47,3 +67,25 @@ func TestAddVars(t *testing.T) {
...
+	},
+		That("add-vars [&foo=bar]").Then("del-vars [foo]").Then("put $foo").DoesNotCompile("variable $foo not found"),
+		That("add-vars [&a=A &b=B &c=C]").Then("del-vars [a b]").Then("put $c").Puts("C").Then("del-vars [c]").Then("put $c").DoesNotCompile("variable $c not found"),
```
xiaq, line 77: This test case seems to be testing two behaviors:

1. That deleting variables `a` and `b` leaves the variable `c` intact;
2. That deleting variable `c` makes it no longer exist.

Behavior 2 is already tested above and doesn't need to be repeated.

pkg/eval/eval.go
```diff
@@ -174,6 +174,19 @@ func (ev *Evaler) ExtendGlobal(ns Nser) {
 	ev.global = CombineNs(ev.global, ns.Ns())
 }
 
+// DeleteGlobal deletes a binding from the global namespace.
+func (ev *Evaler) DeleteGlobal(name string) {
```
xiaq, line 178: Take a string slice rather than a single string.

PR #1630, opened 2022-12-10
Validate generated code is up to date

krader1961:
While working on a change to augment `make lint` to be more comprehensive I noticed that `go generate ./...` produced unexpected errors on my system (see below). This augments the CI environment checks to verify that the generated code is up to date. This also augments the pre-push script to perform the checks by the tools/check-content script run by the CI environment.

FWIW: The errors shown below were because my version of the `stringer` tool was out of date (fixed by running `go install golang.org/x/tools/stringer@latest`) Regardless, the point of this change is that the pre-push script and the CI environment should verify that the generated code is up to date. This change adds those checks.

> go generate ./...
stringer: internal error: package "fmt" without types was imported from "src.elv.sh/pkg/getopt"
pkg/getopt/getopt.go:11: running "stringer": exit status 1 stringer: internal error: package "fmt" without types was imported from "src.elv.sh/pkg/md"
pkg/md/md.go:163: running "stringer": exit status 1 stringer: internal error: package "fmt" without types was imported from "src.elv.sh/pkg/parse"
pkg/parse/parse.go:11: running "stringer": exit status 1
Exception: go exited with 1
[tty 102], line 1: go generate ./...

### DIFF
--- a/tools/check-content.sh
+++ b/tools/check-content.sh
@@ -1,17 +1,39 @@
 #!/bin/sh -e
 
-# Check Go source files for disallowed content.
+# Check Go source files for disallowed or stale content.
 
-echo 'Disallowed import of net/rpc:'
-if find . -name '*.go' | xargs grep '"net/rpc"'; then
+# Verify the generated code is up to date.
+go generate ./...
+exit 0
+x="$(git diff || true)"
+if test "$x" != ""; then
+    echo "======================================================================"
+    echo "Generated Go code is out of date. See"
+    echo "https://github.com/elves/elvish/blob/master/CONTRIBUTING.md#generated-code"
+    echo "======================================================================"
+    echo "$x"
     exit 1
-else
-    echo '  None!'
 fi
 
-echo 'Disallowed call of unix.{Umask,Getrlimit,Setrlimit}'
-if find . -name '*.go' | egrep -v '\./pkg/(mods/unix|daemon|testutil)' | xargs egrep 'unix\.(Umask|Getrlimit|Setrlimit)'; then
+# Verify the code does not include undesired package imports.
+x="$(find . -name '*.go' |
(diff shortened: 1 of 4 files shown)


tools/check-content.sh
```diff
@@ -1,17 +1,39 @@
...
 
-# Check Go source files for disallowed content.
+# Check Go source files for disallowed or stale content.
 
-echo 'Disallowed import of net/rpc:'
-if find . -name '*.go' | xargs grep '"net/rpc"'; then
+# Verify the generated code is up to date.
+go generate ./...
+exit 0
+x="$(git diff || true)"
```
xiaq, line 8: Using `git diff` is not sufficient to determine whether the generated files are stale, since it doesn't detect untracked files. Using `git status` is better.

tools/check-content.sh
```diff
@@ -1,17 +1,39 @@
 #!/bin/sh -e
 
-# Check Go source files for disallowed content.
+# Check Go source files for disallowed or stale content.
 
-echo 'Disallowed import of net/rpc:'
-if find . -name '*.go' | xargs grep '"net/rpc"'; then
+# Verify the generated code is up to date.
```
xiaq, line 5: This new check doesn't really belong in this script. It is very different from the existing checks, it depends on Git and requires a clean Git repo.

tools/pre-push
```diff
@@ -14,9 +14,13 @@ if ! git diff HEAD --quiet; then
         echo 'Local changes exist and some are staged; not stashing.'
         echo 'Make a commit, stash all the changes, or unstage all the changes.'
         exit 1
-	fi
+    fi
 fi
 
+# This script is run by the CI environment but is not run by the `make`
+# targets below.
+./tools/check-content.sh
```
xiaq, line 22: There is already a make target `check-content` for this script.

PR #1637, opened 2023-01-05
Expand the description of setting environment variables

iandol:
Elvish used to allow the `foo=bar cmd` syntax similar to other shells. This has changed and as it is a common motif I thought it would be good to mention it in the elvish<->bash comparison table (at least me as a generic end-user often goes there to work out how to do things). I added both the `{ tmp ... }` option and also using `env` command as a shortcut, although I know this is not strictly "elvish" syntax...

I also thought that as environment variable are often important for many shell tasks, expanding the Setting variables section to detail the use of builtins like `set-env` or the `E:` namespace will help new users orient themselves.


xiaq (commented):
I'd rather move environment variables in their own sections, which I'll do after merging this.

PR #1638, opened 2023-01-05
pkg/edit: implement edit:history:accept

### LINKED ISSUE #1636: Way to stop history mode and start command mode without clearing line.
I wanted to close history navigation and start command(vi) mode to edit command in the history. 

Up -> ESC

But it clears the line. So I tried this but it doesn't work properly.

```
set edit:history:binding["Ctrl-["] =  { edit:command:start }
```

image

And this doesn't work as edit:close-mode clears the line.
```
set edit:history:binding["Ctrl-["] =  { edit:close-mode; edit:command:start }
```

I am searching for some api to preserve current line in history mode. 
Or let me know if anyone knows the trick. I'm not sure some new api should be introduced for this.

aca:
It's api to and update codebuffer with current entry in histwalk mode.

Resolves #1636

With this new api, I was able to switch to command-mode directly from histwalk mode.
```
set edit:history:binding["Ctrl-["] =  { edit:history:accept; edit:close-mode; edit:command:start }
```

I'm not familiar with elvish code base. Please let me know it this api itself and it's implementation is correct.

### DIFF
--- a/pkg/cli/modes/histwalk.go
+++ b/pkg/cli/modes/histwalk.go
@@ -17,6 +17,8 @@ type Histwalk interface {
 	Prev() error
 	// Walk to the next entry in history.
 	Next() error
+	// Update buffer with current entry.
+	Accept() error
 }
 
 // HistwalkSpec specifies the configuration for the histwalk mode.
@@ -119,4 +121,16 @@ func (w *histwalk) updatePending() {
 			Content: cmd.Text[len(w.Prefix):],
 		}
 	})
+}
+
+func (w *histwalk) Accept() error {
+	cmd, _ := w.cursor.Get()
+	txt := cmd.Text[len(w.Prefix):]
+	w.attachedTo.MutateState(func(s *tk.CodeAreaState) {
+		s.Buffer = tk.CodeBuffer{
+			Content: txt,
+			Dot:     len(txt),
+		}
+	})
+	return nil
 }
--- a/pkg/edit/histwalk.d.elv
+++ b/pkg/edit/histwalk.d.elv
@@ -21,3 +21,6 @@ fn history:down-or-quit { }
 # Import command history entries that happened after the current session
 # started.
 fn history:fast-forward { }
+
+# Update buffer with current entry.
+fn history:accept { }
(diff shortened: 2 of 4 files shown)


aca:
What is going on 
```
go test -v -run 'TestHistWalk_Accept' -race -count=100

=== RUN   TestHistWalk_Accept
--- PASS: TestHistWalk_Accept (0.03s)
=== RUN   TestHistWalk_Accept_Close
--- PASS: TestHistWalk_Accept_Close (0.03s)
=== RUN   TestHistWalk_Accept
--- PASS: TestHistWalk_Accept (0.03s)
=== RUN   TestHistWalk_Accept_Close
--- PASS: TestHistWalk_Accept_Close (0.03s)
=== RUN   TestHistWalk_Accept
    histwalk_test.go:35: wanted buffer not shown:
        Width = 50, Dot = (0, 9)
        ┌──────────────────────────────────────────────────┐
        │~> echo a$                                        │
        └──────────────────────────────────────────────────┘
    histwalk_test.go:35: Last buffer: Width = 50, Dot = (0, 9)
        ┌──────────────────────────────────────────────────┐
        │~> echo a$                                        │
        │ HISTORY #1 $                                     │
        └──────────────────────────────────────────────────┘
--- FAIL: TestHistWalk_Accept (0.14s)
=== RUN   TestHistWalk_Accept_Close
--- PASS: TestHistWalk_Accept_Close (0.04s)
```

aca:
replaced
	
    evals(f.Evaler, `edit:close-mode`)

with

    f.TTYCtrl.Inject(term.K('[', ui.Ctrl))

Both should produce same result but test fail with first one.

pkg/cli/modes/histwalk.go
```diff
@@ -120,3 +122,15 @@ func (w *histwalk) updatePending() {
 		}
 	})
 }
+
+func (w *histwalk) Accept() error {
```
xiaq, line 126: - The state manipulation _removes_ the prefix, that seems wrong - if I type `echo`, press Up, get `echo foo`, and call this, the buffer will be ` foo`.

- The correct behavior is implemented by `(*tk.CodeAreaState).ApplyPending`, which is used in the `Handle` method on line 62.

- For consistency with `edit:listing:accept`, this should also terminate histwalk mode.

PR #1643, opened 2023-01-16
Fix draft release notes URL; release notes wording

dunsany:
0.18.0 -> 0.19.0
note -> notes
I would also lowercase these links, but I suppose you're not OK with this. :-)

### DIFF
--- a/website/get/prelude.md
+++ b/website/get/prelude.md
@@ -31,7 +31,7 @@ archive:
     <th>arm64</th>
   </tr>
   <tr>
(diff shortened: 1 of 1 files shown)


xiaq:
> I would also lowercase these links, but I suppose you're not OK with this. :-)

Lower case seems more correct here. I guess you didn't do this because of #1628? But that was a different case of homepage headings.

I think the title of individual articles also fall into the title case category (they are literally titles), but the individual "release notes" links are used more like common nouns, not titles of articles in their original form - that is, I would use "release notes" in lower case, but "Elvish 0.18.0 Release Notes" in title case.

PR #1647, opened 2023-01-21
Various `has-key` fixes

### LINKED ISSUE #1646: has-key documentation uses invalid range syntax
This [has-key example](https://elv.sh/ref/builtin.html#has-key) doesn't work for me (with version 0.18):

```
~> has-key [v1 v2] 0:2
▶ $true
```

I get `$false`. Looking at the [language specification](https://elv.sh/ref/language.html#list), I think it should be:

```
~> has-key [v1 v2] 0..2
▶ $true
```

dunsany:
Fixes #1646 and improves tests.

### DIFF
--- a/pkg/eval/builtin_fn_container.d.elv
+++ b/pkg/eval/builtin_fn_container.d.elv
@@ -134,9 +134,9 @@ fn has-value {|container value| }
 # ▶ $true
 # ~> has-key [v1 v2] 2
 # ▶ $false
-# ~> has-key [v1 v2] 0:2
+# ~> has-key [v1 v2] 0..2
 # ▶ $true
-# ~> has-key [v1 v2] 0:3
+# ~> has-key [v1 v2] 0..3
 # ▶ $false
 # ```
 #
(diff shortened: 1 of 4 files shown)


dunsany:
I'm sorry for the number of commits, please squash those when merging.

krader1961:
> I'm sorry for the number of commits, please squash those when merging.

Note that you can squash the changes on your end and force push the resulting change; which is likely to make this easier to review. Alternatively, it is reasonable for you to intend for the changes to be separate. Squash merges are typically done when a pull-request change is modified in response to feedback. Not because the author added additional changes to the original pull-request unrelated to feedback.

Consider your last commit, [9f67e25](https://github.com/elves/elvish/pull/1647/commits/9f67e251ec57512d790d9df98d08db064baa159d), which was clearly meant to fix a bug in your previous commit and should be squashed. My recommendation in a situation like this is to squash the related commits on your end and force push an update to the pull-request to minimize the number of commits needing review (preferably to a single commit).

dunsany:
> Note that you can squash the changes on your end

I'm well aware of that. It's just that currently I only have access to Github on the web, which doesn't offer that function for non-maintainers. Admittedly, I could have postponed making these changes until later, but I figured that one click on maintainer's end isn't too high a price and too much to ask for.

PR #1666, opened 2023-03-05
fix version info ref

chenrui333:


chenrui333:
Also, it might be good to mark the release as `latest` for 0.19.1 cc @xiaq

xiaq:
/facepalm

I updated the tag but forgot to update the version info. I suppose I should skip ahead to 0.19.2 now...

xiaq:
Hmm OK, so you're actually updating the version of the `master` branch.

The mainline is now considered a dev version of the next feature release, which is 0.20.0. I did this in https://github.com/elves/elvish/commit/37f8c4768476e1367d82b7caf55b2962b0516799, so I'll close this now.

Thanks!

chenrui333:
@xiaq thanks for the new release for the version info issue.

QQ, why not mark the latest release as `latest`?

relates to https://github.com/Homebrew/homebrew-core/pull/124871

xiaq:
> QQ, why not mark the latest release as `latest`?

Can you elaborate what that means - do you mean a Git tag? Is that a common practice?

ZhongRuoyu:
All releases are marked as pre-release on https://github.com/elves/elvish/releases. Was that intended? Because pre-releases are not ready for widespread use and package managers generally frown on distributing them.

From https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository:

> Optionally, to notify users that the release is not ready for production and may be unstable, select This is a pre-release.

chenrui333:
ping @xiaq

xiaq:
> All releases are marked as pre-release on https://github.com/elves/elvish/releases. Was that intended? Because pre-releases are not ready for widespread use and package managers generally frown on distributing them.
> 
> From https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository:
> 
> > Optionally, to notify users that the release is not ready for production and may be unstable, select This is a pre-release.

"Pre-release" can mean many things and GitHub's doc doesn't clarify it nearly enough, it's really up to individual packagers to interpret what that means. Elvish is not production-ready in the sense that there are known crashing bugs, and it is also unstable in the sense that there will be backward-incompatibility changes. But it is probably more production-ready and stable than a lot of post-1.0 software...

I don't feel too strongly about this, but since you seem to feel strongly about it (is there a Homebrew policy specifically about the pre-release flag on GitHub?) and it's an easy change I've gone ahead and removed the pre-release flag on the 0.19.2 release.

ZhongRuoyu:
Thanks for your response, @xiaq. You were correct in that the pre-release status can convey different meanings, and your view on this is valid and understandable; I was asking for an confirmation because ...

> is there a Homebrew policy specifically about the pre-release flag on GitHub?

Sort of. We do have an audit rule [here](https://github.com/Homebrew/brew/blob/3e2c713c92ba94caddb11d89b057ab896fb75703/Library/Homebrew/utils/shared_audits.rb#L75-L89) that bans pre-releases unless they are explicitly exempted (which was what we were going to do in https://github.com/Homebrew/homebrew-core/pull/124871/commits/8f009523b6fe36398220fefe2095f2247a4e4533).

> I've gone ahead and removed the pre-release flag on the 0.19.2 release.

Thanks for doing that. In this case, we'll skip adding the audit exception for now. But now that your view on pre-releases is clear, if you decide to continue marking upcoming releases as pre-release (which is fine), we can add it back at future version bumps. After all, this is why we have `github_prerelease_allowlist` exceptions can be allowed.

xiaq:
Thanks for the clarification, I'll continue to _not_ mark releases as pre-releases from this point on.

Issue #1674, opened 2023-03-16
glob expansion should ignore files that cannot be stat()'ed


krader1961:
@hanche noticed that `stat /var/db/DifferentialPrivacy` fails with error "operation not permitted". This is because of a macOS security mechanism known as SIP. This causes Elvish glob expansion to short-circuit when it is unable to stat() the file:

```elvish-transcript
~> cd /var/db
> put *
▶ Accessibility
▶ BootCache.data
▶ BootCache.playlist
▶ BootCaches
▶ CVMS
▶ ConfigurationProfiles
▶ CoreDuet
▶ DetachedSignatures
▶ DiagnosticPipeline
```

This is related to issue #1240 that I fixed almost a year ago. It didn't occur to me at the time that there could be scenarios, short of a broken filesystem, that stat() could fail. The simplest solution is to silently ignore any file that cannot be stat()'ed. However, in the context of a simple glob expansion we should still be able to include the problematic path name in the generated output.

krader1961:
I've got a tentative fix for the glob expansion problem involving files that are protected by [macOS SIP](https://en.wikipedia.org/wiki/System_Integrity_Protection). With my change I can execute `put /var/db/*` and enumerate every file in that directory; including */var/db/DifferentialPrivacy*. It even passes all the existing unit tests. However, this is a pretty big change to the behavior of glob expansion so I want to cogitate on my fix for a few days before opening a pull-request. In particular, it's not at all obvious how to verify the change since this is a corner case that is hard to simulate with the current Elvish unit test infrastructure.

With my fix this now works:

```elvish-transcript
~> put /var/db/*
▶ /var/db/Accessibility
▶ /var/db/BootCache.data
▶ /var/db/BootCaches
▶ /var/db/CVMS
▶ /var/db/ConfigurationProfiles
▶ /var/db/CoreDuet
▶ /var/db/DiagnosticPipeline
▶ /var/db/DiagnosticsReporter
▶ /var/db/DifferentialPrivacy
▶ /var/db/DuetActivityScheduler
...
▶ /var/db/vmware
▶ /var/db/volinfo.database
```

krader1961:
I'll also note that by "pretty big change" I do not mean the number of lines changed by my fix. That is actually quite small. It's a big change in the sense that glob expansion now ignores EPERM errors (on Unix) from `os.Lstat()` and it's unclear what the equivalent is on Windows.

krader1961:
[...]

PR #1678, opened 2023-03-23
Improve consistency of the testutil.Set() pattern

krader1961:
While working on the fix for issue #1674 I noticed that most, but not all, uses of testutil.Set() explicitly modified a pointer to a function being mocked. However, there are a few cases that use a different pattern. This changes those cases so future readers of the code aren't wondering why there are two patterns for using testutil.Set(). As a bonus this allows removing one source file that exists only to enable the alternative pattern being eliminated.

This change also includes a few renaming changes that aren't strictly speaking necessary (e.g., `osExit` => `OSExit`) but are included for consistency with similar cases.


xiaq:
The "dependency variables" shouldn't be exposed as part of the package's API, so they are unexported. It seems that you've changed a lot of them to be exported instead.

Some packages (say `pkg`) have tests that live in a separate `pkg_test` package that can't access unexported symbols, and in that case I've used this pattern:

```go
// In code.go, a file that uses a dependency:
package pkg

import "os"

var sleep = os.Sleep

// In testexports_test.go, a test file in the same package:
package pkg

var Sleep = &sleep

// In some_test.go, a test file in the separate pkg_test package:
package pkg_test

import (
  "testing"
  "src.elv.sh/pkg/testutil"
)

func TestFoo(t *testing.T) {
  testutil.Set(Sleep, fakeSleep)
}
```

I haven't reviewed the entire PR but it seems that at least some of the changes are not respecting this pattern.

Issue #1659, opened 2023-02-18
Expose more Go `os` package functions (e.g., `os.Remove`)


krader1961:
While rewriting my Bash scripts into Elvish to automate tasks, such as building and testing Elvish on all my systems, I noticed I was running this shell command on my primary development system: `rm -f elvish`. It shouldn't be necessary to run an external command to do that. Yet I couldn't find an existing Elvish builtin to do the same thing. So I am opening this issue to propose that the Go `os` package be exposed (not necessarily in its entirety) as an Elvish module. Alternatively, we might want to implement portions of the `os` package in other Elvish modules; e.g., the [`file`](https://elv.sh/ref/file.html) or [`path`](https://elv.sh/ref/path.html) module. I am ambivalent about which approach is preferable since a precedent has already been established by `file:open` wrapping the Go `os.Open` function. We'll probably want to continue building on that precedent by adding a `path:remove` builtin that wraps the Go `os.Remove` function. However, I feel this needs some discussion before anyone creates a pull-request.

krader1961:
Note that resolving this issue also enables resolving issue #1661 since it would allow replacing a lot of external commands used by the `epm` module with Elvish builtins.

krader1961:
This [comment by @xiaq](https://github.com/elves/elvish/pull/1679#issuecomment-1509292632) on my change to implement a `path:remove` has caused me to think more about this issue. There are reasonable arguments for and against deviating from the Go packaging of this functionality in Elvish. After sleeping on the question I am inclined to agree that OS specific functionality, such as removing files or creating directories, should live in an Elvish `os` module. There is a good argument that such functionality applies to file paths and therefore should reside in the Elvish `path` module. But on balance I agree with @xiaq that such functionality is fundamentally tied to interacting with the OS rather than manipulating filesystem path names (such as by concatenating path name components).

TBD is what to do about the existing `path:is-dir` and `path:is-regular` commands. Arguably they should be deprecated and moved to an `os` module. And if they aren't deprecated and moved what does that mean for new commands such as `os:stat` (aka `path:stat`) I want to see implemented?

xiaq:
Simple filesystem functions from `os` now have bindings in Elvish's `os:` module, so closing this. Separate issues should be opened for adding bindings for more functions.

PR #1679, opened 2023-03-27
Implement `path:remove`

krader1961:
I considered adhering more closely to the traditional Unix `rm` and `rmdir` commands. For example, by implementing each independently and supporting the `--force` option. However, I concluded there are insufficient reasons to do so. In particular, the POSIX `--force` (or `-f`) option is most often used solely for its side-effect of making elimination of a non-existent path a non-error. I also don't like that `-f` also fixes permission errors to allow allow removing files. If we decide that feature is needed it should be added under a distinct option (e.g. `&fix-perms`).

Related: #1659

### DIFF
--- a/0.20.0-release-notes.md
+++ b/0.20.0-release-notes.md
@@ -2,6 +2,8 @@ Draft release notes for Elvish 0.20.0.
 
 # Notable new features
 
+-   A new `path:remove` command ([#1659](https://b.elv.sh/1659)).
+
 # Breaking changes
 
 -   The `except` keyword in the `try` command was deprecated since 0.18.0 and is
--- a/pkg/mods/path/path.d.elv
+++ b/pkg/mods/path/path.d.elv
@@ -197,3 +197,36 @@ fn temp-dir {|&dir='' pattern?| }
 # ▶ /some/dir/elvish-RANDOMSTR
 # ```
 fn temp-file {|&dir='' pattern?| }
+
+# Removes one or more path names.
+#
+# If passed zero path names it does nothing; otherwise, it iterates over the
+# list of path names and attempts to remove each one. If a path name does not
+# exist or cannot be removed (perhaps because it is a non-empty directory or
+# permissions do not allow the operation) an exception is raised.
+#
+# Like the traditional Unix `rm` command an error processing a path name does
+# not immediately terminate processing the list of path names. This command
+# attempts to remove the remaining path names. This can result in a "multiple
+# error" exception that documents each path that could not be removed.
+#
+# If the `&ignore-missing` option is set to true a path name that does not
+# exist is silently ignored.
+#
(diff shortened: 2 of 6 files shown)


xiaq:
- I don't see a good reason to deviate from Go, which has `os.Remove` (non-recursive) and `os.RemoveAll` (recursive). It's not clear why you reimplemented recursive removal instead of using `os.RemoveAll`.

- I feel `&ignore-missing` is better named `&missing-ok` (there's precedent in `*[nomatch-ok]`). It is not exactly ambiguous, but it requires a bit of thinking to deduce what "ignore" exactly entails (in this case, ignoring any error from missing files).

I also feel this command shouldn't live in `path`. It's my fault for letting in `is-dir`, `is-regular`, `temp-dir` and `temp-file`, but I'd like to make `path` just about manipulating paths themselves, and create a new `os` module to mirror Go's `os` package. But feel free to leave it there and I'll do the cleanup after merging.

krader1961:
I mostly agree with your feedback, @xiaq. I need to think about whether a new `os` module is preferable. On the one hand I am inclined to agree given issue #1659 that I opened. I even noted this tension on where such commands should live in that issue. The question is how tightly we want to bind Elvish builtins to the Go package that underlies their functionality? This obviously has implications for the [`path:stat`](#1688) command I want to see merged. Should that also be named `os:stat`?

krader1961:
> I don't see a good reason to deviate from Go, which has os.Remove (non-recursive) and os.RemoveAll (recursive). It's not clear why you reimplemented recursive removal instead of using os.RemoveAll.

Because the Go `os.RemoveAll` documentation says this:

> RemoveAll removes path and any children it contains. It removes everything it can but returns the first error it encounters. If the path does not exist, RemoveAll returns nil (no error). If there is an error, it will be of type *PathError.

It is not clear that is the preferable behavior to expose since it elides (hides) some errors. My implementation exposes all errors. I also haven't tested the `os.RemoveAll` behavior. Does it actually delete all files that can be deleted even if there is an error deleting a particular path? The documentation of `os.RemoveAll` is ambiguous on that point.

krader1961:
> I don't see a good reason to deviate from Go, which has os.Remove (non-recursive) and os.RemoveAll (recursive). It's not clear why you reimplemented recursive removal instead of using os.RemoveAll.

The main problem with using `os.RemoveAll` is that it silently ignores a non-existent path. Which we only want to do if the `&missing-ok` (or `&ignore-missing`) is true. So using that Go function requires adding an explicit test whether the path exists. It is simpler to just use `os.Remove` and recurse if the path is a non-empty directory. Not to mention aggregating removal errors which `os.RemoveAll` makes possible while `os.Remove` does not.

I am going to close this P.R. since the implementation is moving from the `path` module to a new `os` module in light of previous feedback. The new P.R. will include a comment why the Go `os.RemoveAll` is not used.

PR #1680, opened 2023-03-27
Remove some external command dependencies

krader1961:
The `epm` module depends unnecessarily on several external commands; e.g., `cat` and `test`.  Remove those dependencies. It also has a bug in its `-first-upper` function due to using the obsolete colon slice index syntax rather than the double period (`..`) syntax. Replace that function with the `str:title` builtin.

Related: 1661

### DIFF
--- a/pkg/mods/epm/epm.elv
+++ b/pkg/mods/epm/epm.elv
@@ -1,3 +1,8 @@
+# Uncomment this line to make it easy to identify dependencies on external
+# commands:
+#
+# pragma unknown-command = disallow
+use path
 use re
 use str
 use platform
@@ -65,7 +70,7 @@ fn dest {|pkg|
 
 # Returns a boolean value indicating whether the given package is installed.
 fn is-installed {|pkg|
-  bool ?(test -e (dest $pkg))
+  path:is-dir (dest $pkg)
 }
 
 fn -package-domain {|pkg|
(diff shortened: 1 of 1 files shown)


krader1961:
@zzamboni, Please review this change.

krader1961:
Note that there are other possible improvements to the `epm` module. For example, when PR#1679 (https://github.com/elves/elvish/pull/1679) is merged the sole use of the `rm` command can be replaced by `path:remove`. Similarly, I intend to introduce a `path:mkdir` command which will allow replacing the existing dependency on the external `mkdir` command with a builtin. Since the `epm` module has had few changes since it was added there are probably other improvements possible but those are outside the scope of this series of changes.

pkg/mods/epm/epm.elv
```diff
@@ -1,3 +1,8 @@
+# Uncomment this line to make it easy to identify dependencies on external
+# commands:
+#
+# pragma unknown-command = disallow
```
xiaq, line 4: It's better to leave this pragma uncommented, and declare all external commands explicitly:

```elvish
pragma unknown-command = disallow
var git~ = $e:git~
var rsync = $e:rsync~
```

krader1961: I can live with that but I chose my approach to make it easy to identify any external commands used by the `epm` module using the documented pragma approach. Your approach makes it a tiny bit harder until something like my proposal in issue #1681 is implemented. Yes, I can just `grep e: **.elv` but that is prone to false positives. However, this has made me wonder if additional info should be provided to my proposed external command callback to make it easier to know the context in which the external command is being run; e.g., which module, script, or interactive context the external command is invoked.

PR #1686, opened 2023-04-04
Implement `path:stat`

krader1961:
I originally started work on this to provide a mechanism to verify the behavior of other unit tests; e.g., a `path:chmod` command I'm working on. Primarily because the external `stat` command has different option names depending on whether you're on a BSD or Linux platform making it hard to use portably. However, since it seems like this could be generally useful I went ahead and implemented support for the metadata that is found on some, not all, platforms.

TBD is providing better support for interpretating the `mode` value.

Related: #1659

### DIFF
--- a/0.20.0-release-notes.md
+++ b/0.20.0-release-notes.md
@@ -2,6 +2,8 @@ Draft release notes for Elvish 0.20.0.
 
 # Notable new features
 
+-   A new `path:stat` command ([#1659](https://b.elv.sh/1659)).
+
 # Breaking changes
 
 -   The `except` keyword in the `try` command was deprecated since 0.18.0 and is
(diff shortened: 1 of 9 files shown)


krader1961:
FYI, I merged this into my branch that implements `path:chmod`. Replacing the external `stat` command with the `path:stat` implemented by this change makes those unit tests portable between GNU Linux and BSD platforms. That is a weak argument for adding `path:stat` but I believe it is useful even outside of Elvish unit tests. For example, it could be used to implement a pure Elvish version of the `ls` command.

krader1961:
Closing for the moment since I realized some changes, specifically to the publicly visible mode value, will make it easier to use. Especially when composed with a `path:chmod` to set the permissions the same as a reference file; e.g., `path:chmod (path:stat ref-file)[mode] a-file`. Will reopen in a day or two after I've made those changes.

PR #1688, opened 2023-04-13
Implement `path:stat`

krader1961:
In theory the `path:stat` command introduced by this change allows an Elvish implementation of something like the Unix `ls`, or Windows `dir`, command. More importantly, it provides a more general API than builtins like `path:is-dir` and `path:is-regular`. Making it possible to test many other path attributes. I didn't deprecate those related commands because they are probably used often enough it doesn't make sense to deprecate them.

Related: #1659

### DIFF
--- a/0.20.0-release-notes.md
+++ b/0.20.0-release-notes.md
@@ -2,6 +2,8 @@ Draft release notes for Elvish 0.20.0.
 
 # Notable new features
 
+-   A new `path:stat` command ([#1659](https://b.elv.sh/1659)).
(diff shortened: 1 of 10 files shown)


krader1961:
I should also mention that my original motivation for this command was to make it easier to unit test related commands such as the `path:chmod` I have in my pipeline of `path` module changes to resolve #1659.

krader1961:
Closing until the `os` module is introduced after which I will reinstate this change.

PR #1689, opened 2023-04-13
Implement `str:fields`

rsteube:


pkg/mods/str/str.d.elv
```diff
@@ -54,6 +54,19 @@ fn count {|str substr| }
...
+# Splits `$str` around each instance of one or more consecutive white space
+# characters.
+#
+# ```elvish-transcript
+# ~> str:split "lorem ipsum   dolor"
+# ▶ lorem
+# ▶ ipsum
+# ▶ dolor
+# ~> str:split "   "
+# ```
```
krader1961, line 67: Add a "See also" line to str:split and vice-versa. That will make it easier to discover.

krader1961:
LGTM. Make it so (ST:NG Picard reference).

PR #1695, opened 2023-04-25
`pprint` (and `repr`) now sorts map keys

krader1961:
This change is more invasive than I expected because the existing Elvish value comparison code was in `pkg/eval` rather than `pkg/eval/vals`. This moves the value comparison logic into the `vals` package so it can be used by packges other than `eval`. For example, the vals package `repr` implementation. Not to mention other Elvish modules which may want to compare Elvish values.

I also decided to introduce a `strutil.Dedent` function to make it easier to write multiline string literals in unit tests while retaining readability.

Related: 1495

### DIFF
--- a/0.20.0-release-notes.md
+++ b/0.20.0-release-notes.md
@@ -7,6 +7,11 @@ Draft release notes for Elvish 0.20.0.
 -   The language server now supports showing the documentation of builtin
     functions and variables on hover ([#1684](https://b.elv.sh/1684)).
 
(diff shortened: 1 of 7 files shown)


krader1961:
Closing this PR since I've decided to also include sorting psuedo-map (struct-map) keys and the easiest way to do that is to change the type used to hold the keys for sorting to hold key/value pairs. I'll open a new PR in a couple of hours when I've finished that work.

Issue #1599, opened 2022-08-11
Should `order` support heterogenous value types?


krader1961:
The [discussion](#1025) about the `keys` function implicitly sorting its output resulted in a decision not to do so. However, since the keys of an Elvish map do not have to be homogenous that implies the `order` function should support heterogenous types if we want examples like this to be well defined without raising an exception:

```
> var m = [&a=b &(num 1)=2]
> order [(keys $m)]
Exception: bad value: inputs to "compare" or "order" must be comparable values, but is uncomparable values
```

krader1961:
Obviously, supporting heterogenous value types requires defining an ordering among the types. Which is going to be arbitrary but possible to define. For example, we could define type `bool` < `num` < `string` < `list`. The main issue is the cognitive cost of such an arbitrary ordering. Still, if we're going to allow heterogenous keys for maps it seems to me the `order` (and `compare`) command has to allow heterogenous values in order to get well defined behavior from doing `keys $map | order`.

P.S., At the moment `compare` (and thus `order`) doesn't allow comparing maps. Should it? What does it mean to compare two maps vis-a-vis ordering? The reason I ask is that the Elvish map implementation allows using a maps as keys which begs the question what does `keys $map | order` mean if the keys for `$map` are themselves maps?

krader1961:
Note that Go doesn't support comparing maps. See https://go.dev/ref/spec#Comparison_operators. You can't declare a map that uses a map as a key. This is an error in Go: `var m map[map]bool`. While Elvish isn't Go, and Elvish maps are ostensibly immutable, I still think Elvish should disallow maps as keys in a map. I can't think of a single valid use case and it is easy to inadvertently use a map as a key given the Elvish grammar.

dunsany:
> if we want examples like this to be well defined without raising an exception

Why would we want that?

krader1961:
> What is so special in the keys $map | order bit, that we should, in your opinion, make sure that it never fails?

Precisely because Elvish map keys are not currently required to be homogenous. I've argued [elsewhere](#1574) that they should be homogenous. In which case this proposal can be rejected. But if we're going to continue allowing heterogenous keys then `keys $map | order` should be well defined and not raise an exception.

dunsany:
I don't follow you here. Your argument, as I currently understand it, goes like this:

> K: `order` should support heterogenous value types.
> D: Why?
> K: Because `keys $map | order` shouldn't ever fail.
> D: Why?
[...]

PR #1696, opened 2023-04-26
`pprint` (and `repr`) now sorts map keys

krader1961:
This change is more invasive than I expected because the existing Elvish value comparison code was in `pkg/eval` rather than `pkg/eval/vals`. This moves the value comparison logic into the `vals` package so it can be used by packges other than `eval`. For example, the vals package `repr` implementation. Not to mention other Elvish modules which may want to compare Elvish values.

I also decided to introduce a `strutil.Dedent` function to make it easier to write multiline string literals in unit tests to improve readability of the expected value.

This also sorts the keys of pseudo (struct) maps for consistency with mutable maps. This arguably makes the output of pretty-printing pseudo maps less friendly since a struct map definition typically orders its members in a deliberate fashion (unlike a regular map). However, it is more important that the output of printing the output of regular and pseudo maps (via `repr` or `pprint`) be consistent with regard to the order of their keys.

Related: 1495


krader1961:
Closing in light of my recent thoughts regarding a total ordering of all Elvish value types (see https://github.com/elves/elvish/issues/1599). I'm going to rework these two changes to make the ordering of different types well defined. Thus making the implicit sorting of map keys by `pprint` and `repr` well defined rather than ill-defined.

Issue #1495, opened 2022-02-25
Improve the output of pprint


xiaq:
The current implementation of pprint is pretty naive, in particular when it comes to printing nested data structures (lists and maps).

Concretely, there are several things to improve:

* [ ] Sort the map keys.
* [ ] Ensure that the map values are actually aligned (current implementation just inserts Tabs and hope they align).
* [ ] Print lists and maps on one line when there is enough space to do so.

Citing an example from https://github.com/elves/elvish/issues/1494#issuecomment-1051282865:

> This is the current output of `pprint $unix:rlimits` on my machine:
> 
> ```
> [
>  &nofile=	[
>    &cur=	(num 256)
>   ]
>  &fsize=	[&]
>  &nproc=	[
>    &max=	(num 2666)
>    &cur=	(num 2666)
>   ]
>  &memlock=	[&]
>  &cpu=	[&]
>  &core=	[
>    &cur=	(num 0)
>   ]
>  &stack=	[
>    &max=	(num 67092480)
>    &cur=	(num 8372224)
>   ]
>  &rss=	[&]
>  &data=	[&]
> ]
> ```
> 
> But ideally it should look like this instead:
> 
> ```
> [
>  &cpu=     [&]
>  &core=    [&cur=(num 0)]
>  &data=    [&]
>  &fsize=   [&]
>  &memlock= [&]
>  &nofile=  [&cur=(num 256)]
>  &nproc=   [&cur=(num 2666) &max=(num 2666)]
>  &rss=     [&]
>  &stack=   [&cur=(num 8372224) &max=(num 67092480)]
> ]
> ```

krader1961:
I'm a +1 for improving the output of `pprint` but I don't think that eliminates the need for dedicated commands for formatting the output of specific pieces of data such as UNIX resource limits or Elvish abbreviations. The `pprint` command is fundamentally targeted at Elvish **programmers** not shell **users**.

krader1961:
For the benefit of anyone stumbling upon this issue I'll note that I've published a module that provides a human friendly command for displaying and manipulating UNIX resource limits. See https://github.com/krader1961/elvish-lib/ (which has been added to the [Awesome Elvish](https://github.com/elves/awesome-elvish) page. For example, on my primary macOS server:

```
> ulimit
resource      current    maximum
========== ========== ==========
core                0  unlimited
cpu         unlimited  unlimited
data        unlimited  unlimited
fsize       unlimited  unlimited
memlock     unlimited  unlimited
[...]

Issue #1656, opened 2023-02-07
Allow escaping newlines in double-quoted strings to remove the EOL char(s)


krader1961:
There was a question in the chat channels about the Elvish equivalent for a POSIX here-doc illustrated by this example:

```
cat <<EOF
line 1
line 2
EOF
```

AFAIK, the only straightforward equivalent is this:

```
echo 'line 1
line 2' | cat
```

Or this:

```
echo ^
'line 1
line 2' | cat
```

If double-quoted strings allowed backslash to precede the EOL char(s) then this would be valid:

```
print "\
line 1
line 2
" | cat
```

Alternatively, a triple-quoted string would be aesthetically pleasing and arguably easier to use:

```
echo '''
line 1
line 2
''' | cat
```

Here the first and last EOL char(s) are elided by the triple-quoting mechanism. This, however, is not backwards compatible given the existing rule for interpreting single-quotes inside a single-quoted string. There is a similar problem for double-quoted strings. But, if we're willing to special-case triple quotes at the start and end of a line (when the sequence is otherwise a token with no adjacent chars other than EOL), then this is unlikely to break any existing code. It also allows us to introduce an additional filtering rule that removes  leading whitespace common to all lines. For example:

```
elvish> echo '''
                 line 1
                   line 2
                 line 3
             ''' | cat
line 1
  line 2
line 3
```

Eliding EOL char(s) by prefacing them with a backslash in a double-quoted string is backward compatible and should be straightforward to implement. I can't see any downsides to supporting it and believe it should be implemented. I am ambivalent about the triple-quoted string idea but include it here to see what others think about the idea and whether it sparks any other proposals for making it easier, and clearer, to implement the equivalent of a POSIX here-doc in Elvish. I am also opposed to introducing the POSIX here-doc (`cat <<WORD\n...\nWORD`) syntax (even if its quirks are avoided) because I don't think it adds enough value relative to `echo | cat` (or `print | cat`) to justify the feature.

See also issue #1261.

krader1961:
[...]

PR #1698, opened 2023-05-05
`pprint` and `repr` now sort map keys

krader1961:
This change is more invasive than I expected because the existing Elvish value comparison code was in `pkg/eval`.  This moves the value comparison logic into `pkg/eval/vals` so it can be used by packges other than `eval`. For example, the vals package `repr` implementation as well as other Elvish modules which may want to compare Elvish values.

I also decided to introduce a `strutil.Dedent` function to make it easier to write multiline string literals in unit tests to improve readability of the expected value.

This also sorts the keys of pseudo (struct) maps for consistency with mutable maps. This arguably makes the output of pretty-printing pseudo maps less friendly since a struct map definition typically orders its members in a deliberate fashion (unlike a regular map). However, it is more important that the output of printing the output of regular and pseudo maps (via `repr` or `pprint`) be consistent with regard to the order of their keys.

Related: #1495

### DIFF
--- a/0.20.0-release-notes.md
+++ b/0.20.0-release-notes.md
@@ -7,6 +7,14 @@ Draft release notes for Elvish 0.20.0.
 -   The language server now supports showing the documentation of builtin
     functions and variables on hover ([#1684](https://b.elv.sh/1684)).
 
+# Notable fixes and enhancements
+
+-   The `pprint` and `repr` commands now sort map keys
+    ([#1495](https://b.elv.sh/1495)).
+
+-   The `compare` and `order` commands have a new `&types` option to order
+    otherwise uncomparable types.
+
 # Breaking changes
 
 -   The `except` keyword in the `try` command was deprecated since 0.18.0 and is
--- a/pkg/eval/builtin_fn_io.d.elv
+++ b/pkg/eval/builtin_fn_io.d.elv
@@ -192,6 +192,9 @@ fn echo {|&sep=' ' @value| }
 # ]
 # ```
 #
+# Maps (both mutable and [pseudo](language.html#pseudo-map)) are output with
+# their keys sorted in the same order as `keys $map | order &types`.
+#
 # The output format is subject to change.
 #
 # See also [`repr`]().
@@ -204,6 +207,9 @@ fn pprint {|@value| }
 # ~> repr [foo 'lorem ipsum'] "aha\n"
 # [foo 'lorem ipsum'] "aha\n"
 # ```
+#
+# Maps (both mutable and [pseudo](language.html#pseudo-map)) are output with
+# their keys sorted in the same order as `keys $map | order &types`.
 #
 # See also [`pprint`]().
 #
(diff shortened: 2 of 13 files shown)


krader1961:
This also modifies the `compare` and `order` commands by introducing a `&types` option so they can be used to compare (order) otherwise uncomparable types. This is for symmetry with the behavior of `repr` and `pprint` when outputting maps. I have had conflicting opinions about this. My original change had the `compare` and `order` commands always impose a total ordering on uncomparable types. I have decided it makes more sense for users of `compare` and `order` to have to explicitly ask for ordering based on the type of otherwise uncomparable values while having `repr` and `pprint` implicitly order based on value and type.

krader1961:
Regarding the introduction of a `Dedent` function see my [comment](https://github.com/elves/elvish/issues/1656#issuecomment-1537037343) regarding new literal string syntax. The `Dedent` functionality is useful outside of its use in the unit tests of this pull-request.

xiaq:
The `CmpTypes` function does not handle all the builtin types, which includes nil, exceptions, functions and more "esoteric" ones like styled strings.

Printing maps with ordered keys does not require a total ordering of all values. You can group values by types, and only sort the groups that are actually sortable.

> This also sorts the keys of pseudo (struct) maps for consistency with mutable maps. This arguably makes the output of pretty-printing pseudo maps less friendly since a struct map definition typically orders its members in a deliberate fashion (unlike a regular map). However, it is more important that the output of printing the output of regular and pseudo maps (via `repr` or `pprint`) be consistent with regard to the order of their keys.

Pseudo maps and maps are already different. I would rather pseudo maps keep their original key order.

> Regarding the introduction of a `Dedent` function see my [comment](https://github.com/elves/elvish/issues/1656#issuecomment-1537037343) regarding new literal string syntax. The `Dedent` functionality is useful outside of its use in the unit tests of this pull-request.

A `testutil.Dedent` function already exists.

xiaq:
If you'd like to continue contributing this change, please break the change that moves `cmp` to `pkg/eval/vals` into a separate PR. It's a bit hard to review the PR as it is today.

You can also leave this feature to me as I have a pretty good idea of how I'd like it to work at this point.

krader1961:
> The CmpTypes function does not handle all the builtin types, which includes nil, exceptions, functions and more "esoteric" ones like styled strings.

Neither does the existing `cmp` implementation. The obvious solution is to use reflection when one of the types is unknown. If both types are the same just treat the values as equal. This has the value of not having to explicitly enumerate all types since there isn't a natural ordering for types such as exceptions and functions. Yet it allows ordering sequences that contain such types without having to deal with an exception from the `compare` or `order` command (assuming the types are homogenous). A similar solution is available to the `CmpTypes` implementation; only in that case compare the type names to derive an ordering between the types.

Closing this since I'll push an update to PR #1700 that implements the idea in the previous paragraph for the `vals.Cmp` function.

PR #1699, opened 2023-05-08
feat(snippets): add basic snippets

EmilySeville7cfg:
[This](https://github.com/nushell/vscode-nushell-lang/blob/main/snippets/convention.md) naming convention for snippets has been used. The goal of this naming convention to make snippet names predictable and easily memorizable.

### DIFF
--- a/vscode/snippets/snippets.json
+++ b/vscode/snippets/snippets.json
@@ -1,0 +1,72 @@
+{
+    "lambda": {
+        "prefix": "lambda",
+        "description": "home",
+        "body": "{|${1:arguments}| ${2:body}}"
+    },
+    "var builtin": {
+        "prefix": "var",
+        "description": "\"var\" invocation",
+        "body": "var ${1:name} = ${2:value}"
+    },
+    "if builtin": {
+        "prefix": "if",
+        "description": "\"if\" invocation",
+        "body": [
+            "if (${1:condition}) {",
+            "\t${2:body}",
+            "}"
+        ]
+    },
+    "if else builtin": {
+        "prefix": "if-else",
+        "description": "\"if\" invocation",
+        "body": [
+            "if (${1:condition}) {",
+            "\t${2:body}",
+            "} else {",
+            "\t${3:body}",
+            "}"
+        ]
+    },
+    "while builtin": {
+        "prefix": "while",
+        "description": "\"while\" invocation",
+        "body": [
+            "while (${1:condition}) {",
+            "\t${2:body}",
+            "}"
+        ]
+    },
+    "while else builtin": {
+        "prefix": "while-else",
+        "description": "\"while\" invocation",
+        "body": [
+            "while (${1:condition}) {",
+            "\t${2:body}",
+            "} else {",
+            "\t${3:body}",
+            "}"
+        ]
+    },
+    "for builtin": {
+        "prefix": "for",
+        "description": "\"for\" invocation",
+        "body": [
+            "for ${1:variable} ${2:container} {",
+            "\t${3:body}",
+            "}"
+        ]
+    },
+    "for else builtin": {
+        "prefix": "for-else",
+        "description": "\"for\" invocation",
+        "body": [
+            "for ${1:variable} ${2:container} {",
+            "\t${3:body}",
+            "} else {",
+            "\t${4:body}",
+            "}"
+        ]
+    }
+}
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -89,7 +89,13 @@
                     "description": "Trace communication between VS Code and the Elvish language server. Trace messages are shown in the Elvish Language Server output channel."
                 }
             }
-        }
+        },
+        "snippets": [
+            {
+                "language": "elvish",
+                "path": "./snippets/snippets.json"
+            }
+        ]
     },
     "scripts": {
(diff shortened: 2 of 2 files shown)


EmilySeville7cfg:
@xiaq, there are failing tests, I have no idea why (I've checked CI failing details).

krader1961:
> @xiaq, there are failing tests, I have no idea why (I've checked CI failing details).

It looks like the CI failure is due to a change by @xiaq to my implementation of the "`peach &num-workers`" feature. That change has introduced non-deterministic behavior of the `peach` related unit tests. My implementation of those unit tests was deliberately very conservative to avoid this type of non-deterministic failure. In other words, the CI test failure has nothing to do with your change. The failure is because a new unit test, introduced prior to your change, is inherently non-deterministic.

krader1961:
@EmilySeville7cfg, you wrote

> [Nushell](https://github.com/nushell/vscode-nushell-lang/blob/main/snippets/convention.md) naming convention for snippets has been used.

I don't use VScode or Nushell but have a passing familiarity with both. It is unclear from your commit comment is why this change is being made. It may help to expand on the "Nushell naming convention for snippets has been used." rationale in the commit comment.

EmilySeville7cfg:
@krader1961, the point of my comment is to mention the fact that the way snippets are named is the same as for Nushell. The goal of this naming convention to make snippet names predictable and easily memorizable. As I wrote this convention for Nushell already, I decided just to refer it here to make you know how snippets are named and why. ;)

krader1961:
> This convention can be used not just for Nushell, for many ones. I just try to keep things standardized across several shell extensions for VS Code as I use several shells.

Okay, but that rationale should be included in the commit comment. My point is that someone like myself, who has some idea what both VScode and Nushell are, without having used them, won't be able to immediately understand the point of this change. Let alone someone who has no familiarity with either technology. Reducing the ambiguity of the commit comment is a net positive.

vscode/snippets/snippets.json
```diff
@@ -0,0 +1,72 @@
+{
+    "lambda": {
+        "prefix": "lambda",
+        "description": "home",
```
xiaq, line 4: This description (`"home"`) doesn't make sense - copy-paste error?

I'm a bit confused by how the the object key (`"lambda"`) and the description (`"home"`) differ - they seem to be both human-readable description strings. AFAIK it works like this in VS Code:

- If the description is defined, it is used as the description. The object key doesn't seem to be used anywhere
- If the description is not defined, the object key is used as the description.

So it seems to me that just using the object key as the description is sufficient. I'll remove the description fields.

vscode/snippets/snippets.json
```diff
@@ -0,0 +1,72 @@
...
+    "var builtin": {
+        "prefix": "var",
+        "description": "\"var\" invocation",
+        "body": "var ${1:name} = ${2:value}"
+    },
+    "if builtin": {
+        "prefix": "if",
+        "description": "\"if\" invocation",
+        "body": [
+            "if (${1:condition}) {",
```
xiaq, line 16: The `( )` around the conditions is not part of the `if` command. When you see something like:

```elvish
if (not $foo) {
  ...
}
```

The `( )` is an [output capture](https://elv.sh/ref/language.html#output-capture) expression.

vscode/snippets/snippets.json
```diff
@@ -0,0 +1,72 @@
...
+            "\t${2:body}",
+            "}"
+        ]
+    },
+    "if else builtin": {
+        "prefix": "if-else",
+        "description": "\"if\" invocation",
+        "body": [
+            "if (${1:condition}) {",
```
xiaq, line 25: Same

vscode/snippets/snippets.json
```diff
@@ -0,0 +1,72 @@
...
+            "} else {",
+            "\t${3:body}",
+            "}"
+        ]
+    },
+    "while builtin": {
+        "prefix": "while",
+        "description": "\"while\" invocation",
+        "body": [
+            "while (${1:condition}) {",
```
xiaq, line 36: Same

vscode/snippets/snippets.json
```diff
@@ -0,0 +1,72 @@
...
+            "\t${2:body}",
+            "}"
+        ]
+    },
+    "while else builtin": {
+        "prefix": "while-else",
+        "description": "\"while\" invocation",
+        "body": [
+            "while (${1:condition}) {",
```
xiaq, line 45: Same

xiaq:
There are some things I would change, but I'll merge this and apply those changes as a followup commit.

xiaq:
> @krader1961, the point of my comment is to mention the fact that the way snippets are named is the same as for Nushell. The goal of this naming convention to make snippet names predictable and easily memorizable. As I wrote this convention for Nushell already, I decided just to refer it here to make you know how snippets are named and why. ;) This convention can be used not just for Nushell, for many ones. I just try to keep things standardized across several shell extensions for VS Code as I use several shells.

I think for most users the only thing that matters for these snippets is the prefix, so I wouldn't spend much effort trying to keep the descriptions standardized. Nu also has a lot more different types of syntax than Elvish - for Elvish the snippets are all just commands - so there's less need to adopt a rigorous naming system. As long as future contributors to this snippet file keep things consistent it's fine.

PR #1700, opened 2023-05-14
Move the `cmp` function into the `vals` package

krader1961:
This moves the value comparison logic into `pkg/eval/vals` so it can be used by packges other than `eval`. For example, the vals package `repr` implementation as well as other Elvish modules which may want to compare Elvish values. This is a baby step to augmenting all the code which order values, or might want to (e.g., the `pprint` command), to also handle ordering otherwise uncomparable types.

Related #1495

### DIFF
--- a/pkg/eval/builtin_fn_pred.go
+++ b/pkg/eval/builtin_fn_pred.go
@@ -1,9 +1,6 @@
 package eval
 
 import (
-	"math"
-	"math/big"
-
 	"src.elv.sh/pkg/eval/errs"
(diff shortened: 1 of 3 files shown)


krader1961:
I'm going to open a new pull-request with the enhancement I discussed in PR #1698 to treat the values of types not explicitly recognized as being equal. Thus eliminating one source of unexpected exceptions when doing things like `compare $nil $nil` or `put ?(fail y) ?(fail x) | order` since the types or obviously comparable while the current behavior says they are not comparable.

PR #1701, opened 2023-05-16
Move the `cmp` function into the `vals` package

krader1961:
Move the value comparison logic into `pkg/eval/vals` so it can be used by packges other than `eval`. For example, the vals package `repr` implementation as well as other Elvish modules which may want to compare Elvish values.

This also augments the comparison logic to handle types that do not have an obvious, explicitly handled, ordering. For example, Elvish exceptions, functions, and styled text. If the types are the same then the values are considered equal. This means you can now do things like `compare $nil $nil` or `put ?(fail y) ?(fail x) | order` without raising an exception.

Related #1495

### DIFF
--- a/pkg/eval/vals/compare.go
+++ b/pkg/eval/vals/compare.go
@@ -1,0 +1,145 @@
+package vals
+
+import (
+	"math"
+	"math/big"
+	"reflect"
+)
+
+type ordering uint8
+
+const (
+	CmpLess ordering = iota
+	CmpEqual
+	CmpMore
+	CmpUncomparable
+)
+
+// Cmp compares two arbitrary Elvish types and returns a value that indicates
+// whether the values are not comparable or the first value is less, equal, or
+// greater than the second value.
+func Cmp(a, b any) ordering {
+	switch a := a.(type) {
+	case bool:
+		if b, ok := b.(bool); ok {
+			switch {
+			case a == b:
+				return CmpEqual
+			//lint:ignore S1002 using booleans as values, not conditions
+			case a == false: // b == true is implicit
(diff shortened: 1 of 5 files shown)


krader1961:
I force pushed an update because I realized there were some minor tweaks to my this change that I made in a subsequent change that really belongs in this change. These tweaks to the code are all cosmetic but avoid introducing the same tweaks in subsequent changes I have in my pipeline.

pkg/eval/vals/compare.go
```diff
@@ -0,0 +1,145 @@
...
+				return CmpMore
+			}
+		}
+	// TODO: Figure out a sane solution for comparing maps (both regular and
+	// struct). Obviously the definition of "less than" is ill-defined for maps
+	// but it would be nice if we had a solution similar to that for the List
+	// case above. One approach is to treat each map as an ordered list of [key
+	// value] pairs and use the List comparison logic above. Yes, that behavior
+	// is hard to document and expensive but has the virtue of providing an
+	// unambiguous means of sorting maps. For now we assume all maps are equal.
```
xiaq, line 90: Please remove this. As I have said elsewhere, I don't want to make maps comparable.

xiaq (changes requested):

xiaq:
I did the move in https://github.com/elves/elvish/commit/1e249ed7fbcab6c5c0e3d18e523af0eac620c69e so I'll close this now.

As a general point of advice, it's easier for me to merge PRs when the PR doesn't mix simple changes (like moving a function to a different package, which is what I requested) and bigger changes that impact the existing semantics (like changing how `compare` behaves).

PR #1703, opened 2023-05-20
Implement a well defined ordering among types

krader1961:
This builds on the previous changes to implement a total ordering by replacing the dependency on the address at which various types are defined (which is effectively random) with an explicit ordering. For types without an explicit ordering it uses an ordering based on the order in which each unknown type is compared. This makes the ordering predictable based on the execution of Elvish code without regard to unpredictable factors such as the layout of data structures in the Elvish binary.

Note that the use of mechanisms like the `peach` command can introduce unpredictable ordering of unknown value types since this change handles unknown types at run time and thus, in theory, will produce different results when Elvish code that invokes `compare &total` or `order &total` is run concurrently.  Whereas the prior logic is not affected by the parallelism of Elvish code.  Nonetheless, I maintain that the predictability of ordering heterogenous values introduced by this change is a net positive since the new behavior will be predictable and reproducible (as a practical matter) while the prior logic (based on the random layout of data structures in the Elvish binary) is neither.

Related #1495

### DIFF
--- a/pkg/eval/builtin_fn_stream_test.go
+++ b/pkg/eval/builtin_fn_stream_test.go
@@ -131,18 +131,17 @@ func TestOrder(t *testing.T) {
 		That("put 10 1 5 2 | order &reverse &key={|v| num $v }").
 			Puts("10", "5", "2", "1"),
 
-		// &total orders the values into groups of different types, and sorts
-		// the groups themselves. Test that without assuming the relative order
-		// between numbers and strings.
-		That(
-			"put (num 3/2) (num 1) c (num 2) a | order &total | var li = [(all)]",
-			"put $li",
-			"has-value [[a c (num 1) (num 3/2) (num 2)] [(num 1) (num 3/2) (num 2) a c]] $li").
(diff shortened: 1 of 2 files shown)


xiaq:
Having a central place that enumerates all types makes it necessary to add every new type there, but there's no way to guarantee that this has happened when a new type is added. The approach in this PR requires Elvish's runtime type system to be _closed_ when it is in fact an _open_ system.

A predictable ordering between types is a good property, but the cost outweighs the benefit.

If there is a compelling use case for a well-defined ordering between types (which doesn't seem to be the case now), there is a simpler, less error-prone way to do this: compare the fully qualified names of the [reflect.Type](https://pkg.go.dev/reflect#Type) (using `t.PkgPath() + "." + t.Name()`) values.

xiaq:
Also: I might decide on future to actually make the runtime type system closed, at which point imposing a well-defined total order among types will make more sense. That approach has some big advantages but also some big disadvantages, so I'm not committing to it happening any time soon, or indeed at all.

PR #1711, opened 2023-07-16
Make Concat return a new globPattern.

### LINKED ISSUE #1245: Incorrect results when combining braced list with wildcard
```
~> touch xy.{u v w} xz.w
~> put x*
⮕ xy.u
⮕ xy.v
⮕ xy.w
⮕ xz.w
~> put x*.{u v w}
⮕ xy.w
⮕ xz.w
⮕ xy.w
⮕ xz.w
⮕ xy.w
⮕ xz.w
```

The cause of the problem is probably thus: the internal glob data structure is being inadvertently shared between the expanded `x*.u`, `x*.v` and `x*.w`, and since the last one was last evaluated it overwrote the previous two data structures.

mikelward:
I don't know if this is fully sufficient, but it seems to fix the specific test case mentioned.

Fixes #1245

### DIFF
--- a/pkg/eval/glob.go
+++ b/pkg/eval/glob.go
@@ -136,8 +136,11 @@ func (gp globPattern) Index(k any) (any, error) {
 func (gp globPattern) Concat(v any) (any, error) {
 	switch rhs := v.(type) {
 	case string:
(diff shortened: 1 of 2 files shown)


xiaq:
Thanks!

Issue #695, opened 2018-06-07
`styled` should detect non-tty output and disable colors


zzamboni:
Most programs which produce color output include detection of when the output is not to a tty (e.g. being redirected to a file) and automatically disable color codes in this case.

For Elvish scripts, it would be nice if this detection could be embedded in the new `styled` command, so that the feature comes "for free".

For completeness, there should be a `&force` option which makes it possible to always generate the color codes, even when output is being redirected.

krader1961:
I was writing a `type` function and wanted to produce styled output when outputting to a tty but otherwise produce non-styled output. However, that isn't really possible given the current design of the styled mechanism. It currently emits a `type Text` object to the value stream:

```
> styled wtf red
▶ (ui:text (ui:text-segment wtf &fg-color=red))
```

You have to coerce it to a string to get the terminal byte sequence that produces styled output as in these examples:

```
> echo (styled wtf red)
> styled wtf red | to-lines
```

And therein lies the problem. We can't simply have the `styled` command test whether its output is to a tty. Nor is it sufficient to modify the `Text.String` method to test if it's writing to a tty since it might be used in a command capture context or a pipeline where the ultimate destination is a tty.

This means that the proposed enhancement, as written, appears infeasible. The only practical solution would appear to be to modify the `styled` command to emit the string without styling if the shell, as a whole, is not interactive. That is, assume that if the shell is interactive then we want styled output; otherwise, produce plain text output. Someone can use the proposed `&force` option to produce styled output even in non-interactive shells.

PR #1714, opened 2023-08-03
support monochrome mode

tw4452852:
Should help with #695


xiaq:
No color mode shouldn't affect what is actually stored in styled texts; it should only affect how they are printed. If users write code that uses styled texts but doesn't actually print them, it shouldn't be affected by NO_COLOR.

I am also slightly torn on use of a global flag, but this one feels justified. It gives the user no way to opt out of the behavior, but that may actually be a good thing.

tw4452852:
@xiaq How about the updated version?

PR #1717, opened 2023-08-13
Fix the docs for ordering heterogenous types

krader1961:


xiaq:
The new wording is not correct; for example, `compare &total [&a=b] [&b=c]` outputs 0, as is described by the current wording.

As I explained in the user group, cases like `compare &total a b` (both arguments being strings) are covered by a clause before, and the reader would only think it outputs 0 if they don't read from top to bottom. It probably makes sense to make the outline of the algorithm a bit more explicit by using an ordered list though. 62795d907d99d5bf6100a68ea8466bdb7811426f did this.

fbde7b12080cea8349bb527539252f4dc7eabd61 fixed the `&total` option for `order`.

PR #1730, opened 2023-10-27
Implement `os:chmod`

krader1961:
Related: #1659

### DIFF
--- a/0.20.0-release-notes.md
+++ b/0.20.0-release-notes.md
@@ -29,6 +29,8 @@ Draft release notes for Elvish 0.20.0.
     variable. Builtin UI elements as well as styled texts will no have colors if
     it is set and non-empty.
 
+-   A new `os:chmod` command ([#1659](https://b.elv.sh/1659)).
+
 # Notable bugfixes
 
 -   `has-value $li $v` now works correctly when `$li` is a list and `$v` is a
--- a/pkg/eval/errs/errs.go
+++ b/pkg/eval/errs/errs.go
@@ -2,11 +2,18 @@
 package errs
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 
 	"src.elv.sh/pkg/parse"
 )
+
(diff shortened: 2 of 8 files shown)


krader1961:
There are other uses of generic errors that, arguably, should use the `errs.Generic` this change introduces. Such as `ttySetupErr` in `TestReadCode_AbortsWhenTTYSetupReturnsError` (*pkg/cli/app_test.go*). I considered those cases too far outside the scope of this change.

xiaq:
Thanks for the contribution. My thoughts:

- I don't like the introduction of `errs.Generic`; it already has two different use cases in this PR - in _test_ code where the _identity_ of the error matters, and in non-test code that only cares about the _presence_ of the error. The former should continue to use a unique error value (so that it can't accidentally match _other_ places that use `errs.Generic`), and the latter should be refactored to use a boolean instead to keep track of whether the conversion succeeded.

  In general, duplication is preferable than ambiguous abstraction.

- The `os:stat` command already has a convention for representing non-permission non-type bits using a separate `special-modes` list. This API is easier to use than the traditional bit flag API, at least from Elvish code (Elvish currently lacks bitops, which makes this worse although that should be fixed regardless). It is less efficient than bit flags, which is a trade-off that makes sense for Elvish.

  The `os:chmod` command should use that pattern too: the signature should be `fn chmod {|&special-modes=[] perm file| }`.

- I was a bit ambivalent about whether `os:stat` should return the permission bits as a typed number or a string containing an octal, but ultimately I felt that as a programming language the former was the only sensible choice, so I'd rather `os:chmod` be consistent in this regard.

  One thing I'll introduce later is the ability for numbers to have a _default format_. The perm bits in the output of `os:stat` can then be displayed as base-8 by default (but still with the `0o` prefix so that there is no ambiguity). This is only relevant for converting a number to a string, and not relevant for the conversion in the other direction though - so `os:chmod` should still accept numbers like any other builtin.

Thanks again for the contribution but I won't be merging this PR; I have prepared my implementation that I will push shortly.

Issue #1282, opened 2021-04-04
Cluttered completion when window is small


rsteube:
This is a bit weird as the other completers work and just this one gives problems (and only when the window is small, when it's larger it works).
Should be a side effect from one of the descriptions, but i can't see what is causing it at the moment.

```elvish
edit:completion:arg-completer[mv] = [@arg]{
    set opts = '[{"Value":"-b","Display":"-b (like --backup but does not accept an argument)"},{"Value":"--backup","Display":"--backup (make a backup of each existing destination file)"},{"Value":"--context","Display":"--context (set SELinux security context of destination file to default type)"},{"Value":"-Z","Display":"-Z (set SELinux security context of destination file to default type)"},{"Value":"--force","Display":"--force (do not prompt before overwriting)"},{"Value":"-f","Display":"-f (do not prompt before overwriting)"},{"Value":"--help","Display":"--help (display this help and exit)"},{"Value":"--interactive","Display":"--interactive (prompt before overwrite)"},{"Value":"-i","Display":"-i (prompt before overwrite)"},{"Value":"--no-clobber","Display":"--no-clobber (do not overwrite an existing file)"},{"Value":"-n","Display":"-n (do not overwrite an existing file)"},{"Value":"--no-target-directory","Display":"--no-target-directory (treat DEST as a normal file)"},{"Value":"-T","Display":"-T (treat DEST as a normal file)"},{"Value":"--strip-trailing-slashes","Display":"--strip-trailing-slashes (remove any trailing slashes from each SOURCE argument)"},{"Value":"--suffix","Display":"--suffix (override the usual backup suffix)"},{"Value":"-S","Display":"-S (override the usual backup suffix)"},{"Value":"--target-directory","Display":"--target-directory (move all SOURCE arguments into DIRECTORY)"},{"Value":"-t","Display":"-t (move all SOURCE arguments into DIRECTORY)"},{"Value":"--update","Display":"--update (move only when the SOURCE file is newer than the destination file or when the destination file is missing)"},{"Value":"-u","Display":"-u (move only when the SOURCE file is newer than the destination file or when the destination file is missing)"},{"Value":"--verbose","Display":"--verbose (explain what is being done)"},{"Value":"-v","Display":"-v (explain what is being done)"},{"Value":"--version","Display":"--version (output version information and exit)"}]'
    echo $opts | from-json | all (one) | each [c]{ edit:complex-candidate $c[Value] &display=$c[Display] }
}
```

[...]

PR #1736, opened 2023-12-06
pkg/cli/tk/listbox_window.go: Fix divide by zero

rsteube:
Fix `divide by zero` when resizing terminal.

Steps to reproduce:
1. invoke completion menu
2. reduce height of terminal window

related https://github.com/elves/elvish/issues/1282


xiaq:
This does fix the crash, but it means that when the available height is 1 all that is visible is the scrollbar, which isn't ideal. Better to show one row of actual content instead. I'm preparing that fix.

xiaq:
I've implemented the fix I described so I'll close this PR. Thanks for the contribution.

PR #1741, opened 2024-01-02
Changed os reference documentation to use follow-symlink instead of old eval-symlinks

aftix:


xiaq:
Thanks. The Markdown needs re-wrapping (which the CI checks) so I'll merge it to a different branch first.

xiaq:
Huh, just realized that I could also just edit your commit. Anyway, this is now merged into `master`.

PR #1752, opened 2024-01-13
Provide more info for an invalid doc link

### LINKED ISSUE #1749: The `website/tools/check-rellinks.py` program has a surprising quirk regarding links that end in a slash
I was cleaning up my change to add a `learn/useful-customizations.html` page before opening a pull-request and was pulling my hair out because the `make check-rellinks` build target was failing thusly despite the link being valid and working in the resulting web page when traversed via my browser:

```
Found broken links:
learn/useful-customizations.html
    ../ref/edit.html/#edit:histlist:start
```

After more than an hour of debugging I finally realized it was because of the trailing slash. I have no objection to requiring such links not include a trailing slash but it would be nice if the tool checking those links provided a more useful explanation for why the link is "broken" since, in this instance, it is valid. Alternatively, augment the *website/tools/check-rellinks.py* program to allow a trailing slash in such URLs.

krader1961:
Make it more obvious why a documentation reference link is invalid when the URL path incorrectly includes a trailing slash. This would have saved me an hour of debugging when I was working on a change that included a reference like this one:

    [`edit:histlist:start`](../ref/edit.html/#edit:histlist:start)

Fixes #1749

### DIFF
--- a/website/tools/check-rellinks.py
+++ b/website/tools/check-rellinks.py
@@ -47,6 +47,13 @@ def main(args):
                 path = filename
             else:
                 if os.path.splitext(path)[1] == '':
+                    # Tell people who inadvertently append a trailing slash to
+                    # a URL not meant to refer to a index.html file they have
+                    # made a mistake.
+                    # See https://github.com/elves/elvish/issues/1749.
+                    if path.endswith(".html/"):
+                        print(f'WARNING: A trailing slash implies you are referring to {path}index.html.')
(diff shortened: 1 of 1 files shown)


xiaq:
This check should be generalized to check for all bogus slashes as well as missing slashes. I have posted a bit more explanation in https://github.com/elves/elvish/issues/1749#issuecomment-1904223636.

PR #1753, opened 2024-01-15
Add comment why os:chmod tests might fail

### LINKED ISSUE #1748: master branch os:chmod tests fail on FreeBSD
Commit cf9ec15fa68e02799190582e539c338d4f9a4a55 added an `os:chmod` command modeled on [my change](https://github.com/elves/elvish/pull/1730) that was rejected. The problem is that the replacement for my change fails on FreeBSD:

```
--- FAIL: TestChmod (0.02s)
    --- FAIL: TestChmod/os:mkdir_d;_os:chmod_&special-modes=[setuid_setgid_sticky]_0o400_d;_put_(os:stat_d)[special-modes] (0.02s)
        os_unix_test.go:14: got value out (-want +got):
              []string{
            -   (
            -           """
            -           [
            -            setuid
            -            setgid
            -            sticky
            -           ]
            -           """
            -   ),
              }
        os_unix_test.go:14: unexpected exception
        os_unix_test.go:14: got: *fs.PathError: chmod d: operation not permitted
        os_unix_test.go:14: stack trace: []string{"os:chmod &special-modes=[setuid setgid sticky] 0o400 d"}
        os_unix_test.go:14: want: <nil>
FAIL
FAIL    src.elv.sh/pkg/mods/os  0.096s
```

My change explicitly special-cased FreeBSD in the `TestChmod` unit test:

```
+   // These two tests fail on FreeBSD with "inappropriate file type or format"
+   // and "operation not permitted" respectively. I don't know why but will
+   // trust that if they pass on Linux and macOS then the mode bits are
+   // correctly mapped.
+   if runtime.GOOS != "freebsd" {
+       TestWithEvalerSetup(t, useOS,
+           That("echo >f; os:chmod (num 0o1567) f; var s = (os:stat f); "+
+               "put $s[perm] (count $s[special-modes]) $s[special-modes][0]").
+               Puts(0o567, 1, "sticky"),
+           That("echo >g; os:chmod (num 0o2444) g; var s = (os:stat g); "+
+               "put $s[perm] (count $s[special-modes]) $s[special-modes][0]").
+               Puts(0o444, 1, "setgid"),
+       )
+   }
```

That special-case is probably wrong since it papers over what might be a serious problem. Nonetheless, it allowed the unit tests to pass on all supported platforms, including FreeBSD, while the current master branch fails. Either my unit test special-case for FreeBSD should be added or a proper fix introduced for this behavior on FreeBSD.

krader1961:
Fixes #1748

### DIFF
--- a/pkg/mods/os/os_unix_test.go
+++ b/pkg/mods/os/os_unix_test.go
@@ -11,12 +11,16 @@ import (
 )
 
 func TestChmod(t *testing.T) {
+	// Note: On FreeBSD the /tmp directory is usually mounted with the "suiddir"
+	// option. This causes new directories, such as created by this unit test,
+	// to inherit the /tmp group owner. If the account running this unit test is
+	// not a member of that group (typically "wheel") you'll get test failures.
 	TestWithSetup(t, func(t *testing.T, ev *eval.Evaler) {
 		testutil.InTempDir(t)
 		useOS(ev)
 	},
-		That(`os:mkdir d; os:chmod 0o400 d; put (os:stat d)[perm]`).Puts(0o400),
-		That(`os:mkdir d; os:chmod &special-modes=[setuid setgid sticky] 0o400 d; put (os:stat d)[special-modes]`).
+		That(`os:mkdir d1; os:chmod 0o400 d1; put (os:stat d1)[perm]`).Puts(0o400),
+		That(`os:mkdir d2; os:chmod &special-modes=[setuid setgid sticky] 0o400 d2; put (os:stat d2)[special-modes]`).
 			Puts(vals.MakeList("setuid", "setgid", "sticky")),
 	)
 }


xiaq:
Thanks for bringing the issue to my attention.

The PR has some issues:

- The change to the actual test code is unnecessary - it seems to be based on the assumption that subsequence test cases share the same directory, so you'll need to ensure that the directory names are different. But that's not the case: when using `evaltest.TestWithSetup` the setup is run for every individual test, so each test gets run in its own temp directory.
- The comment you added also seems wrong - this has nothing to do with how tmpfs is mounted, this is just how mkdir behaves on FreeBSD, as explained in https://github.com/golang/go/issues/19596.

We should also simply fix the issue - as shown in https://go-review.googlesource.com/c/go/+/41430/5/src/cmd/go/internal/work/build_test.go, doing chgrp on the parent of the directory that one wishes to call chmod on does the trick.

krader1961:
The problem on my FreeBSD VM was because when I installed the OS it didn't put me in the "wheel" group. And the */tmp* mount point was owned by the "wheel" group with the "sticky" attribute. I fixed it by adding my account to that group. Doing chgrp on */tmp* is not a viable solution to this failure mode. The comment my change introduced was meant to only provide a hint to a future developer who might wonder why this unit test was failing.

The functional change to the unit test was only meant to make it obvious that the two tests operated on different directories. There are quite a few unit tests which do the same thing because it doesn't require the reader to recognize that `TestWithSetup` runs each test in a distinct temp dir and thus reduces the cognitive load when interpreting the test code.

krader1961:
I see your fix changes the ownership of the created temp directory rather than the parent of the created temp dir. I misunderstood your previous comment as implying the ownership of */tmp* should be changed. Still, the comment in your fix is not very clear about the nature of the "Work around a quirk of FreeBSD's mkdir". I removed the `if runtime.GOOS == "freebsd" {` statement and the unit test passed on every platform I use for testing these things (macOS, Windows, FreeBSD, and several Linux distros). In other words, there is no reason to predicate the `os.Chown` on the OS being FreeBSD, AFAICT. The chown is a nop on non FreeBSD platforms.

xiaq:
Sure, done in 464cfb64c64a9f61fbb1fe48ec43078860492960.

PR #1756, opened 2024-01-16
Detect invalid default flag value in `flag:call`

### LINKED ISSUE #1755: flag::call does not recognize options with default $nil
flag::call fails to find options with a default of $nil. Any other default works. A minimal example:

```
~> flag:call {|&port=$nil| echo $port } [-port 1]
Exception: flag provided but not defined: -port
[tty 23]:1:1: flag:call {|&port=$nil| echo $port } [-port 1]
~> flag:call {|&port=0| echo $port } [-port 1]
1
```

krader1961:
Fixes #1755

### DIFF
--- a/pkg/mods/flag/flag.go
+++ b/pkg/mods/flag/flag.go
@@ -30,7 +30,9 @@ func call(fm *eval.Frame, fn *eval.Closure, argsVal vals.List) error {
 	fs := newFlagSet("")
 	for i, name := range fn.OptNames {
 		value := fn.OptDefaults[i]
-		addFlag(fs, name, value, "")
+		if err := addFlag(fs, name, value, ""); err != nil {
+			return err
+		}
 	}
 	err = fs.Parse(args)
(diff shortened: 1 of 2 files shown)


pkg/mods/flag/flag.go
```diff
@@ -30,7 +30,9 @@ func call(fm *eval.Frame, fn *eval.Closure, argsVal vals.List) error {
 	fs := newFlagSet("")
 	for i, name := range fn.OptNames {
 		value := fn.OptDefaults[i]
-		addFlag(fs, name, value, "")
+		if err := addFlag(fs, name, value, ""); err != nil {
```
xiaq, line 33: This pattern is popular in certain Go projects as it saves one line of code, but IMO `if` with an init clause should only be used when the init clause is only used to calculate something that's only used in the condition or the if body. In this case `addFlag` has side effects that's relevant outside the `if` block.

PR #1758, opened 2024-01-18
Align `pprint` map values

krader1961:
This modifies the behavior of `pprint` to vertically align map values using the maximum key width rather than using a tab character which often results in vertically misaligned values. It also changes the indent from one to two spaces to improve readability of the output.

This also replaces the existing text dedent function with a less rigid implementation. The new implementation makes it easier to write indented text in a natural manner that (a) should reliably produce the expected result, and (b) not catasrophically fail if the input has odd indentation. In the latter case the expectation is that using the dedented text will detect it has an unexpected representation. But the new implementation also produces what I would consider correct output where the old implementation would panic. Ultimately I think this dedent functionality should be exposed by the `str` module but didn't want to go that far with this change.

Related: #1495


krader1961:
I'm reasonably happy with this change. Consider the output of `pprint $edit:insert:binding` with, and without, this change. Obviously there are some details (such as the amount of indentation) which is debatable. I would be happy to change the default indentation in order to see this merged. Even better would be making the indentation configurable with a reasonable default (whether 1, 2, or 4). But that can be done by a separate change. I've tested this with a wide variety of cases and have not found one where the old behavior is preferable to the new behavior introduced by this change.

krader1961:
I also have another change that builds on this one to compact the output when a structure fits on a single line as defined by the terminal width. That change is even more invasive than this one. I think it is worthwhile but the added complexity might not be justified solely to reduce the number of lines of output. But that's a discussion for a separate pull-request.

xiaq:
Please use pkg/wcwidth rather than introducing a new dependency. I haven't reviewed the bulk of the PR and will do so after that is done.

krader1961:
> Please use pkg/wcwidth rather than introducing a new dependency. I haven't reviewed the bulk of the PR and will do so after that is done.

It's odd that I didn't notice that package. I'll compare the two packages and use the stdlib version if, as expected, it is of comparable quality.

krader1961:
> It's odd that I didn't notice that package. I'll compare the two packages and use the stdlib version if, as expected, it is of comparable quality.

I suspect I chose the third-party package because the google searches I performed (in the past and now) list the github.com/mattn/go-runewidth near the top of the search results and the stdlib package is not in the first page of search results. It's also interesting that [pkg.go.dev](https://pkg.go.dev/github.com/mattn/go-runewidth?tab=importedby) results for the mattn/go-runewidth lists 1891 imports. Whereas [pkg.go.dev](https://pkg.go.dev/src.elv.sh/pkg/wcwidth?tab=importedby) lists zero uses of the stdlib `wcwidth` package. That, obviously, has to be a bug of pkg.go.dev since it is improbable the stdlib `wcwidth` package has zero uses, but may contribute to why so many projects use the mattn/go-runewidth package.

krader1961:
The mattn/go-runewidth package also provides more functionality than the stdlib wcwidth package, but that functionality isn't needed by Elvish. So that's not a good reason for using it. I expect that I will conclude the stdlib package does what Elvish needs and will update my PR in the next day or two to use the stdlib `wcwidth`.

krader1961:
/facepalm

I just realized the `wcwidth` package referred to by @xiaq in their previous comment is part of the Elvish project, not the Go stdlib. It seems unlikely that the Elvish `wcwidth` package is more correct than an actively maintained third-party package. I am willing to use the Elvish `wcwidth` package but I don't see a good reason for a home grown implementation when there is at least one good third-party implementation that is more likely to produce correct results.

krader1961:
FYI, I threw together the program below to compare the width of all code points reported by the two packages.

```go
package main

import (
        "fmt"

        "github.com/mattn/go-runewidth"
        "src.elv.sh/pkg/wcwidth"
)

func main() {
        for k := 0; k < 0x40000; k++ {
                r := runewidth.RuneWidth(rune(k))
                n := wcwidth.OfRune(rune(k))
                if r != n {
                        fmt.Printf("%06X %d %d\n", k, r, n)
                }
        }
}
```

The first two codepoints the two packages don't agree on are [U+00AD](https://codepoints.net/U+00AD?lang=en) and [U+0487](https://codepoints.net/U+0487?lang=en). The go-runewidth reports zero for both (the correct value for Elvish's purposes) and wcwidth reports one. That program reports 11928 codepoints where the two packages disagree.

Obviously I did not hand check all differences. But interestingly the third difference was U+0591 which is a nonspacing mark so I would expect the width to be zero. Which is what wcwidth reports while runewidth reports one. On the other hand in that range are code points like U+0600 which has a width of one which runewidth correctly reports as one while wcwidth reports zero.

I picked a few more differences at random, such as U+135D, another nonspacing mark, which runewidth gets right and wcwidth gets wrong.

Codepoint U+4DC1 is width one according to runewidth but two according to wcwidth. Yet, when I display that char in iTerm2 it appears to be width one. And U+27B0 which displays as width two in my terminal and runewidth gets right while wcwidth reports one.

Clearly both packages have code points where they report the wrong width. But on balance it looks like the go-runewidth package produces more accurate results. I'm happy to update the PR to use the existing Elvish `wcwidth` package, @xiaq, but I think that long term this isn't a wheel we should reinvent. Your thoughts?

xiaq:
There isn't really such a thing as "correct" result in the context of a wcwidth implementation for Elvish. What Elvish needs to do is to guess what the terminal thinks a character is, and every terminal will disagree on that, so there is no single implementation that will be correct all of the time. Instead pkg/wcwidth offers a function for the user to [override the wcwidth](https://pkg.go.dev/src.elv.sh/pkg/wcwidth#Override) (exposed as `-override-wcwidth`), so that the user can manually patch Elvish's wcwidth implementation to agree with whatever terminal they are using. I can see people to create packages just contain a bunch of those -`override-wcwidth` calls for each different terminal emulator.

I took a look at https://github.com/mattn/go-runewidth and I am not confident that depending on it is a good idea. There are a few PRs that seem pretty important that haven't been merged or closed. I have a lot of respect for mattn, but he seems to be busy person and maintaining this package may not be terribly high on his list of priorities.

There are also some technical choices that I don't quite agree with, like the idea of relying on Unicode properties for character width when [Unicode TR11](https://unicode.org/reports/tr11/) explicitly advises against it:

> Note: The East_Asian_Width property is not intended for use by modern terminal emulators without appropriate tailoring on a case-by-case basis. Such terminal emulators need a way to resolve the halfwidth/fullwidth dichotomy that is necessary for such environments, but the East_Asian_Width property does not provide an off-the-shelf solution for all situations. The growing repertoire of the Unicode Standard has long exceeded the bounds of East Asian legacy character encodings, and terminal emulations often need to be customized to support edge cases and for changes in typographical behavior over time.

The pkg/wcwidth implementation could use updates for widths that all modern terminals agree on. Switching to go-runewidth is not the solution here.

---

Now for something else. You might perceive this as me "berating" you, but I feel it's necessary to give you some advice for how to be a productive contributor:

- Do a bit more research before rushing to post. You are entitled to your opinion and welcome to post your learning process on, say, your personal blog, but posting each step of how you figure something out here is distracting. I have to spend time reading your comments - a lot of which are made relevant by your later comments - and other people trying to contribute to the discussion will also be confused, especially when you often post with a strong sense of conviction.

  I try my best to write well-researched and considered comments. If I don't have something of value to say, I don't comment at all. I don't expect everyone in the community to do the same, and I deliberately refrain from policing other people in order to maintain an open atmosphere of discussion. However, as you comment on Elvish's issue tracker a lot, and those comments are often in response to mine, my bar for you will be proportionally higher.

- Don't assume anything you don't understand in the Elvish codebase is a bug waiting to be fixed. Yes, some of such places are corners I've neglected, but at the same time, a lot of places are not. There are nuances, unobvious tradeoffs, messy history, researches I'd done, all of which I would have written down if I had enough time, but I don't always have enough time. If you engage constructively, I will gladly add comments and make code easier to understand, but bear in mind I don't owe you any of that. But if you start off by assuming that I don't know what I am doing, you may sometimes be right, but all it does is making it harder for me to engage in the conversation.

I give you these pieces of advice with the assumption that your goal is to be a helpful member of the community and get your contribution accepted and recognized quickly. If your goal is to prove your intellectual superiority over other people (including me) or use this space as your personal blog, this is not the place.

I don't agree with everything in Rich Hickey's essay [Open Source is Not About You](https://gist.github.com/richhickey/1563cddea1002958f96e7ba9519972d9), but I think it's useful for resetting your expectations. Avery Pennarun's essay [The Gift of It's Your Problem Now](https://apenwarr.ca/log/20211229) I mostly agree with.

I hope this is last conversation of this depth I need to have with you on how to engage with me and the rest of the community. It's getting a bit tiring for me - I'd rather spend time writing code and sketching designs when I do get to spend time on Elvish - but I trust that you can do better.

---

Finally, back to the PR itself. It seems that aside from introducing the runewidth-go dependency and some other unrelated changes (which I'd refer you to my comments above on not assuming I don't know what I am doing), the relevant change is to `MapReprBuilder.WritePair`. However, this implementation is over-simplistic:

- It doesn't handle the case where the key prints to multiple lines.

- When the keys get long enough it'd be better to just print the values on a new line instead.

I'll probably spend some time later implementing this feature and will close this PR now.

PR #1760, opened 2024-01-25
support big num for base cmd

tw4452852:


pkg/eval/builtin_fn_str.go
```diff
@@ -49,14 +49,21 @@ func toString(fm *Frame, args ...any) error {
 // greater than 36.
 var ErrBadBase = errors.New("bad base")
 
-func base(fm *Frame, b int, nums ...int) error {
+// ErrInvalidInput is thrown by the "base" builtin if the input num is invalid
+var ErrInvalidInput = errors.New("invalid input")
+
+func base(fm *Frame, b int, nums ...string) error {
```
krader1961, line 55: Did you try this: `base 16 (math:pow 2 16)`? Num objects are not automatically converted to strings when calling a builtin, but num objects are automatically converted to int if they fit.

pkg/eval/builtin_fn_str_test.go
```diff
@@ -39,6 +39,8 @@ func TestBase(t *testing.T) {
 		That(`base 16 42 233`).Puts("2a", "e9"),
 		That(`base 1 1`).Throws(ErrBadBase),
 		That(`base 37 10`).Throws(ErrBadBase),
+		That(`base 16 18446744073709551607`).Puts("fffffffffffffff7"),
```
krader1961, line 42: The unit tests also need at least one case that explicitly tests passing a num object. Such as

```
That(`base (num 16) (math:pow 2 8)`).Puts( "100"),
```

pkg/eval/builtin_fn_str.go
krader1961: Hint: Replace `string` with `vals.Num`. Then add a type switch to allow only int and bigint and reject rational and float64.

pkg/eval/builtin_fn_str_test.go
krader1961: I should have been clearer that my prior comment was not about your change. A test of the sort I proposed should already be present in the project. Had it been present it would have alerted you to the fact your change is incorrect.

krader1961: And, obviously, there should be tests added that explicitly test non-integer values throw the expected exception. Which should already be present but should be added if you want to pursue this change. Such as something like the following:

```
That(`base 8 (num 1/8)`).Throws(something),
```

That's because the current implementation implicitly detects that problem and throws a "wrong type" exception since `num 1/8` is not an integer. But your change requires an explicit test for that scenario since your modified change (per my previous comments) would otherwise accept at `num 1/8` object.

pkg/eval/builtin_fn_str.go
tw4452852: Thanks for the tip. Adapted in the new version.

pkg/eval/builtin_fn_str_test.go
tw4452852: Thanks for the comments. I've added more test cases in the new version.

pkg/eval/builtin_fn_str.go
```diff
@@ -49,14 +49,31 @@ func toString(fm *Frame, args ...any) error {
...
 	out := fm.ValueOutput()
 	for _, num := range nums {
-		err := out.Put(strconv.FormatInt(int64(num), b))
+		bigInt := new(big.Int)
+		switch num := num.(type) {
+		case int:
+			bigInt.SetInt64(int64(num))
+		case *big.Int:
+			bigInt.Set(num)
+		case string:
```
krader1961, line 68: This case is impossible. The `vals.Num` constraint will cause the Elvish runtime to convert any string into a `vals.Num` if it is a valid number. It's helpful to run `make cover` to catch such mistakes.

pkg/eval/builtin_fn_str_test.go
```diff
@@ -39,6 +39,9 @@ func TestBase(t *testing.T) {
...
+		That(`base 10 "0xfffffffffffffff7"`).Puts("18446744073709551607"),
+		That(`base 8 (num 1/8)`).Throws(ErrInvalidInput),
```
krader1961, line 44: You should also add the test I proposed (or something very similar): `That(`base (num 16) (math:pow 2 8)`).Puts( "100")`.

pkg/eval/builtin_fn_str.go
tw4452852: Done.

pkg/eval/builtin_fn_str_test.go
tw4452852: Done.

Issue #474, opened 2017-08-27
Tab completion should not expand ~ or variable


xiaq:
Typing `echo ~/<Tab>` will expand the home directory; this is not desirable.

mqudsi:
Thank you. I considered it a bug for myself, but presumed it was intentional.

krader1961:
There was a recent question about this behavior on the chat channels. My reply to that question is below....

I never use Tab completion when the path includes a leading tilde so I've never been bothered by this. That's because I only do cd ~/someplace when the path is very short; e.g., cd ~ or cd ~/log. Pretty much for anything else I use location mode since I've usually already navigated to the directory and it's in my location history. Still, I agree that path completion shouldn't substitute the actual home dir for the tilde. However, I think variables should be expanded by Tab completion. If you don't want the command history to reflect the actual value at the time the command was executed then don't use Tab completion. If you do type something like cd ${A-var}/subdir followed by Tab then I would expect the variable to be expanded.

krader1961:
Note that expansion of a leading tilde by Tab completion is counterproductive given that user home directories almost never change. Expansion by Tab completion replaces a single character with many characters that are unlikely to change and thus has close to zero value. It is preferable to leave the expansion of the leading tilde to the code that interprets the path.

haxscramper:
Additionally, expanding tilde makes it harder to copy-paste things from the history to documentation, for sending code to the others as example. This is more of a niche case and this functionality can be made into opt-out instead of changing the default behavior.

krader1961:
> This is more of a niche case and this functionality can be made into opt-out instead of changing the default behavior.

While backward incompatible changes should be avoided this is one case where I think doing so is acceptable. Because (a) it only affects interactive use and thus won't break any existing Elvish code, and (b) I'm confident that most Elvish users will prefer the new, non-tilde expansion, Tab completion behavior.

iandol:
From the chat server:

> Hi, if I write cd /home/user/te TAB ... the input is autocompleted to cd /home/user/temp, so far so good.
> But, if I write cd ~/te TAB ... the "COMPLETING argument" menu is shown with only one option. > It's a bit annoying.
> Any way to avoid this?

xiaq replied: fix this bug, so hope remains!

PR #1764, opened 2024-02-07
Elminate implicit `cd`

### LINKED ISSUE #1099: `/` (division command) invoked with no args does `cd /`
While chipping away at issue #1062 to improve unit test coverage I noticed this block of code:

https://github.com/elves/elvish/blob/3300c81c8bf5c6754c34414c1f315625e2c90a11/pkg/eval/builtin_fn_num.go#L281-L289

I noticed it because no unit test exercises the `return fm.Chdir("/")` statement and I was surprised that module did not have 100% coverage.

I could add a test for that case but I think the feature should be removed. Note that the only place I can find that documents the "implicit cd" feature is for the [`/` command](https://elv.sh/ref/builtin.html#----).

Note that the current behavior can result in silent bugs:

```
> pwd
/Users/krader
> l = []
> / $@l
> pwd
/
```

That chdir should not have occurred.

krader1961:
Replace the CLI implicit `cd` behavior with location mode support for literal directory paths.

Fixes #1099

### DIFF
--- a/0.21.0-release-notes.md
+++ b/0.21.0-release-notes.md
@@ -32,3 +32,6 @@
         }
     }
     ```
+
+-   The implicit cd feature has been removed and replaced by the ability to type
+    a literal path in location mode ([#1099](https://b.elv.sh/1099)).
--- a/go.sum
+++ b/go.sum
@@ -1,6 +1,7 @@
 github.com/creack/pty v1.1.21 h1:1/QdRyBaHHJP61QkWMXlOIBfsgdDeeKfK8SYVUWJKf0=
 github.com/creack/pty v1.1.21/go.mod h1:MOBLtS5ELjhRRrroQr9kyvTxUAFNvYEK993ew/Vr4O4=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
+github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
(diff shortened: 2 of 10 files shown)


krader1961:
Force push to fix the conflict with the change from the 0.20.0-release-notes.md to the 0.21.0-release-notes.md file since this PR was initially created. No other changes were made.

xiaq:
After playing with this a bit, it feels non-ideal:

- You can't trigger a tab completion from location mode - I realize that I've already mentioned this in https://github.com/elves/elvish/issues/1099#issuecomment-673727104, but now that I've played with it seems that I almost _always_ use tab completion with implicit cd.

- `~` doesn't work in location mode (this should be fixable though)

I now feel literal cd via location mode may not be the best idea. Maybe adding a completer for `cd` to only show directories and fixing #474 would be a better UX, albeit with slightly more keypresses.

As for the removal of implicit cd, I'd rather deprecate it first and only remove it after the next release.

The headings in `tour.md` were chosen intentionally to reflect the general functionality rather than Elvish-specific terms, please don't change them.

Thanks for implementing this, and sorry that I've changed my mind. I'll close this PR and add deprecation for implicit cd.

krader1961:
> I now feel literal cd via location mode may not be the best idea. Maybe adding a completer for cd to only show directories and fixing https://github.com/elves/elvish/issues/474 would be a better UX, albeit with slightly more keypresses.

That is actually what I would prefer, @xiaq. I only did this as an experiment and to gain more experience writing interactive unit tests. I've always maintained that implict `cd` isn't worth saving just two keystrokes compared to this pull-request. However, if those two keystrokes are worth eliminating it might be worth exploring how to make tab completion work in location mode. Also, the fact my change doesn't handle a `~/` prefix was an oversight and could probably be handled easily.

PR #1769, opened 2024-03-03
Pre exit hook

xofyarg:
allow user to customize pre-exit hook

It is useful when user wants to clean up things that associated with
shell session.

Provide a CallHook internal API for the common usage.

### DIFF
--- a/pkg/eval/eval.go
+++ b/pkg/eval/eval.go
@@ -91,7 +91,11 @@ type Evaler struct {
 // NewEvaler creates a new Evaler.
 func NewEvaler() *Evaler {
 	builtin := builtinNs.Ns()
-	beforeChdirElvish, afterChdirElvish := vals.EmptyList, vals.EmptyList
+
+	newListVar := func(l vals.List) vars.PtrVar { return vars.FromPtr(&l) }
+	preExitHookElvish := newListVar(vals.EmptyList)
+	beforeChdirElvish := newListVar(vals.EmptyList)
+	afterChdirElvish := newListVar(vals.EmptyList)
 
 	ev := &Evaler{
 		global:  new(Ns),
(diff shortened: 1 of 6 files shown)


pkg/eval/hook.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+	"os"
+
+	"src.elv.sh/pkg/diag"
+	"src.elv.sh/pkg/eval/vals"
+)
+
+// CallHook runs all the functions in the list "hook", with "args". If
+// "cfg" is not specified, current eval's ports will be used.
```
xiaq, line 12: > current eval's ports will be used

This seems to contradict the implementation, which uses standard files to build ports. Evalers don't actually have associated ports (any more - they probably used to).

Also a TODO is warranted here, since none of the hooks should actually use the standard files:

 - Chdir hooks should use the frame from where `Evaler.Chdir` is called. This is the more problematic case as `cd` can easily be called with redirections.

 - Editor lifecycle should use the TTY passed to initialize it - this is less problematic because the editor is hooked to the stderr, but it makes testing impossible.

pkg/eval/eval.go
```diff
@@ -91,7 +91,11 @@ type Evaler struct {
 // NewEvaler creates a new Evaler.
 func NewEvaler() *Evaler {
 	builtin := builtinNs.Ns()
-	beforeChdirElvish, afterChdirElvish := vals.EmptyList, vals.EmptyList
+
+	newListVar := func(l vals.List) vars.PtrVar { return vars.FromPtr(&l) }
+	preExitHookElvish := newListVar(vals.EmptyList)
+	beforeChdirElvish := newListVar(vals.EmptyList)
+	afterChdirElvish := newListVar(vals.EmptyList)
```
xiaq, line 98: I don't like this change, it makes the `CallHook` calls below ugly: we need an unnecessary `.Get()`, and we lose type safety and have to use a type assertion.

xiaq (approved):
I'll merge this to a branch and add a fixup commit.

xofyarg:
> I'll merge this to a branch and add a fixup commit.

Thanks for taking over!

pkg/eval/eval.go
xofyarg: I thought the var needs to be passed around by reference while protected by a mutex. Is there a general rule for this, like which variables need to be put behind mutex?

xiaq: Oh that's a great point I missed, I'll keep your change.

Yes all _global_ variables need to be mutex-protected, and these are all global variables.

PR #1782, opened 2024-04-01
try / else docs

### LINKED ISSUE #1734: try {...} else {...} is documented but doesn't work
The language specification (https://elv.sh/ref/language.html#try), section 8.9 (Exception control: Try), item 3, states:

> 3. If no exception occurs and `else` is present, `else-block` is executed. Example:
> 
> ```
> ~> try { nop } else { echo well }
> well
> ```

But this does not work. I get:

```
% try { nop } else { echo well }
Compilation error: try must be followed by a catch block or a finally block
  [tty 8]:1:1: try { nop } else { echo well }
```

Unfortunate, because this would be a useful feature if it worked.

I am using the latest release, 0.19.2.

Duncan3142:
Clarify code examples regarding required `catch` or `finally` when using `else`

Fixes #1734

### DIFF
--- a/website/ref/language.md
+++ b/website/ref/language.md
@@ -2102,8 +2102,17 @@ This control structure behaves as follows:
     Example:
 
     ```elvish-transcript
-    ~> try { nop } else { echo well }
+    ~> try { nop } catch { echo bad } else { echo well }
     well
+    ```
+
+    **Note**: `else` requires a `catch` or `finally` to be present. The
+    following code is invalid:
+
+    ```elvish-transcript
+    ~> try { nop } else { echo well }
+    Compilation error: try must be followed by a catch block or a finally block
+      [tty 1]:1:1-30: try { nop } else { echo well }
     ```
 
 4.  If `finally-block` is present, it is executed. Examples:


Duncan3142:
As an observation, the syntax for `try/else/finally` is a bit confusing as a new user.

The syntax implies the `else` happens instead of the `try`

Also, is there a need for such a construct

For example, instead of

```
try {
  cat log-file > /tmp/log-snapshot
} else {
  aws s3api put-object --bucket logs --key logs/latest --body /tmp/log-snapshot
} finally { 
  rm -f /tmp/log-snapshot
}
```

would it not be sufficient to simply do

```
try {
  cat log-file > /tmp/log-snapshot
  aws s3api put-object --bucket logs --key logs/latest --body /tmp/log-snapshot
} finally { 
  rm -f /tmp/log-snapshot
}
```

If there is a valid use case for `try/else/finally` I'd be happy to add it to the above PR

website/ref/language.md
```diff
@@ -2102,10 +2102,19 @@ This control structure behaves as follows:
...
 
+    **Note**: `else` requires a `catch` or `finally` to be present. The
+    following code is invalid:
+
+    ```elvish-transcript
+    ~> try { nop } else { echo well }
+    Compilation error: try must be followed by a catch block or a finally block
+      [tty 1]:1:1-30: try { nop } else { echo well }
+    ```
+
```
xiaq, line 2117: The text after the bullet list explains that either `catch` or `finally` must be present, so I'd move this explanation to be part of that.

xiaq:
> As an observation, the syntax for `try/else/finally` is a bit confusing as a new user.
> 
> The syntax implies the `else` happens instead of the `try`
> 
> Also, is there a need for such a construct?
> 
> For example, instead of
> 
> ```
> try {
>   cat log-file > /tmp/log-snapshot
> } else {
>   aws s3api put-object --bucket logs --key logs/latest --body /tmp/log-snapshot
> } finally { 
>   rm -f /tmp/log-snapshot
> }
> ```
> 
> would it not be sufficient to simply do
> 
> ```
> try {
>   cat log-file > /tmp/log-snapshot
>   aws s3api put-object --bucket logs --key logs/latest --body /tmp/log-snapshot
> } finally { 
>   rm -f /tmp/log-snapshot
> }
> ```
> 
> If there is a valid use case for `try/else/finally` I'd be happy to add it to this PR

Right, there is no reason to use `try/else/finally`, the `else` clause is there to pair with the `catch`.

Took a look at Python's `try` syntax and it does require `else` to come after a `catch` clause, so I'll also make that a requirement too.

website/ref/language.md
Duncan3142: I made the change but it looks like this PR is already merged

I've created a new PR which addresses this comment, and also fixed some other issues with the command outputs I noticed

https://github.com/elves/elvish/pull/1783

Duncan3142:
> > As an observation, the syntax for `try/else/finally` is a bit confusing as a new user.
> > The syntax implies the `else` happens instead of the `try`
> > Also, is there a need for such a construct?
> > For example, instead of
> > ```
> > try {
> >   cat log-file > /tmp/log-snapshot
> > } else {
> >   aws s3api put-object --bucket logs --key logs/latest --body /tmp/log-snapshot
> > } finally { 
> >   rm -f /tmp/log-snapshot
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > would it not be sufficient to simply do
> > ```
> > try {
> >   cat log-file > /tmp/log-snapshot
> >   aws s3api put-object --bucket logs --key logs/latest --body /tmp/log-snapshot
> > } finally { 
> >   rm -f /tmp/log-snapshot
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If there is a valid use case for `try/else/finally` I'd be happy to add it to this PR
> 
> Right, there is no reason to use `try/else/finally`, the `else` clause is there to pair with the `catch`.
> 
> Took a look at Python's `try` syntax and it does require `else` to come after a `catch` clause, so I'll also make that a requirement too.

Thanks for clarifying

PR #1783, opened 2024-04-01
Try docs

Duncan3142:
Add clarifications to `try` construct docs and extend examples

Addresses comments from https://github.com/elves/elvish/pull/1782 which weren't addressed before it got merged


xiaq:
Thanks. After making the change to require `catch` when using `else` I feel it makes more sense to document that as part of the `else` bullet rather than the overall structure. I've cherry-picked the examples you added in 1d45b6e9ed154b63ebbc3f34ddbb16151ba5fa42, so I'll close this now.

PR #1787, opened 2024-04-10
chore: fix function names in comment

goodfirm:


krader1961:
This looks fine to me but I'm curious how you found these errors. Is this something that can be automated as a lint check that reports such errors? Such as those done by `make most-checks`.

krader1961:
> Thank you for your reply. At present, there is no relevant Lint. I obtained it locally by parsing the ast and manually selecting it

I am curious if you are you working on creating a tool similar to `goimports` or `gofmt` that automates detection of documentation errors such as those fixed by this pull-request. Such a tool would have immense value if the false positives can be kept to zero so that it only reports actual documentation errors that need correction. But even if your tool has a small number of false positives it could still be useful.

xiaq:
Thanks!

@goodfirm do you mind sharing the check you've worked on? Godoc comments for exported functions in the Elvish codebase are always supposed to start with the symbol being documented, so having a check specifically against that would still be useful.

PR #1791, opened 2024-04-19
fix: close cpu profile

testwill:


xiaq:
Thanks!

There's some CI failure due to whitespace formatting.

PR #1792, opened 2024-04-19
Improve error messages for failed plugin loads

dann-merlin:
This improves the error messages you get when elvish fails to load a go module.
Currently elvish almost always just fails with a NoSuchModule error, even when the error was actually a library mismatch or a missing Ns symbol in the .so file.

### DIFF
--- a/pkg/eval/builtin_special.go
+++ b/pkg/eval/builtin_special.go
@@ -40,8 +40,12 @@ var IsBuiltinSpecial = map[string]bool{}
 
 // NoSuchModule encodes an error where a module spec cannot be resolved.
 type NoSuchModule struct{ spec string }
+
 // ModuleLoadError encapsulates an error message a failed module load operation.
-type ModuleLoadError struct{ spec string; errormsg string }
+type ModuleLoadError struct {
+	spec     string
+	errormsg string
+}
 
 // Error implements the error interface.
 func (err NoSuchModule) Error() string { return "no such module: " + err.spec }
--- a/pkg/eval/builtin_special.go
+++ b/pkg/eval/builtin_special.go
@@ -40,9 +40,14 @@ var IsBuiltinSpecial = map[string]bool{}
 
 // NoSuchModule encodes an error where a module spec cannot be resolved.
 type NoSuchModule struct{ spec string }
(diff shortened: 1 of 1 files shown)


dann-merlin:
Forgot to run the linters/tests. Updated it now.

pkg/eval/builtin_special.go
```diff
@@ -41,8 +41,17 @@ var IsBuiltinSpecial = map[string]bool{}
 // NoSuchModule encodes an error where a module spec cannot be resolved.
 type NoSuchModule struct{ spec string }
 
+// ModuleLoadError encapsulates an error message a failed module load operation.
+type ModuleLoadError struct {
+	spec     string
+	errormsg string
```
xiaq, line 47: Wrapper errors in Go usually store the original `error` and implement the `Unwrap` method as documented in https://pkg.go.dev/errors.

pkg/eval/builtin_special.go
```diff
@@ -41,8 +41,17 @@ var IsBuiltinSpecial = map[string]bool{}
...
+}
+
 // Error implements the error interface.
 func (err NoSuchModule) Error() string { return "no such module: " + err.spec }
+func (err ModuleLoadError) Error() string {
+	return "when trying to load module: \"" + err.spec + "\" this error occurred: " + err.errormsg
```
xiaq, line 53: Elvish's codebase follows this convention: https://go.dev/wiki/CodeReviewComments#error-strings. I'd write this error message as simply:

```go
"load module " + err.spec + ": " + err.errmsg
```

pkg/eval/builtin_special.go
```diff
@@ -41,8 +41,17 @@ var IsBuiltinSpecial = map[string]bool{}
 // NoSuchModule encodes an error where a module spec cannot be resolved.
 type NoSuchModule struct{ spec string }
 
+// ModuleLoadError encapsulates an error message a failed module load operation.
+type ModuleLoadError struct {
```
xiaq, line 45: This should probably be `PluginLoadError` because it's not used for Elvish modules.

xiaq:
Thanks, there are some style issues that I'll fix up after merging.

Issue #1142, opened 2020-09-05
Repeated variable names in assignments or function arguments should be disallowed


hanche:
Consider this:

```
⬥ x x = foo bar
⬥ put $x
⮕ bar
```

or this:

```
⬥ fn foo [x x]{ put $x }
⬥ foo this that
⮕ that
```

You can do the same with option arguments as well. The following example seems particularly confusing:

```
⬥ fn foo [&x=one x]{ put $x }
⬥ foo two
⮕ one
⬥ foo &x=three two
⮕ three
```

This should be an error. It does not seem useful, and can lead to confusing errors.

**Note** however that repeated uses of the blackhole variable `_` is useful and should remain legal:

```
⬥ fn butlast [@_ x _]{ put $x }
⬥ butlast 1 2 3 4 5
⮕ 4
```

krader1961:
I'm inclined to agree with all your points. The `fn foo [&x=one x]{ put $x }` example is particularly insidious because you might expect the option and parameter namespaces to be distinct; and perhaps they should be but that's a different issue.

**Note:** The blackhole var is valid as an option:

```
> [&_=v]{ echo $_ } &_=v2
v2
```

PR #1796, opened 2024-04-29
Add an error handling when duplicated function arguments found 

jeffxchu:
Avoid duplicated arguments in the lambda.

_Before_ we allowed the following `fn` definition including duplicated args
```
> fn foo {|x x| puts $x} 
```
_Now_ we catch duplicated argument 'x' error and display an error message at the compile time in the console. e.g.
```
> compilation error: [interactive]:1:12-12: duplicate argument name 'x'
```

Related to https://github.com/elves/elvish/issues/1142

### DIFF
--- a/pkg/eval/compile_value.go
+++ b/pkg/eval/compile_value.go
@@ -377,10 +377,12 @@ func (cp *compiler) lambda(n *parse.Primary) valuesOp {
 		restArg       int = -1
 		optNames      []string
 		optDefaultOps []valuesOp
+		seenNames     map[string]bool
 	)
 	if len(n.Elements) > 0 {
 		// Argument list.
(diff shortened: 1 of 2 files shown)


pkg/eval/compile_value.go
```diff
@@ -377,10 +377,12 @@ func (cp *compiler) lambda(n *parse.Primary) valuesOp {
 		restArg       int = -1
 		optNames      []string
 		optDefaultOps []valuesOp
+		seenNames     map[string]bool
```
xiaq, line 380: This variable is only used inside the `if len(n.Elements) > 0 { ... }` block, so it's better kept inside that block. The other variables are declared here because they are used throughout the function.

PR #1801, opened 2024-05-03
pkg/cli: disable termios `IXON` flag, which prevents `C-q` and `C-s` keymaps.

plaukiu:
`IXON` and the messages that it enables have no use in current terminals. According to wiki, it's been used to temporarily halt the process whenever slow hardware devices couldn't catch up with outputs. This is not a problem nowadays, so it should be safe to keep this flag disabled.

### DIFF
--- a/pkg/cli/term/setup_unix.go
+++ b/pkg/cli/term/setup_unix.go
@@ -26,6 +26,7 @@ func setup(in, out *os.File) (func() error, error) {
 	term.SetICanon(false)
 	term.SetIExten(false)
 	term.SetEcho(false)
+	term.SetIXON(false)
 	term.SetVMin(1)
 	term.SetVTime(0)
 
--- a/pkg/sys/eunix/termios.go
+++ b/pkg/sys/eunix/termios.go
@@ -63,6 +63,11 @@ func (term *Termios) SetICRNL(v bool) {
 	setFlag(&term.Iflag, unix.ICRNL, v)
 }
 
+// SetICRNL sets the IXON iflag bit
+func (term *Termios) SetIXON(v bool) {
+	setFlag(&term.Iflag, unix.IXON, v)
+}
+
 func setFlag(flag *termiosFlag, mask termiosFlag, v bool) {
 	if v {
 		*flag |= mask


plaukiu:
(I'm sorry if I was too quick with this PR, I get too excited sometimes haha)

pkg/sys/eunix/termios.go
```diff
@@ -63,6 +63,11 @@ func (term *Termios) SetICRNL(v bool) {
 	setFlag(&term.Iflag, unix.ICRNL, v)
 }
 
+// SetICRNL sets the IXON iflag bit
```
xiaq, line 66: Should be "SetIXON ..."

pkg/cli/term/setup_unix.go
```diff
@@ -26,6 +26,7 @@ func setup(in, out *os.File) (func() error, error) {
 	term.SetICanon(false)
 	term.SetIExten(false)
 	term.SetEcho(false)
+	term.SetIXON(false)
```
xiaq, line 29: It should still be possible for the user to explicitly opt into flow control with `stty ixon`. Instead of clearing IXON here (at every new prompt), Elvish should clear IXON exactly once at startup.

Another complicating factor is that Elvish also tries to leave the termios properties unchanged if it's called from another shell - if you call `elvish` from `bash` and exit `elvish`, the termios should be left unchanged. So the overall logic should be like:

At startup, check if IXON is on:
- If it's off, nothing to do.
- If it's on, turn it off and turn it back on before Elvish exits.

The place to do that is probably in `pkg/shell/interact.go`.

xiaq:
I've merged this to a separate branch and noted down some issues to address before merging into master. I'll probably be able to do the fixup myself on Monday, but feel free to open a further PR before that if you like.

pkg/cli/term/setup_unix.go
plaukiu: The usual way to keep termios unchanged that I encountered in terminal applications is to store a copy of the original termios and then reset it on exit.

xiaq:
Sorry I forgot about integrating this for a while 🥲. I applied my followup changes and it's been merged in https://github.com/elves/elvish/commit/9bfc49aa9c1de210b245481888a13784052045bb!

PR #1816, opened 2024-05-26
value-types.md: render `for` as inline code

extemporalgenome:


xiaq:
Thanks, but the usage was a deliberate stylistic choice.

PR #1823, opened 2024-07-06
Update documentation for kind-of

dann-merlin:
This commit updates the documentation of the builtin kind-of function to include the word "type", so it's easier to find for people that are looking for it under that name.


xiaq:
The "kind" is not _really_ a type but I take the point that mentioning the word there is useful (as long as it still exists...)

PR #1831, opened 2024-08-12
extend lastcmd mode to support all the previous cmds

tw4452852:
Sometimes, I found that what I want is not just last command, but even older one.

### DIFF
--- a/pkg/cli/tk/listbox.go
+++ b/pkg/cli/tk/listbox.go
@@ -35,6 +35,10 @@ type ListBoxSpec struct {
 	OnSelect func(it Items, i int)
 	// A function called on the accept event.
 	OnAccept func(it Items, i int)
+	// A function called when selecting before the beginning of the items.
+	OnUnderFlow func(ListBox)
+	// A function called when selecting after the ending of the items.
+	OnOverFlow func(ListBox)
 	// Whether the listbox should be rendered in a horizontal layout. Note that
 	// in the horizontal layout, items must have only one line.
 	Horizontal bool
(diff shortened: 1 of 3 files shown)


xiaq:
I feel "rolling over" to the previous/next command is not exactly idea:

- I might hold down Up or Down to navigate to a far away item, and the worst thing that would happen is that I would stop at the first or last item. But now there's a danger of going over to another command
- It can also take a bit too many keystrokes to fully go over one command

Left and Right are not used in lastcmd mode, what about using them to navigate commands?

tw4452852:
@xiaq How about using `Alt + <Up/Down>`? I prefer to reserve `<Left/Right>` to be used as selecting left/right item in the future.

xiaq:
> @xiaq How about using `Alt + <Up/Down>`? I prefer to reserve `<Left/Right>` to be used as selecting left/right item in the future.

Alt-Up/Down sound fine too!

tw4452852:
> > @xiaq How about using `Alt + <Up/Down>`? I prefer to reserve `<Left/Right>` to be used as selecting left/right item in the future.
> 
> Alt-Up/Down sound fine too!

Updated in the latest version.

pkg/cli/tk/listbox.go
```diff
@@ -284,6 +294,12 @@ func (w *listBox) Handle(event term.Event) bool {
 	case term.K(ui.Enter):
 		w.Accept()
 		return true
+	case term.K(ui.Up, ui.Alt):
+		w.OnUnderFlow(w)
+		return true
+	case term.K(ui.Down, ui.Alt):
+		w.OnOverFlow(w)
+		return true
```
xiaq, line 302: It would be nicer if this is part of the keybinding rather than hardcoded in the event handler.

PR #1838, opened 2024-09-04
pkg/cli: Add histlist option to reverse listing

mpenick:
This provides a variable (`edit:histlist:reverse`) to reverse the listing, with more recent commands at the top, but leaves the default order as more recent at the bottom.

```
set edit:histlist:reverse = $true
```

Reversing (most to least recent commands) the histlist makes it easier to see more recent commands quickly without having to look near the bottom of the screen.

### DIFF
--- a/pkg/cli/modes/histlist.go
+++ b/pkg/cli/modes/histlist.go
@@ -24,6 +24,9 @@ type HistlistSpec struct {
 	// Dedup is called to determine whether deduplication should be done.
 	// Defaults to true if unset.
 	Dedup func() bool
+	// Reverse is called to determine whether the commands should be listed most recent to less recent.
+	// Default to false if unset.
+	Reverse func() bool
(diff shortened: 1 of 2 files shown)


krader1961:
Does this have any implications for the `&newest-first` option of the [`edit:command-history`](https://elv.sh/ref/edit.html#edit:command-history) command? Should there be an explicit mention that setting this to `$true` does not affect the default order of that command? And should the documentation for that command have at least a "see also" reference to this variable? Also, AFAICT, this doesn't document the new variable and it doesn't include any unit tests to verify the behavior introduced by this change.

xiaq:
Hi! Not sure about the exact reason you closed the PR, but this seems like a reasonable feature request.

I'm in the process of rewriting the Elvish TUI so I probably won't be accepting this PR as is (in particular, I'll probably implement a different API), but I filed #1847 to track this.

PR #1854, opened 2024-11-06
Prefer powershell script over cmd scripts

### LINKED ISSUE #1853: Prefer `.ps1` than `.cmd` when both file exists in path
### What new feature should Elvish have?

Currenlty on Windows, when both `.ps1` and `.cmd` exists for a command (e.g. `scoop`), elvish will prefer invoking the `.cmd`.

However, cmd scripts are bad on UX. Specifically, It prompts "Terminate batch job (Y/N)" when <kbd>Ctrl</kbd>+<kbd>C</kbd> pressed. Therefore, it should be avoided if alternatives exists.

image

0.21.0+official

### Code of Conduct

balthild:
Closes #1853

### DIFF
--- a/pkg/fsutil/search_windows.go
+++ b/pkg/fsutil/search_windows.go
@@ -13,10 +13,15 @@ func isExecutable(stat os.FileInfo) bool {
 	return !stat.IsDir() && isExecutableExt(filepath.Ext(stat.Name()))
 }
 
-// Determines determines a file name extension is considered executable.
-// It honors PATHEXT but defaults to extensions ".com", ".exe", ".ps1", ".bat", ".cmd"
+// Determines a file name extension is considered executable.
+// It honors PATHEXT but defaults to extensions ".com", ".exe", ".bat", ".cmd"
 // if that env var isn't set.
+// Additionally, it treats ".ps1" as executable.
 func isExecutableExt(ext string) bool {
+	if ext == ".ps1" {
+		return true
+	}
+
 	validExts := make(map[string]bool)
 	if pathext := os.Getenv(env.PATHEXT); pathext != "" {
 		for _, e := range filepath.SplitList(strings.ToLower(pathext)) {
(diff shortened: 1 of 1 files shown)


xiaq:
Thanks, I like the idea in principle, but looking at the implementation, a problem is that there are actually two different PowerShell's:

- The classical "Windows PowerShell" that comes pre-installed in modern Windows versions, with the binary name `powershell.exe`.
- The modern "PowerShell Core" that needs to be separately installed, with the binary name `pwsh.exe`.

As far as I know, the modern `pwsh.exe` is not completely backwards compatible with the classical `powershell.exe`. The current implementation picks `powershell.exe`, but I don't feel Elvish should be making the choice on the user's behalf.

We could in principle implement an option for choosing which PowerShell to use, but I feel this would be overcomplicating what's supposed to simplify things for users.

I noticed that PowerShell itself will happily execute a .ps1 file in PATH (both `powershell.exe` and `pwsh.exe`, although the former has an annoying policy thing that prevents it by default), but then PowerShell knows what its own path is, and if you're already using say `pwsh.exe`, this is arguably some signal that you prefer it. The corollary seems to be that Elvish *is* in a position to special-case `.elv` files on Windows...

All in all I'm leaning towards rejecting this PR. Let me know if I missed something.

balthild:
Maybe a solution is to find out which program that `.ps1` files are associated with, but it requires some extra work to implement. If you think this works, I may find take some time to try figuring it out.

balthild:
Hello! Recently, I found in [an MS doc](https://learn.microsoft.com/en-us/powershell/scripting/whats-new/differences-from-windows-powershell?view=powershell-7.5) that `pwsh.exe` could be compatible with `powershell.exe` through .NEt Standard 2.0:

> With the advent of .NET Standard 2.0, PowerShell can load many traditional Windows PowerShell modules without modification. Additionally, PowerShell 7 includes a Windows PowerShell Compatibility feature that allows you to use Windows PowerShell modules that still require the full framework.
?
Maybe it's OK to just try `pwsh.exe` first, and fallback to `powershell.exe` if `pwsh.exe` doesn't exist?

balthild:
> Maybe it's OK to just try pwsh.exe first, and fallback to powershell.exe if pwsh.exe doesn't exist?

Update: Just found that this is just what `cmd` does.

PR #1862, opened 2024-11-18
Fix defer in a loop causing early exit.

### LINKED ISSUE #1849: Defer inside a for loop causes early exit
### What happened, and what did you expect to happen?

This is unexpected:

```elvish-transcript
for v [a b c] {
  defer { put defered }
  put $v
}
▶ a
▶ defered
```

0.22.0-dev.0.20241002144851-c20fac4126aa-dirty

### Code of Conduct

MrOverlord:
Fixes #1849.
My confidence is a bit low in this change, as I am not sure I understood the semantics of error propagation in Elvish's code correctly.


xiaq:
Thanks for the fix! I'm slightly puzzled how this fix works - it seems there's something deeper that's wrong. I'll look into this a bit more and get back to this.

Issue #1833, opened 2024-08-25
epm not recognizing custom domain config


ndarwincorn:
### What happened, and what did you expect to happen?

Similar to #745, I host my elvish theme module in sourcehut.

Just updated to v0.21 and when ripping out some module imports whose maintainers haven't updated to replace eawk, `epm` isn't recognizing my `epm-domain.cfg`:

```
$ epm:upgrade git.sr.ht/~ndarwincorn/elvish-theme
=> No config for domain 'git.sr.ht'.

$ echo $epm:managed-dir
$E:HOME/.local/share/elvish/lib # actual output here is my home dir, not "$E:HOME"

$ cat $E:HOME/.local/share/elvish/lib/git.sr.ht/epm-domain.cfg 
{
    "method": "git",
    "protocol": "https",
    "levels": "2"
}
```

I see the epm module hasn't been changed in 7 months so I'm not sure where to start digging on this. `$epm:-domain-config-file~` looks right in my local shell at a cursory glance.

### Output of "elvish -version"

v0.21.0

### Code of Conduct

- [X] I agree to follow Elvish's [Code of Conduct](https://src.elv.sh/CODE_OF_CONDUCT.md).

TrevorBender:
I believe this defect is introduced by the commit: https://github.com/elves/elvish/commit/3dad55a2523afd47abf5a87aece3cfd75e8b1475  and more specifically the logic on line 185.

The issue specifically exists if the domain config file is a symbolic link. Which may be the case if you use something like stow or nix to manage the file.

To demonstrate the issue:

```elvish
use epm
use path

# Previous logic with test -f where config file is a  symlink
put ?(test -f (epm:-domain-config-file my.custom.domain))     # $ok

# new logic with path:is-regular where config file is a symlink
path:is-regular (epm:-domain-config-file my.custom.domain)     # $false

# The fix, is to use the follow-symlink option
path:is-regular &follow-symlink=$true (epm:-domain-config-file my.custom.domain)    # $true
```

xiaq:
Should be fixed by #1878 now.

PR #1878, opened 2024-12-29
fix: epm domain config file can be a symlink

TrevorBender:
With v0.21.0, if the custom package domain file is a symbolic link epm commands will fail with:

```text
=> No config file for domain 'my.custom.domain'
```

This may be the case if the file is managed by something like stow, nix, or home-manager.

Fix by following symbolic links.

issue: #1833


xiaq:
Thanks!

PR #1890, opened 2025-02-28
pkg/mods/epm: Add sourcehut to default domain list

ejrichards:
This adds [sourcehut](https://git.sr.ht/) as a source for epm.  I feel it has become a popular enough open source hosting to warrant adding it to the default list. I'm currently in the process of moving my personal projects there.


xiaq:
Thanks!

Issue #1897, opened 2025-04-25
30 second startup delay on windows


chtenb:
### What happened, and what did you expect to happen?

Hi, I'm new to elvish. It looks like a fun crossplatform shell that I'm considering as replacement for nushell. However, I'm having a strange issue.

Elvish.exe suffers from startup delays of about 20-30 seconds on my work windows laptop. I'm not sure why. I don't have any rc.elv.
Usually it happens the first time I start it, then for some time new instances spawn instantly, but then after a minute or so it's slow again.

On WSL this problem does not occur. On my personal windows PC it also does not occur.
It does not matter which terminal emulator I use.
However, it does not seem to be an Anti-Virus issue at first glance, since windows defender does not report any action.
Deploying a windows dev-drive or adding exclusions to the windows virus scanner also does not change the startup delay.
Moreover, `elvish -h` does not have the delay, so it's really caused by something that is done by elvish.exe.

**Daemon Log**

```
[daemon] 2025/04/25 09:52:43 pid is 30160
[daemon] 2025/04/25 09:52:43 going to listen C:\Users\ChieltenBrinke\AppData\Local\Temp\elvish-ChieltenBrinke\sock
[store] 2025/04/25 09:52:43 initializing store
[store] 2025/04/25 09:52:43 initialized store
```
It hangs after this line, but no messages are written after that.

**Procmon events in the slow case:**

```
"6:44:42.7799126 PM","elvish.exe","10112","RegQueryValue","HKLM\System\CurrentControlSet\Services\Tcpip\Parameters\Domain","SUCCESS","Type: REG_SZ, Length: 2, Data: "
"6:44:42.7799246 PM","elvish.exe","10112","RegCloseKey","HKLM\System\CurrentControlSet\Services\Tcpip\Parameters","SUCCESS",""
"6:45:04.9706849 PM","elvish.exe","10112","CreateFile","\\AzureAD\PIPE\samr","BAD NETWORK PATH","Desired Access: Generic Read/Write, Disposition: Open, Options: , Attributes: n/a, ShareMode: Read, Write, AllocationSize: n/a"
"6:45:04.9712146 PM","elvish.exe","10112","RegOpenKey","HKLM\System\CurrentControlSet\Services\Tcpip\Parameters","REPARSE","Desired Access: Read"
[...]

PR #1898, opened 2025-04-25
Use USERPROFILE on windows instead of HOME

chtenb:
My first Go pull request.
Addresses one part of the issue described in https://github.com/elves/elvish/issues/1897


oliverhr:
`HOME` is not set by default on windows, Powershell and Git-Bash shell set this for it's on interactive sessions. Maybe it would be a good idea to set `$E:HOME` if is not set to `USERPROFILE` too.

PR #1903, opened 2025-07-07
website/ref/language: add hints for list operations

matttproud:
This commit modifies the language reference section on lists and provides hints how a user can emulate concatenating or prepending to a list.  The goal is to surface what is to be answers to common usage needs, especially for users coming to Elvish from other language ecosystems.


matttproud:
@xiaq: apologies for letting this sit so long.  I had a chance to look at what failed in CI and fixed it in the current revision.

PR #1913, opened 2025-11-15
Support Git-based versioning in epm

giancosta86:
Elvish is a *superlative* shell, always surprising me with its excellent quality and minimalist flexibility - which is why I feel now the need for extending it with **semver-based package versioning**, at least for **Git repositories**, so as to be able to handle the evolution of software architectures over time. 

The mechanics of the present pull request are thoroughly described  in the README of my [epm-plus](https://github.com/giancosta86/epm-plus) project, which contains the very same code as the PR, plus a few lines to patch epm at runtime.

Please, find attached a suite of proof-of-concept tests: they are based on my [velvet](https://github.com/giancosta86/velvet) test framework and runner for Elvish, and can be run via the `verify.elv` script:

```elvish
elvish pkg/mods/epm/verify.elv
```

from the Elvish project directory.


giancosta86:
I've just added - via a dedicated commit - the concept of **metadata-driven install**, as well as the related `devDependencies` field for `metadata.json`; the details and the rationale for both are described, as usual, in the documentation for my [epm-plus](https://github.com/giancosta86/epm-plus) project.

PR #1914, opened 2025-11-16
Provide autosuggestions

### LINKED ISSUE #322: browser style autocompletion
elvish has tab-completion, but the feature i miss most from fish is browser style as-you-type autocompletion, where typing a command will show a preview of an autocompletable command. requesting this!

aca:
Provide fish like autosuggestion. By default it suggest from history. But can be customized (maybe use it to provide AI suggestions?)

```
# enabled by default
set edit:auto-suggestion:enabled = $true

# optional
set edit:auto-suggestion:provider = {|code|
  if (str:has-prefix $code "hello ") {
    put "world"
  }
}

# default keymap
set edit:insert:binding[Ctrl-F] = $edit:auto-suggestion:accept~
```

fixes #322


aca:
image

PR #1917, opened 2026-02-04
pkg/edit: add support for OSC 133 semantic prompts

### LINKED ISSUE #1808: Support semantic prompts (OSC 133)
### What new feature should Elvish have?

There doesn't appear to be an option in `edit` for the elvish editor to automatically emit [OSC 133 codes](https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md). These codes allow REPLs/shells to inform a terminal about the structure of their output, enabling features like those of the late [Final Term](https://web.archive.org/web/20150207021142/http://finalterm.org/) or just [wezterm](https://wezfurlong.org/wezterm/shell-integration.html).

This would be possible to implement oneself in `rc.elv`, except OSC escape codes emitted by the prompt are escaped (heh), which means OSC 133 B (the one that separates the prompt from user input) cannot be emitted at the end of the prompt. Emitting it at the end of `edit:before-readline` makes the prompt considered part of user input (at least by wezterm), which is unideal.

I'm not much of a Go person, but if there's maintainer interest in this feature, I could probably submit a PR for it.

0.20.1+archlinux4

### Code of Conduct

jparise:
OSC 133 semantic prompt sequences help terminals understand shell command structure for features like command navigation, intelligent text selection, and command status indicators.

https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md

This change adds support for OSC 133 A (prompt start), B (input start), C (command start), D (command exit), and P;k=r (right prompt). These sequences are only emitted to TTYs in interactive shell sessions.

I introduced the concept of "zero-width" buffer cells to prevent these "invisible" sequences from contributing to line width calculations. This adds a small (`int`) cost to each `Cell` instance. I explored some alternatives (interfaces, out-of-band tracking, a magic "style" value), but this ended up being the clearest and most efficient approach.

This feature is controlled via a new `edit:shell-integration` variable. It defaults to `$true` (enabled) but can be disabled at any time. It could be generalized to a map of shell integration features later while maintaining backwards compatibility (e.g. `$true` means all enabled):

    set edit:shell-integration = [ &osc133=$true &osc8=$true ]

I also explored ways to implement this support at the script level, but the existing `edit:` hooks don't cover enough cases, and the prompt's asynchronous, control-sequence-stripping output was too limiting.

Closes #1808

### DIFF
--- a/pkg/cli/app.go
+++ b/pkg/cli/app.go
@@ -69,6 +69,7 @@ type app struct {
 	RPromptPersistent func() bool
 	BeforeReadline    []func()
 	AfterReadline     []func(string)
+	ShellIntegration  func() bool
 	Highlighter       Highlighter
 	Prompt            Prompt
 	RPrompt           Prompt
@@ -99,6 +100,7 @@ func NewApp(spec AppSpec) App {
 		RPromptPersistent: spec.RPromptPersistent,
 		BeforeReadline:    spec.BeforeReadline,
 		AfterReadline:     spec.AfterReadline,
+		ShellIntegration:  spec.ShellIntegration,
 		Highlighter:       spec.Highlighter,
 		Prompt:            spec.Prompt,
 		RPrompt:           spec.RPrompt,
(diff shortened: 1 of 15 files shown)


jparise:
Here's an example of this running in Ghostty using our semantic prompt inspector overlay:

CleanShot 2026-02-03 at 19 15 53@2x

(orange regions are prompts, blue regions are inputs)

jparise:
Now marking the continuation indent region as a secondary prompt:

image

